PR #1, opened 2015-12-31
First beta to be released as v0

arekkas:
* [x] Client: OAuth2 client implementation
* [x] Generator: Token generator and validator
* [x] Hash: Secure hashing implementation
* [x] Rand: Random helpers using crypto
* [x] Enigma: Helpers for encrypting data (and rotating keys)
* [x] Session: (Authorize) Session management
* [x] OAuth2 grants
  * [x] authorize code grant (auth and token endpoint)
  * [x] implicit grant
  * [x] client credentials grant
  * [x] resource owner password credentials grant
  * [x] refresh grant
* [x] dont allow redirects to unencrypted endpoints unless it's localhost
* [x] Add helpers for validating scopes, expiry time and other things


README.md
```diff
@@ -1,2 +1,176 @@
...
+    return
+}
+
+// you have now access to authorizeRequest.Scope, ...Code ...ResponseTypes ...Scopes ...
+
+// decide what to do based on scope and response type
+// e.g: response, err = oauth.HandleAuthorizeRequest(authorizeRequest)
+
+// set up a session
+// session := oauth2.NewAuthorizeSession(123)
```
agtorre: Is this going to use the JSONSession? I don't totally understand the use-case for this yet. Otherwise I like the separation of concerns. is the fosite.NewOAuth(config) necessary or can it requests just come directly from the config -- similar to the oauth2 client.

README.md
```diff
@@ -1,2 +1,176 @@
...
+Because each component returns a different type, we can be (if safeguards are installed) quite sure, that the developer
+implemented the work flow the right way:
+
+1. `NewAuthorizeRequest(args...) *AuthorizeRequest`: Fetch authorize request information
+2. do whatever you like
+3. `HandleAuthorizeRequest(args...) *AuthorizeResponse`: Handle authorize request (check scopes and response types, hydrate response...)
+4. do whatever you like
+5. `oauth2.NewAuthorizeSession(*AuthorizeResponse) *AuthorizeSession`: A session
+6. do whatever you like, e.g. `session.SetExtra(map[string]interface{"foo": "bar"})`
+7. `oauth2.PersistAuthorizeRequest` persists the request in the database so the token endpoint can look up information
```
agtorre: It sort of seems like this should just happen as part of HandleAuthorizeRequest? You could create the session after NewAuthorize request and pass it into Handle.

arekkas: Good point. Do you have a good wording for `WriteResponse` to imply persistence as well? `PersistAndWriteResponse` is ok but a bit lengthy

README.md
arekkas: Yes, it is going to be JSONSession, I have to reword the readme here.

The session basically takes the AuthorizeRequest object and converts it to a session. For example, it removes they "key" part of the token and enables you to set custom data with `SetExtra` or `Set`. Custom data could be additional metadata for the request or some token or a remote identity provider identifier (dropbox, google).

You're right, it should be a good idea to "merge" NewDefaultConfig and NewOAuth similar to the oauth2 client. Good point! :)

arekkas: i've added a section to the session part. does that make things clearer?

README.md
```diff
@@ -1,2 +1,195 @@
...
+    //      "userEmail": "foo@bar",
+    //      "lastSeen": new Date(),
+    //      "usingIdentityProvider": "google",
+    // })
+
+    // Now is the time to handle the response types
+
+    // ** This part of the API is not finalized yet **
+    // response = &AuthorizeResponse{}
+    // err = oauth2.HandleResponseTypes(authorizeRequest, response, session)
```
agtorre: The one thing to note here is that the OAuth2 errors differs from the root cause of the error. Frequently the OAuth2 error isn't very useful for debugging so there has to be some way to extract one or the other.

README.md
```diff
@@ -1,2 +1,195 @@
...
+    // you have now access to authorizeRequest, Code ResponseTypes, Scopes ...
+    // and can show the user agent a login or consent page.
+
+    // it would also be possible to redirect the user to an identity provider (google, microsoft live, ...) here
+    // and do fancy stuff like OpenID Connect amongst others
+
+    // Once you have confirmed the users identity and consent that he indeed wants to give app XYZ authorization,
+    // you will use the user's id to create an authorize session
+    user := "12345"
+    session := session.NewAuthorizeSession(authorizeRequest, user)
```
agtorre: The session stuff is looking better. I wonder if maybe naming it something else would be better? I suspect this package will conflict with things like gorilla session and other cookie storage.

README.md
```diff
@@ -1,2 +1,195 @@
...
+
+    // You can store additional metadata, for example
+    // session.SetExtra(map[string]interface{}{
```
agtorre: It would also be nice to have something here like sessions.AddExtra(key, value)

README.md
```diff
@@ -1,2 +1,195 @@
...
+import "github.com/ory-am/fosite/session"
+import "github.com/ory-am/fosite/storage"
+
+// Let's assume that we're in a http handler
+func handleAuth(rw http.ResponseWriter, req *http.Request) {
+    oauth2 := fosite.NewDefaultOAuth2()
+
+    // ** This part of the API is not finalized yet **
+    // var store storage.Storage
+    // this should be something like: store := fosite.NewPostgreSQLStore()
```
agtorre: It may be useful to attach the storage to the oauth2 above ^
foside.NewDefaultOAuth2(store).
Then you can inject oauth2 into middleware context etc and it'll use a global storage.

README.md
arekkas: That's true, error handling is quite tricky. fosite uses [go-errors](https://github.com/go-errors/errors) so you can have an internal stack trace with `err.(*errors.Error).ErrorStack()`. Do you think that is enough?

README.md
arekkas: Good point, do you have any suggestions? Or just keep it in the root package (fosite)?

README.md
arekkas: Good point, I will add this immediately :)

README.md
arekkas: Good thinking!

arekkas:
@agtorre thanks for your continued feedback, it helps me a lot :)

README.md
agtorre: This may be enough. In practice I like to use a logger like logrus and log both the oauth2 error and the internal error in the same place which makes it easier to correlate errors reported by a user to an internal error.

README.md
agtorre: Yeah, the root package is probably simplest, or some other similar word to session like term.
term := term.NewAuthorizeTerm(authorizeRequest, user).

agtorre:
No problem, I'm definitely excited to see where the project goes and to contribute however I can.

leetal:
:+1: Then we are 2 ;)

arekkas:
:) I'll try to work on fosite later today..stay tuned for updates :)

README.md
arekkas: Hey I just saw that SetExtra is defined to use interface{}, not map[string]interface{}. I think this is smarter because the developer has complete freeness over what he passes to extra. What do you think? You can check the interface definition here: https://github.com/ory-am/fosite/blob/unstaged/session.go#L12

README.md
arekkas: ok cool, I think go-errors in combination with logrus is very powerful :)

README.md
arekkas: i moved it to the root package, it seemed the most natural :)

README.md
arekkas: :white_check_mark:

README.md
arekkas: It's now combined in and worded as `FinishAuthorizeRequest` :white_check_mark:

session.go
```diff
@@ -0,0 +1,112 @@
...
+		clientID:      ar.Client.GetID(),
+		state:         ar.State,
+		scopes:        ar.Scopes,
+		redirectURI:   ar.RedirectURI,
+		userID:        userID,
+	}
+}
+
+func (s *defaultSession) SetExtra(extra interface{}) error {
+	result, err := json.Marshal(extra)
```
agtorre: Why does this do a json.Marshal of extra? Can't it just assign it directly?

arekkas: Good question :) The problem is that the session has to persist throughout requests. Most persistent DBALs (sql or mgo) don't handle interface{} types well. I had quite the troubles with [osin](https://github.com/RangelReale/osin) regarding that matter. You can read more on this issue here: https://github.com/ory-am/osin-storage#limitations

That's why I decided to leave Session an interface and give people the chance to implement their own solutions. The `NewAuthorizeSession` was previously known as `NewJSONSession` to indicate that it stores extra data as a JSON string. Maybe I'll rename it to make things clearer.

Does this answer your question sufficiently?

arekkas: I created an issue for this #7

agtorre: I think this makes sense. I'm wondering if this will create some unforeseen side-effects, but can't think of any immediately.

arekkas: yeah, I'm not quite sure myself. I think it might be smarter naming it: `NewSQLAuthorizeSession`. This would imply that this session is specifically optimized for SQL and could justify the json weirdness better. What do you think?

agtorre: Maybe, another option is to use something like gob and store it as a []byte. My concern is that Marshal makes assumptions about interfaces{} so you may not get back out the same thing when you Unmarshal the results. If it were possible to retain the original interface as closely as possible that'd be much better.

arekkas: Jup get it, gob is definitely the better option! I'll change that :) If people want JSON (e.g. beacuse postgres supports JSON columns) I'll leave the old one there as well

README.md
```diff
@@ -1,2 +1,193 @@
...
+        Name string
+    } {
+        Name: "foo"
+    })
+
+    // Now is the time to handle the response types
+
+    // you can use a custom list of response type handlers by setting
+    // oauth2.ResponseTypeHandlers = []fosite.ResponseTypeHandler{}
+    response, err := oauth2.HandleResponseTypes(ctx, authorizeRequest, r)
```
arekkas: The question here is, if we should pass the session as well to HandleResponseTypes. This would enable ResponseTypeHandlers to modify the session and persist data for the grant request (usually /token endpoint). Of course, ResponseTypeHandlers are also able to use their own stores (e.g. OpenIDHandler uses OpenIDConnectStore), which might be more reasonable and would reduce naming conflicts etc.

handler.go
```diff
@@ -0,0 +1,45 @@
...
+// Important: Every ResponseTypeHandler should return ErrInvalidResponseType if it is unable to handle
+// the given request and an arbitrary error if an error occurred
+func (o *OAuth2) NewAuthorizeResponse(ctx context.Context, ar *AuthorizeRequest, r *http.Request) (*Response, error) {
+	var resp = new(Response)
+	var err error
+	var found bool
+
+	for _, h := range o.ResponseTypeHandlers {
+		// Dereference http request and authorize request so handler's can't mess with it.
+		err = h.HandleResponseType(ctx, resp, *ar, *r)
```
arekkas: NewAuthorizeResponse iterates through all ResponseTypeHandlers defined in the configuration. Each handler has access to the response and is able to modify the response. For example, the TokenResponseTypeHandler would initiate the [implicit grant](https://tools.ietf.org/html/rfc6749#section-4.2.1) and return an *access_token* instead of an *code*.

I am not sure how conflicts in the response's output keys should work. Should it be allowed? Should it throw an error? Should the response types decide? Example: Both AResponseTypeHandler sets `response.Set("foo", "bar")` and BResponseTypeHandler sets `response.Set("foo", "baz")`.

arekkas:
I've made some changes, feel free to comment. Getting closer.. :)

handler.go
```diff
@@ -0,0 +1,45 @@
...
+
+var ErrInvalidResponseType = errors.New("This handler is unable handle any of the response types requested by the auhtorize request")
+var ErrNoResponseTypeHandlerFound = errors.New("None of the handler's are able to handle this authorize request")
+
+type ResponseTypeHandler interface {
+	// HandleResponseType handles an authorize request. To extend the handler's capabilities, the http request
+	// is passed along, if further information retrieval is required.
+	//
+	// If HandleResponseType fails, the handler implementation MUST return ErrInvalidResponseType.
+	HandleResponseType(context.Context, *Response, AuthorizeRequest, http.Request) error
```
arekkas: It's probably a good idea to add `fosite.OAuth2` here as a parameter as well. That way, handlers have access to code generation strategies and much more.

arekkas:
I progressed with the internal tools and realized, that some parts of the API where not suited for handling later on. I have now refactored the developer-facing API so that it contains only 4 functions per endpoint (= 8 in total). The internal and extension APIs are already defined but not final yet and thus not documented.

1. [Authorize Endpoint Draft](https://github.com/ory-am/fosite#authorize-endpoint)
2. [Token Endpoint Draft](https://github.com/ory-am/fosite#token-endpoint)

I think the currenct architecture / draft allows for good extensibility and I think that OpenID Connect should be easy to integrate

agtorre:
This is looking pretty good. I still have to go through some of the earlier prs and comments.

arekkas:
I am proud to present you a working integration test for the first part of the OAuth2 authorize workflow (known as consent or auth endpoint). Next step is to implement helpers for the token endpoint.

Thanks for all the feedback, it helped me shape a better API. You can see a working auth endpoint example [here in the tests](https://github.com/ory-am/fosite/blob/c6dcb90ccbd1d7a179a601e0e6d46cc1004cde92/oauth2_test.go#L71-L107).

As you can see, you only need a couple lines of code to get OAuth2 working and yet have extreme power over what things you want to support. Check out the [code workflow plugin](https://github.com/ory-am/fosite/blob/c6dcb90ccbd1d7a179a601e0e6d46cc1004cde92/plugin/code/code.go) to see how a ResponseTypeHandler works.

arekkas:
I also want to hint at the 88% code coverage the tests are achieving. ;)

arekkas:
I've written a small piece of documentation to explain what handlers are. Check it out in the [docs](https://github.com/ory-am/fosite#extensible-handlers)

arekkas:
Good news, the API design is finishing up. There is still stuff to be done but I think we should have a good first beta by the end of the week.

leetal:
:+1:

arekkas:
The token API is now finalized and [documented](https://github.com/ory-am/fosite#token-endpoint)

leetal:
Jesus you're working fast!

arekkas:
I just created an access and refresh token using the explicit authorize code flow :)

```
Got token map[
expires_in:%!s(float64=1800) refresh_token:z4yIogZjBvJQXivYibXRnIWD8etLtNml/PYA6Z1PslI.CUE9T2HcWRhhnMZ0D78O+ziBGE3sFR/x+frD0v9Gbmk token_type:bearer access_token:i0Pipj7d3vElIpqhJkN6QG5bSGlpdxnFbeEdhXD7y28.EJml7G5HFmC3lUJQchIwheMNRdO2+1y6XIVuVzSwHxk] 
```

I will add more handlers today and tomorrow, maybe we'll se a first beta preview in a few days..

arekkas:
HELL YEAH! Thas was intense. I pushed the library to the point where I created a working server side OAuth2 example for the grant type. Check out the awesome gif [here](https://github.com/ory-am/fosite#see-it-in-action) and read how to get the example running [here](https://github.com/ory-am/fosite#see-it-in-action). You can also check out the example's source code [here](https://github.com/ory-am/fosite/blob/unstaged/fosite-example/main.go).

Are you as excited as I am? A secure OAuth2 server in less than 150 chars?

Obviously, there are still a couple of things missing. I will address those soon.

arekkas:
implicit grant :white_check_mark:

arekkas:
client credentials :white_check_mark: 

Implicit and client credentials are now both included in the examples as well

arekkas:
all grant types have now been implemented - but are not fully tested yet.

arekkas:
done :ok_hand: 

next thing is going to be open id connect, but let's see what feedback people give

rbagjani:
I need some help with implementing security using fosite library. I am unable to figure out how to use the scopes to grant the permission to different APIs based on the users.

 I am able to generate tokens through client_credentials grant type. I am taking that only "fosite" can be used in the scope of appClientConfig variable but I need to figure out how to provide different grantScope(permissions) to different users.

Please help as this scenario is not covered in your example. Thanks.

arekkas:
Please don't dig up old issues to ask questions, especially if your question does not have to do anything with the thread you're digging up. This will ping all people involved with something that isn't even remotely related.

You can check the example implementation here, it should help you with your implementation: https://github.com/ory/fosite-example

PR #3, opened 2016-01-01
Added more go versions to build matrix

taion809:
Happy New Year,

This change adds additional builds to the build matrix.

Go version 1.2 and 1.3 are still supported by Ubuntu 14.04 LTS.
So, even though go users tend to use the latest stable it would
be nice to know at a glance that this library is passing for legacy
users.

Go tip has also been added and placed under allowed failures for
bleeding edge users as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ go:
 matrix:
   allow_failures:
     - go: tip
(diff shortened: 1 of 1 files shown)


arekkas:
Thanks!

PR #19, opened 2016-01-14
Include user session data in all calls to storage handlers

cristiangraz:
When retrieving an auth code, you can optionally set the ```TokenSession.Extra``` data to store additional data like a user id that was stored when you created the authorization code. That session data should be available in ```CreateAccessTokenSession``` and ```CreateRefreshTokenSession```, but it ends up coming through nil.

This pull request ensures that session data passes through to the access token and refresh token so you can -- at the very least -- associate user ids with the access/refresh tokens.go in your storage implementation.

### DIFF
--- a/handler/core/client/client.go
+++ b/handler/core/client/client.go
@@ -50,7 +50,7 @@ func (c *ClientCredentialsGrantHandler) HandleTokenEndpointRequest(ctx context.C
 	access, err := c.Enigma.GenerateChallenge(requester.GetClient().GetHashedSecret())
 	if err != nil {
 		return errors.New(fosite.ErrServerError)
-	} else if err := c.Store.CreateAccessTokenSession(access.Signature, requester, &core.TokenSession{}); err != nil {
+	} else if err := c.Store.CreateAccessTokenSession(access.Signature, requester, &core.TokenSession{Extra: session}); err != nil {
 		return errors.New(fosite.ErrServerError)
 	}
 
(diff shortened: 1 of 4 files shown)


cristiangraz:
There is a similar issue happening when you attempt to exchange a refresh token for an access token, but I haven't been able to pinpoint the fix just yet.

arekkas:
Nice find! PR looks good :)

Let me know if you're having troubles fixing the second issue. If you want you can include the second fix here as well. I'll merge the PR then right afterwards

arekkas:
I think the issue you are pointing at is here: https://github.com/ory-am/fosite/blob/master/handler/core/refresh/refresh.go#L68-L75

It might be a good idea to go through all the handlers and see if I accidentaly copy pasted that behvaiour into others as well

cristiangraz:
@arekkas The original pull request actually missed an issue where the session value won't trickle back up since it was passing ```authSess.Extra``` rather than ```session``` pointer provided by the user.

It's the difference between doing this in your storage:
```
session.Extra = &Session{UserID: userID}
```

And this
```
if sess, ok := session.Extra.(*Session); ok {
	sess.UserID = userID
}
```

The first one replaces the pointer, and would require something like ```authSess.Extra``` to be passed to additional function calls. This would provide the value consistently but wouldn't trickle up to the original session pointer the user passed in their handlers.

The second one ensures the type assertion is always true, and is available in the user handlers. I made sure all calls to storage always include the user's session pointer. This PR should be ready.

arekkas:
Awesome, thanks!

PR #20, opened 2016-01-14
Improve API sanity

arekkas:
This PR is aimed at resolving API inconsitencies and giving more power to signing methods


handler/core/strategy.go
```diff
@@ -0,0 +1,22 @@
+package core
```
arekkas: the core package (basically all oauth2 things) now defines what the different strategies must implement. this is an upside because if we have a "supersecure" package, it could use it's own strategy signatures

handler/core/strategy.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"github.com/ory-am/fosite"
+	"golang.org/x/net/context"
+	"net/http"
+)
+
+type AccessTokenStrategy interface {
+	GenerateAccessToken(ctx context.Context, req *http.Request, requester fosite.AuthorizeRequester, session interface{}) (token string, signature string, err error)
```
arekkas: I'm not sure if it is necessary or even desirable to pass the context as it would right now only offer capabilities for aborting the request, which does not really make sense because we're not waiting for a response or anything

handler/core/strategy/hmacsha.go
```diff
@@ -0,0 +1,36 @@
...
+	"golang.org/x/net/context"
+	"net/http"
+)
+
+type HMACSHAStrategy struct {
+	Enigma enigma.HMACSHAEnigma
+}
+
+func (h HMACSHAStrategy) GenerateAccessToken(_ context.Context, _ *http.Request, requester fosite.AccessRequester, _ interface{}) (token string, signature string, err error) {
+	return h.Enigma.Generate(requester.GetClient().GetHashedSecret())
```
arekkas: @leetal this would be the place where you could handle the token claims etc

handler/core/client/client.go
```diff
@@ -47,14 +47,14 @@ func (c *ClientCredentialsGrantHandler) HandleTokenEndpointRequest(ctx context.C
 		return nil
 	}
 
-	access, err := c.Enigma.GenerateChallenge(requester.GetClient().GetHashedSecret())
+	token, signature, err := c.AccessTokenStrategy.GenerateAccessToken(ctx, req, requester, session)
```
arekkas: instead of making an implicit assumption that every token is of type Challenge, we now only require that the token is verifyable with a signature that can be used for database lookup

handler/core/client/client.go
```diff
@@ -13,8 +13,8 @@ import (
 )
 
 type ClientCredentialsGrantHandler struct {
-	// Enigma is the algorithm responsible for creating a validatable, opaque string.
-	Enigma enigma.Enigma
+	// AccessTokenStrategy is the algorithm responsible for creating a validatable token.
+	AccessTokenStrategy core.AccessTokenStrategy
```
arekkas: instead of relying directly on the enigma interface, we now rely on `AccessTokenStrategy.GenerateAccessToken` which is responsible for generating the access token. So we can do something like: 

```go
handler := ClientCredentialsGrantHandler{
    AccessTokenStrategy: &JWTStrategy{},
}
```

handler/core/strategy/hmacsha.go
leetal: Yep! I've looked through the code and all seems well. We can now adapt the stateless behavior as we want :)

arekkas: ok cool. do you think that the solution is to complicated or do you have an idea how to solve this easier?

arekkas: by the way i think it would be okay to remove the context variable as I'm not sure that the generators/ validators really need it. however, I do not want to change the interface after this PR here any more so I can't really decide if keep or kick :D

leetal:
This all seems good! But i think we'll need to expose (or at least discuss alternate solutions to) at least one callback the developers can "subscribe" to by passing in a function that is called right before the ACCESS token is generated. A little like Osin does, but not the whole genration, just a little part of it. Since there is no way for a developer right now to actually modify the claims (for example) that is to be returned to the callee. What i mean is that for example one resource owner grants acces to his data, the developer might just, right before that access token is returned, fetch the resource owner data from DB and inject parts of that data into the JWT claims. I really don't want to hardcode any behavior except lease times and expire times in the lib. What would be counterintuitive with regards of the actual power and core strength of Fosite. I think i'll add such callback to the JWT mechanism by adding it to the JWTEnigma struct since probably no other auth method (except open id connect) can return data in the token.

handler/core/strategy/hmacsha.go
leetal: My intuition says "keep it". Since there is always a good idea for future
development to have one backup of passing data along the call-chain ;)
Nothing that ever should be a praxis, but one that just might save that
situations from alternating any public interfaces.


Mvh

*Alexander Widerberg*
*Vakans Linköpings universitet i programvaruutveckling*


*DatateknologsektionenLinköpings UniversitetMobil: 0763-471640*

2016-01-14 21:50 GMT+01:00 Aeneas <notifications@github.com>:

arekkas:
I think this is already solved because we pass the user session object which is of type interface!

```go
func (h HMACSHAStrategy) GenerateAccessToken(ctx context.Context, r *http.Request, ar fosite.AccessRequester, session interface{})
```

If the developer wants to use JWT, he needs to choose a session that implements for example some interface. In the JWT strategy you could do:

```go
func (h jwt.JWTStrategy) GenerateAccessToken(ctx context.Context, r *http.Request, ar fosite.AccessRequester, session interface{}) {
    jwtSession, ok := session.(*jwt.Session)
    if ; !ok {
       return err
    }
    GenerateJWTToken(jwtSession.GetClaims())
}
```

The developer then might do:

```
// in token endpoint
// ... NewAuthorizeRequest
session := jwt.NewJWTSession()
session.SetSub("peter")
// ... NewAuthorizeResponse(..., session)
// ... Writeresponse
```go

Would that not be enough?

leetal:
Yes, that will do for sure! :) Totally didn't think of that use for the
session, but as you said before, stateless behavior without hussle is
important!


Mvh

*Alexander Widerberg*
*Vakans Linköpings universitet i programvaruutveckling*


*DatateknologsektionenLinköpings UniversitetMobil: 0763-471640*

2016-01-14 22:07 GMT+01:00 Aeneas <notifications@github.com>:

arekkas:
ok cool, i will implement this then :) thanks for your continued feedback, i honestly appreciate it!

arekkas:
The method signatures are in gernal too long and complex. I will replace them with context variables instead

arekkas:
http://play.golang.org/p/iq7iZfhiHm

PR #23, opened 2016-01-15
docs: define implicitHandler

hagbarddenstore:
Someone forgot to rename the variable name when copy-pasting in the example.

Signed-off-by: Kim Johansson <kim@kumiko.se>


arekkas:
Thanks!

PR #39, opened 2016-04-23
Implement feedback from production and freeze API

arekkas:
* [x] write test for oidc hybrid flow
* [x] Refresh grant in example is broken
* [x] IsExpired etc should now be checked at strategy level
* [x] implement oidc token behaviour


arekkas:
* [ ] write test for oidc hybrid flow

PR #49, opened 2016-06-09
readme: remove api stability note, release v1

arekkas:


arekkas:
Real-life useage of hydra is getting closer and closer. Once we're getting nice results there, this will be merged.

arekkas:
fosite will now be shipped using semver and github tags, using gopkg is obselete and therefore "v1" not really an issue

Issue #56, opened 2016-06-30
Return context


ADone:
Hi! I have special case for returning some additional signatures from such methods like `CreateAccessTokenSession`.
And I have just added a return context for ALL endpoints & storages. For example:
```go
type AuthorizeCodeStorage interface {
	CreateAuthorizeCodeSession(
		ctx context.Context,
		code string,
		request fosite.Requester,
	) (context.Context, error)

	GetAuthorizeCodeSession(
		ctx context.Context,
		code string,
		session interface{},
	) (context.Context, fosite.Requester, error)

	DeleteAuthorizeCodeSession(
		ctx context.Context,
		code string,
	) (context.Context, error)
}
```
All tests passes!
Do you interest in that functionality?

arekkas:
sure, create a PR and I will take a look :)

ADone:
@arekkas could you close this too?)

PR #57, opened 2016-06-30
add return context, fix tests for additional out parameter

ADone:
for #56 

Signed-off-by: Artem Nistratov <anistratov@go-promo.ru>

### DIFF
--- a/access_request_handler.go
+++ b/access_request_handler.go
@@ -33,61 +33,62 @@ import (
 //   credentials (or assigned other authentication requirements), the
 //   client MUST authenticate with the authorization server as described
 //   in Section 3.2.1.
-func (f *Fosite) NewAccessRequest(ctx context.Context, r *http.Request, session interface{}) (AccessRequester, error) {
+func (f *Fosite) NewAccessRequest(ctx context.Context, r *http.Request, session interface{}) (context.Context, AccessRequester, error) {
 	accessRequest := NewAccessRequest(session)
 
 	if r.Method != "POST" {
-		return accessRequest, errors.Wrap(ErrInvalidRequest, "HTTP method is not POST")
+		return ctx, accessRequest, errors.Wrap(ErrInvalidRequest, "HTTP method is not POST")
 	} else if err := r.ParseForm(); err != nil {
-		return accessRequest, errors.Wrap(ErrInvalidRequest, err.Error())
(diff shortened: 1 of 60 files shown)


arekkas:
Thanks for the contribution! What's the rational behind returning implicit values? I can see that it gives flexibility but it's nothing short of returning `interface{}` values. What did you try to solve when applying this?

ADone:
Maybe I misunderstood you, but return type is clear. It's `context.Context` and it was created to resolve such issues like share some variables across all execution pipleline. I return a user model associated with added or extracted token, token usage information and some other options. With context I link refresh and access token in PersistAuthorizeCodeGrantSession, because token saving logic already implemented in CreateAccessTokenSession and CreateRefreshTokenSession. I don't want copy&paste it in PersistAuthorizeCodeGrantSession. Instead I put RefreshToken & AccessToken models into new context in its creation functions. I think there are many other suitable variants for return context usage.

arekkas:
Ah sorry, my bad, I misread :)

arekkas:
Sharing data is intended to be happening through `Requester.GetSession() interface{}` / `Requester.SetSession(session interface{})`. That's where you would store user ids and that stuff. Or did I miss something?

arekkas:
Could you maybe show me some usage examples, that would make it easier to understand for me :)

arekkas:
Do you maybe know an example project where contexts are returned as well? I have looked at some go projects but never seen contexts get returned. [Docker](https://github.com/docker/docker/search?p=5&q=context.Context&utf8=%E2%9C%93) for example only returns `context.Context` within `NewSomethingContext()`.

Only because nobody uses it, it doesn't imply that it's not a good idea. However, I need to nullify any doubt of unwanted sideffects or other issues :)

ADone:
I use the `JWT Session` as session interface. Hm, I didn't look at this from this point. May be more preferable to extend JWT session.
In examples I seen only `nil` in session. It is misleading (

arekkas:
Sorry for that. :( Documentation is on my list, but I have a lot to do right now.

Would you maybe be interested in helping here a little bit? Seems like you've already a good picture of how things work!

ADone:
I will try)
Sorry for wasting your time, I close PR.

arekkas:
No worries, I'm sorry that the docs didn't mention that :)

PR #59, opened 2016-07-27
Extract username and password from header and fallback to body

256dpi:
Hi, thanks for this amazing library!

I started to use it to implement an OAuth2 layer for my small API framework: https://github.com/256dpi/fire

While working on the Resource Owner Credential Flow I realized that the credentials have to be submitted in the Header and the Body of the Request. But this behavior contradicts with the following statement in the docs:

> Including the client credentials in the request-body using the two parameters is NOT RECOMMENDED and SHOULD be limited to clients unable to directly utilize the HTTP Basic authentication scheme (or other password-based HTTP authentication schemes).  The parameters can only be transmitted in the request-body and MUST NOT be included in the request URI.

This PR tries to extract the username and password first from the header and then falls back to the body.

### DIFF
--- a/handler/core/owner/owner.go
+++ b/handler/core/owner/owner.go
@@ -28,8 +28,12 @@ func (c *ResourceOwnerPasswordCredentialsGrantHandler) HandleTokenEndpointReques
 		return errors.Wrap(fosite.ErrInvalidGrant, "")
 	}
 
-	username := req.PostForm.Get("username")
-	password := req.PostForm.Get("password")
+	username, password, ok := req.BasicAuth()
+	if !ok {
+		username = req.PostForm.Get("username")
+		password = req.PostForm.Get("password")
+	}
+
 	if username == "" || password == "" {
 		return errors.Wrap(fosite.ErrInvalidRequest, "")
 	} else if err := c.ResourceOwnerPasswordCredentialsGrantStorage.Authenticate(ctx, username, password); errors.Cause(err) == fosite.ErrNotFound {


arekkas:
Glad you found use for it!

Unfortunately, you confused client credentials with resource owner credentials. Client credentials go in the Authorization header while resource owner credentials go in the body :)

arekkas:
If you're looking for example code or for a HTTP server implementation of fosite, check out [hydra](https://github.com/ory-am/hydra) :)

arekkas:
To prove my point, check out the spec: https://tools.ietf.org/html/rfc6749#section-4.3.1

> The client makes a request to the token endpoint by adding the
   following parameters using the "application/x-www-form-urlencoded"
   format per Appendix B with a character encoding of UTF-8 in the HTTP
   request entity-body:
>
>   grant_type
         REQUIRED.  Value MUST be set to "password".
>
>   username
         REQUIRED.  The resource owner username.
>
>   password
         REQUIRED.  The resource owner password.
>
>   scope

256dpi:
I wasn't aware that a client is needed to accomplish the resource owner password flow, I always thought that in this case the user takes the role of the client. ;) But if I think about it again, it makes sense... Thanks for your fast response!

arekkas:
Glad I could help. You always need to identify the client, as he is the one the token is issued for :)

256dpi:
Even if the client is on a platform that is not trusted like a Browser?

As noted here: https://github.com/simplabs/ember-simple-auth/blob/master/addon/authenticators/oauth2-password-grant.js#L47

arekkas:
Identifying is not authenticating. All grant types, except the implicit one, require authentication of the client. The implicit one is the only one you can use in Browsers / on mobile clients.

256dpi:
I guess I need to read the specs again ;). Thanks again for clarifying!

arekkas:
Welcome! If you are not quite familiar with OAuth2, take a look at the [hydra](http://github.com/ory-am/hydra) project. It's basically fosite as a high throughput server and it works on top of any user authentication plus it's user interface agnostic. TL;DR You get to keep all the candy without doing any of the work ;)

There's also some docs and guides which might help you to get a better understanding of OAuth2!

Issue #76, opened 2016-08-21
Consider upgrading to jwt-go 3.0.0


leetal:
Hi Aeneas!

So, i'm still following the development of all your libraries and we've just now started to implement our interfaces against fosite/hydra. It came to a quick halt when we realized that fosite relies upon v2.7.0 of jwt-go, and that we must use 3.0.0 of jwt-go. We also use glide and this is causing a clash since jwt-go have introduced a breaking change by replacing the map[string]interface{} to a jwt.Claims interface (essentially exactly like your jwt_claims.go). 

Do you want help fixing this change (more or less just upgrade to 3.0.0 of jwt-go and create a new type that implements the Claims interface in fosite) or is this on your backlog?

Regards
Alexander W

arekkas:
@leetal yes, I welcome if you could help me with upgrading to 3.0. This has been a pain ever since they broke bc. :)

PR #77, opened 2016-08-22
Updated: jwt-go is now v3.0.0.

leetal:
This PR is in direct response to #76.

NOTE:
No direct breaking API changes should have been introduced in this pull request, as long as no one explicitly uses the raw values of "iat", "exp" and "nbf" claims (was int, is now float). This is due to the fact that jwt-go needs float64 types when validating the claims. I will create a new PR to jwt-go as well that is going to allow int64 values to be parsed. Hopefully it will get merged fast. If that's the case, after a minor upgrade to fosite removing the float64 casts, nothing API-wise will have changed since before the upgrade to jwt-go 3.0.0. 

// Alex

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,25 +1,12 @@
-hash: bfdddfe32cacf648c65b3dd8350e6c748c3ab84e46b47ea4548e0c20fb068034
-updated: 2016-08-09T09:17:29.732545671+02:00
+hash: 0ac686d5fd98f966adf0df5a617514913bdfe9f3c081919434d29395ded25c2f
+updated: 2016-08-22T11:10:29.823423512+02:00
 imports:
-- name: camlistore.org
-  version: 41516af23d58b64efe7712e8be2d3f9fd94fb693
-  subpackages:
-  - pkg/googlestorage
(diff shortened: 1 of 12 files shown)


leetal:
I have supplied a PR to jwt-go now that will fix the "float" issue. Now it is up to them to decide whether or not to merge it.

arekkas:
Hey @leetal thanks for this! Float64 indeed makes sense because javascript is using 64bit floats always. I think it's ok to increase the minor version with this PR.

Let me know if you want to wait for jwt-go to be merged or just merge this directly and release a new version. Both work for me.

leetal:
Yeah, I think it is OK for now. It's a later discussion whether or not to
change back to int64 or not. For now it will probably work for more or less
everyone, even if someone actually is using the raw values or not.
It will probably be a while until jwt-go merge the change. If ever.. ;)

otraore:
Looking forward to seeing this merged in, gives me a reminder I have to update to the new jwt-go api on some personal projects

arekkas:
Ok, thanks for the hard work!

PR #107, opened 2016-10-05
oauth2: added refresh token generation for password grant type

jrossiter:
I'm currently porting over some old authentication services with fosite and have ran into the issue specified here: https://github.com/ory-am/fosite/issues/78

Where internal clients will make use of the password grant type along with the access / refresh tokens.

There was a comment suggesting that this particular feature be added as a separate PR than the issue raised; and since it's been around a month with no updates I decided to give it a go.

Please let me know if anything doesn't meet the RFC or needs changes.

### DIFF
--- a/fosite-example/main.go
+++ b/fosite-example/main.go
@@ -131,6 +131,11 @@ func tokenEndpoint(rw http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	// Grant requested scopes
+	for _, scope := range accessRequest.GetRequestedScopes() {
+		accessRequest.GrantScope(scope)
+	}
+
 	// Next we create a response for the access request. Again, we iterate through the TokenEndpointHandlers
 	// and aggregate the result in response.
 	response, err := oauth2.NewAccessResponse(ctx, req, accessRequest)
--- a/handler/oauth2/flow_resource_owner_test.go
+++ b/handler/oauth2/flow_resource_owner_test.go
@@ -77,6 +77,7 @@ func TestResourceOwnerFlow_PopulateTokenEndpointResponse(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	store := internal.NewMockResourceOwnerPasswordCredentialsGrantStorage(ctrl)
 	chgen := internal.NewMockAccessTokenStrategy(ctrl)
+	rtstr := internal.NewMockRefreshTokenStrategy(ctrl)
 	aresp := fosite.NewAccessResponse()
 	defer ctrl.Finish()
 
@@ -90,6 +91,7 @@ func TestResourceOwnerFlow_PopulateTokenEndpointResponse(t *testing.T) {
 			AccessTokenStrategy: chgen,
 			AccessTokenLifespan: time.Hour,
 		},
+		RefreshTokenStrategy: rtstr,
 	}
 	for k, c := range []struct {
 		description string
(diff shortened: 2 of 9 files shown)


arekkas:
Awesome, thanks for this! I will take a closer look tonight!

fosite-example/main.go
```diff
@@ -131,6 +131,11 @@ func tokenEndpoint(rw http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	// Grant requested scopes
```
arekkas: Why did you add this? Is it to get the resource owner credentials grant working with scopes?

handler/oauth2/flow_resource_owner_test.go
```diff
@@ -111,6 +113,17 @@ func TestResourceOwnerFlow_PopulateTokenEndpointResponse(t *testing.T) {
...
+		{
+			description: "should pass - offline scope",
+			setup: func() {
+				areq.GrantTypes = fosite.Arguments{"password"}
+				areq.GrantScope("offline")
+				rtstr.EXPECT().GenerateRefreshToken(nil, areq).Return("tokenfoo.bar", "bar", nil)
+				store.EXPECT().CreateRefreshTokenSession(nil, "bar", areq).Return(nil)
+				chgen.EXPECT().GenerateAccessToken(nil, areq).Return("tokenfoo.bar", "bar", nil)
+				store.EXPECT().CreateAccessTokenSession(nil, "bar", areq).Return(nil)
+			},
```
arekkas: Should we add a test here if `"refresh_token"` is part of the response? What do you think?

fosite-example/main.go
jrossiter: In `ResourceOwnerPasswordCredentialsGrantHandler.PopulateTokenEndpointResponse()` 
I'm using `if requester.GetGrantedScopes().Has("offline")` to determine if they need a refresh token but it seems like this wasn't set anywhere. 
I compared it to the authorize endpoint which set the scopes like this.

The NewAccessRequest seems to already throw an error if a requested scope isn't allowed by a client so I thought this may be safe to do here.

What do you think?

handler/oauth2/flow_resource_owner_test.go
jrossiter: Good idea; I'll add to the tests to check for `refresh_token` with and without the `offline` scope.

jrossiter: I've merged the new checks.

fosite-example/main.go
arekkas: Yup sounds good!

handler/oauth2/flow_resource_owner_test.go
```diff
@@ -107,14 +111,34 @@ func TestResourceOwnerFlow_PopulateTokenEndpointResponse(t *testing.T) {
...
+				chgen.EXPECT().GenerateAccessToken(nil, areq).Return(mockAT, "bar", nil)
 				store.EXPECT().CreateAccessTokenSession(nil, "bar", areq).Return(nil)
 			},
 		},
 	} {
 		c.setup()
 		err := h.PopulateTokenEndpointResponse(nil, httpreq, areq, aresp)
 		assert.True(t, errors.Cause(err) == c.expectErr, "(%d) %s\n%s\n%s", k, c.description, err, c.expectErr)
+
+		if err == nil {
```
arekkas: You could probably solve this with less coupling of test cases and test executing by introducing a `expect func()` field to the case struct and check here if it's nil and if not execute it. The passing test case could then define a check function that executes the assertion logic. You can check this code for reference: https://github.com/ory-am/fosite/blob/bef61973fdee1a18aedba4e42a1d8977c3f8cc1c/handler/oauth2/flow_refresh_test.go#L92-L98

What do you think of this approach?

jrossiter: I like this approach. Just updated the branch.

256dpi:
Very cool indeed!

AFAIK, the documentation does not explicitly mention that the "offline" scope is used to determine that a client can use refresh tokens. Could you guys maybe add some documentation about how the mechanism is working in fosite?

arekkas:
[this](https://github.com/ory-am/fosite#scopes) could be the right place for it, do you want to create a PR @256dpi ? :)

256dpi:
@arekkas I would love to, unfortunately I'm not aware of all the details of the implementation. Therefore someone that already read into that particular code section might be better suited for the job. ;)

arekkas:
Added: https://github.com/ory-am/fosite/commit/60a767221625d0f6541f203e41a7ef20a1782eb0

I promise to take the time at some point to write some docs on gitbook.

jrossiter:
Thanks!

Issue #111, opened 2016-10-10
oauth2: introspection and revocation


jrossiter:
Do you think it's worth implementing interfaces for token introspection and revocation?

OAuth 2.0 Token Introspection: https://tools.ietf.org/html/rfc7662
OAuth 2.0 Token Revocation: https://tools.ietf.org/html/rfc7009

Seems like it would be possible to add to the OAuth2Provider interface and add some more interfaces:

- IntrospectRequester
- IntrospectResponder
- RevokeRequester
- RevokeResponder

Additional functions for the OAuth2Provider interface:

- NewIntrospectRequest
- NewIntrospectResponse
- WriteIntrospectError
- WriteIntrospectResponse
- NewRevokeRequest
- NewRevokeResponse
- WriteRevokeError
- WriteRevokeResponse

Would adding to OAuth2Provider break BC?

arekkas:
I generally like the idea, we already have the [TokenValidator](https://github.com/ory-am/fosite/blob/master/validator.go#L11) interface which is helping [ValidateToken](https://github.com/ory-am/fosite/blob/master/oauth2.go#L116) and is sort of doing RFC7662, but you have to implement most of the things yourself.

Adding these things to the OAuth2Provider would break BC, but only if someone implements their own OAuth2 implementation which I doubt. Additionally, we're still in the `0.x.y` tag and labeled as release preview, so it should be ok!

arekkas:
The question is if we even need both the request and the response object or if a simple `RevokeToken` is enough. For example, RFC7009 does not have a real response, just HTTP 200 OK: https://tools.ietf.org/html/rfc7009#section-2.2

jrossiter:
That makes sense for revocation. 

What about a signature like this?

`func (f *Fosite) RevokeToken(rw http.ResponseWriter, req *http.Request)`

With this the function would be able to:

- verify the client (`req.BasicAuth`)
- verify token was issued to the client
- call revocation endpoint handlers to perform the clean up

This would require:

- RevocationEndpointHandler interface
- Addition to the Fosite struct

Am I on the right track with this? If so I can give it a shot.

arekkas:
Yes, I think that's very close! I checked out the RFC and it includes the usual error handling from OAuth2, for example when the client credentials are invalid.

I think we should keep the practice of adding the context and maybe leave the response up to the caller:

```
RevokeToken(ctx context.Context, r *http.Request) error
```

[...]

PR #112, opened 2016-10-12
oauth2: added token revocation handling

jrossiter:
Signed-off-by: Jason Rossiter <jrossiter403@gmail.com>


jrossiter:
Related to #111.

Here's my proof of concept for token revocation.

I tried to keep it as close as the other implementations as possible.

compose/compose_oauth2.go
```diff
@@ -118,3 +118,21 @@ func OAuth2ResourceOwnerPasswordCredentialsFactory(config *Config, storage inter
 		},
 	}
 }
+
+// OAuth2TokenRevocationFactory creates an OAuth2 token revocation handler and registers
+// an access token, refresh token and authorize code validator.
+func OAuth2TokenRevocationFactory(config *Config, storage interface{}, strategy interface{}) interface{} {
+	return &struct {
+		*oauth2.TokenRevocationHandler
+		*oauth2.CoreValidator
```
arekkas: The `CoreValidator` looks like he doesn't belong here. This is probably something that the compose package requires, right? I have to check if we can remove this requirement.

handler/oauth2/flow_revoke.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/ory-am/fosite"
+)
+
+type TokenRevocationHandler struct {
+	TokenRevocationStorage TokenRevocationStorage
+}
+
+// RevokeToken implements https://tools.ietf.org/html/rfc7009#section-2.1
+func (r *TokenRevocationHandler) RevokeToken(ctx context.Context, token string, tokenType fosite.TokenType) error {
+	if tokenType == fosite.RefreshToken {
```
arekkas: It looks like this is clashing with the definition:

>   token_type_hint  OPTIONAL.  A hint about the type of the token
           submitted for revocation.  Clients MAY pass this parameter in
           order to help the authorization server to optimize the token
           lookup.  If the server is unable to locate the token using
           the given hint, it MUST extend its search across all of its
           supported token types.  An authorization server MAY ignore
           this parameter, particularly if it is able to detect the
           token type automatically.  This specification defines two
           such values:

revoke_handler.go
```diff
@@ -0,0 +1,109 @@
...
+	r.ParseForm()
+	token := r.PostForm.Get("token")
+
+	var tokenType = TokenType(r.PostForm.Get("token_type_hint"))
+
+	switch tokenType {
+	case AccessToken:
+	case RefreshToken:
+	default:
+		tokenType = RefreshToken
```
arekkas: See comment above, I think that we should leave tokenType empty if it is not given.

arekkas:
Awesome, this looks really good already! I think that we should try and include this here as well:

> If the particular
token is a refresh token and the authorization server supports the
revocation of access tokens, then the authorization server SHOULD
also invalidate all access tokens based on the same authorization
grant (see Implementation Note).

Refresh tokens and access tokens are linked implicitly through the request object at the moment, for example: https://github.com/ory-am/fosite/blob/master/handler/oauth2/flow_resource_owner.go#L64-L76

We could add a unique id to the request objects and use that id to look up linked tokens.

[request.go](https://github.com/ory-am/fosite/blob/master/request.go)

```go
func (a *Request) GetID() url.Values {
	if a.ID == nil {
		a.ID = uuid.New()
	}
	return a.ID
}
```

What do you think?

jrossiter:
Yes that would seem like the only way to link them together at the moment.

The refresh token ID could be linked with `PersistRefreshTokenGrantSession()`.

Would the access token request be assigned the same request ID or have another field like `GeneratedBy` with the ID of the refresh token?

arekkas:
> The refresh token ID could be linked with PersistRefreshTokenGrantSession().
> 
> Would the access token request be assigned the same request ID or have another field like GeneratedBy with the ID of the refresh token?

I think that we don't need to go that far. The request object and request id is generated only once per OAuth2 request. If two tokens (one refresh, one access) are generated during that same request, and they store the request object (true for both refresh and access), the will have access to that request id parameter. Thus, it will be possible to find which access token has been issued using the given refresh token.

Or did I miss something here?

arekkas:
They don't state if I can revoke an already used refresh token, or if I can revoke only "active" refresh tokens. The first (revoking used rt's) would imply that all tokens followed by that rt must be invalidated. The second would imply that we only need to revoke the at and rt that was issued when performing the `refresh_token` grant.

jrossiter:
Yes that's right, I thought a bit too far ahead for that one.

I suspect it may be the latter option for revoking refresh tokens. The clients should have already cycled their old refresh token if they ever went ahead with the `refresh_token` grant.

Which interface would be responsible for removing the additional access token when revoking a refresh token?

1. Will the implemented storage interface have to be responsible for cleaning up the access token?
If the implemented storage interface is responsible then would there need to be a flag to specify:

- Only delete this refresh token (`refresh_token` grant type in `PersistRefreshTokenGrantSession()` does this)
- Delete this refresh token and the linked access token

arekkas:
> I suspect it may be the latter option for revoking refresh tokens. The clients should have already cycled their old refresh token if they ever went ahead with the refresh_token grant.

I think so too, but I'll check out the RFC one more time to be sure. It would still be possible to achieve the former (revoking expired RTs) by passing down the request id when the `refresh_token` grant is used.

> Which interface would be responsible for removing the additional access token when revoking a refresh token?
> 
> Will the implemented storage interface have to be responsible for cleaning up the access token?

I think yes, the storage will be responsible for removing the access token as well.

> If the implemented storage interface is responsible then would there need to be a flag to specify:
> * Only delete this refresh token (refresh_token grant type in PersistRefreshTokenGrantSession() does this)
> * Delete this refresh token and the linked access token

Good thinking here. One possibility would be a separate API call in the storage interface, so it's super-self explaining and transparent during development. So adding something like to [TokenRevocationStorage](https://github.com/ory-am/fosite/pull/112/files#diff-cd1c2f962b9792b5ca8f8ef720241e77R3):

```go
type TokenRevocationStorage interface {
	RefreshTokenStrategy
	// RefreshTokenStorage
	AccessTokenStorage
	AccessTokenStrategy

	RevokeRefreshToken(ctx, signature)
}
```

We could even go so far and separate `Delete[Refresh|Access]TokenSession` and `Revoke[Access|Refresh]Token`, so it's uber-clear that this store action is for revokation. The idea behind `Delete[Refresh|Access]TokenSession` was originally to let developers think about token revokation, but I think that the signature doesn't make that clear enough.

What is your take on this?

jrossiter:
Adding `RevokeRefreshToken(ctx, signature)` would definitely make things a lot clearer for developers.
Creating `RevokeAccessToken(ctx, signature)` would depend on if there are any specific use cases to do additional logic.

We could add both and leave it up to the developer on what extra functionality they'd want in each?

arekkas:
> We could add both and leave it up to the developer on what extra functionality they'd want in each?

Yes, let's do it that way!

oauth2.go
```diff
@@ -114,6 +114,18 @@ type OAuth2Provider interface {
...
+	//
+	// The following specs must be considered in any implementation of this method:
+	// https://tools.ietf.org/html/rfc7009#section-2.1
+	NewRevocationRequest(ctx context.Context, r *http.Request) error
+
+	// WriteRevocationResponse writes the revoke response.
+	//
+	// The following specs must be considered in any implementation of this method:
+	// https://tools.ietf.org/html/rfc7009#section-2.2
+	WriteRevocationResponse(ctx context.Context, rw http.ResponseWriter, err error)
```
arekkas: My bad here, sorry about that. None of the writers are using the context variable so let's remove this here to keep the API homogenous

jrossiter:
One more thing to consider with `RevokeToken()`.

Should this function only be responsible for making sure both `Revoke[Access|Refresh]Token` are called? Should it leave the token revocation and access token discovery process to the storage implementation?

arekkas:
> Should this function only be responsible for making sure both Revoke[Access|Refresh]Token are called? Should it leave the token revocation and access token discovery process to the storage implementation?

Good question. I think that fosite should be responsible for handling most the logic so that developers who use this library make less mistakes. Therefore, I think that discovery should be done in `RevokeToken`.

jrossiter:
If `RevokeToken` needs to do the discovery then there will need to be an efficient way to retrieve a token by the generated request ID.

We'd need a method to do this for both refresh and access tokens. What do you think?

> If the particular
   token is a refresh token and the authorization server supports the
   revocation of access tokens, then the authorization server SHOULD
   also invalidate all access tokens based on the same authorization
   grant (see Implementation Note).  **If the token passed to the request
   is an access token, the server MAY revoke the respective refresh
   token as well.**

arekkas:
> If RevokeToken needs to do the discovery then there will need to be an efficient way to retrieve a token by the generated request ID.

I think this is something that the storage is responsible for! People could use in-memory access, etcd, redis, or something else. Our goal is to make it effective and easy to use, efficiency with backend stores is not our task imho :)

> If the token passed to the request is an access token, the server MAY revoke the respective refresh token as well.

This makes a lot of sense! If the access token was compromised, the refresh token might be compromised as well, so why not remove both and be safe!

jrossiter:
> Therefore, I think that discovery should be done in **RevokeToken**.

By this did you mean discover should be done in `Revoke[Access|Refresh]Token`? 

I may have misread something; `RevokeToken` being a fosite implementation which would call `Revoke[Access|Refresh]Token` which would then do the discovery? Is this correct?

arekkas:
Ah, sorry for the misunderstanding. I think that token discovery should be done here: https://github.com/ory-am/fosite/pull/112/files#diff-c668075dab6cdbe66be88b6e77f8e7efR14

jrossiter:
Haha, no problem.

Is my understanding of `discovery` correct here? 

Discovery being the workflow where, given one token type, you'd look for the other token pair.
Eg: Refresh token with request ID 1, look for access token with request ID 1.

> I think this is something that the storage is responsible for! People could use in-memory access, etcd, redis, or something else. Our goal is to make it effective and easy to use, efficiency with backend stores is not our task imho :)

If it's up to the storage then `Revoke[Access|Refresh]Token` would be doing the discovery right?

arekkas:
> Is my understanding of discovery correct here?

Partially :D So yes, this is correct:

> Discovery being the workflow where, given one token type, you'd look for the other token pair.
Eg: Refresh token with request ID 1, look for access token with request ID 1.

With the following paragraph:

> I think this is something that the storage is responsible for! People could use in-memory access, etcd, redis, or something else. Our goal is to make it effective and easy to use, efficiency with backend stores is not our task imho :)

I meant that making this look up performant is the task of the store, so a response to:

> If RevokeToken needs to do the discovery then there will need to be an efficient way to retrieve a token by the generated request ID.

which I understood as: "how do we make this lookup fast?"

I hope this clears up the picture :)

jrossiter:
Okay that much is clear :)

What about this?

> If it's up to the storage then Revoke[Access|Refresh]Token would be doing the discovery right?

arekkas:
> If it's up to the storage then Revoke[Access|Refresh]Token would be doing the discovery right?

💯

jrossiter:
PR updated.

- Removed context from Write RevocationResponse

- Added GetID and ID fields to Requester interface

- Fixed RevokeToken logic to match RFC flow

arekkas:
Awesome, I will take a look in the afternoon!

compose/compose_oauth2.go
```diff
@@ -118,3 +118,21 @@ func OAuth2ResourceOwnerPasswordCredentialsFactory(config *Config, storage inter
...
+	}{
+		TokenRevocationHandler: &oauth2.TokenRevocationHandler{
```
arekkas: I think you can return the TokenRevocationHandler immediately, without that struct

request.go
```diff
@@ -17,6 +20,7 @@ type Request struct {
 
 func NewRequest() *Request {
 	return &Request{
+		ID:            uuid.New(),
```
arekkas: We should make sure that the `Request` object is always generated using `NewRequest()`, so the ID is always set

handler/oauth2/flow_revoke.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// RevokeToken implements https://tools.ietf.org/html/rfc7009#section-2.1
+// The token type hint indicates which token type check should be performed first.
+func (r *TokenRevocationHandler) RevokeToken(ctx context.Context, token string, tokenType fosite.TokenType) error {
+	revokeFuncs := []func(){
+		func() {
+			// Refresh token
+			signature := r.TokenRevocationStorage.RefreshTokenSignature(token)
+			r.TokenRevocationStorage.RevokeRefreshToken(ctx, signature)
```
arekkas: This is still lacking the token discovery process, right?

compose/compose_oauth2.go
jrossiter: I'll give this a try.

handler/oauth2/flow_revoke.go
jrossiter: Any suggestions?

Should there be an additional Storage function such as:

```DiscoverToken(ctx, signature, tokenType) (<something>, error)```

```go
func() {
			// Refresh token
			signature := r.TokenRevocationStorage.RefreshTokenSignature(token)
			r.TokenRevocationStorage.RevokeRefreshToken(ctx, signature)
			if something, err := r.TokenRevocationStorage.DiscoverToken(ctx, signature, tokenType); err == nil {
				r.DoThing()
			}
		},
```

Or could `Revoke[Access|Refresh]Token` accept an additional signature and the discovery takes place beforehand?

arekkas: How about:

```go
signature := r.TokenRevocationStorage.RefreshTokenSignature(token)
a, err := r.TokenRevocationStorage.GetRefreshToken(ctx, signature)
r.TokenRevocationStorage.RevokeRefreshTokens(ctx, a.GetID())
r.TokenRevocationStorage.RevokeAccessTokens(ctx, a.GetID())
```

?

jrossiter:
Added discovery, let me know if it's along the lines of what you were thinking.

arekkas:
yes! awesome! are you planning any cleanup or is this ready for final review?

jrossiter:
Final review (hopefully) :)

generate-mocks.sh
```diff
@@ -11,6 +11,7 @@ mockgen -package internal -destination internal/oauth2_explicit_storage.go githu
 mockgen -package internal -destination internal/oauth2_implicit_storage.go github.com/ory-am/fosite/handler/oauth2 ImplicitGrantStorage
 mockgen -package internal -destination internal/oauth2_owner_storage.go github.com/ory-am/fosite/handler/oauth2 ResourceOwnerPasswordCredentialsGrantStorage
 mockgen -package internal -destination internal/oauth2_refresh_storage.go github.com/ory-am/fosite/handler/oauth2 RefreshTokenGrantStorage
+mockgen -package internal -destination internal/oauth2_revoke_storage.go github.com/ory-am/fosite/handler/oauth2 TokenRevocationStorage
```
arekkas: The revoke handler mock is missing here

handler/oauth2/flow_revoke_storage.go
```diff
@@ -0,0 +1,26 @@
...
+	AccessTokenStrategy
+	AccessTokenStorage
+
+	// * https://tools.ietf.org/html/rfc7009#section-2.1
+	// If the particular
+	// token is a refresh token and the authorization server supports the
+	// revocation of access tokens, then the authorization server SHOULD
+	// also invalidate all access tokens based on the same authorization
+	// grant (see Implementation Note).
+	RevokeRefreshToken(ctx context.Context, requestID string)
```
arekkas: Please start godoc for methods with:

```
// RevokeRefreshToken revokes a refresh token [...] as specified in:
// https://tools.ietf.org/html/rfc7009#section-2.1
RevokeRefreshToken(ctx context.Context, requestID string)
```

handler/oauth2/flow_revoke_storage.go
```diff
@@ -0,0 +1,26 @@
+package oauth2
+
+import (
+	"golang.org/x/net/context"
+)
+
+type TokenRevocationStorage interface {
```
arekkas: It would be awesome if you could add a documentation to the interface as well

handler/oauth2/flow_revoke_storage.go
```diff
@@ -0,0 +1,26 @@
...
+	// revocation of access tokens, then the authorization server SHOULD
+	// also invalidate all access tokens based on the same authorization
+	// grant (see Implementation Note).
+	RevokeRefreshToken(ctx context.Context, requestID string)
+
+	// * https://tools.ietf.org/html/rfc7009#section-2.1
+	// If the token passed to the request
+	// is an access token, the server MAY revoke the respective refresh
+	// token as well.
+	RevokeAccessToken(ctx context.Context, requestID string)
```
arekkas: Please start godoc for methods with:

```
// RevokeAccessToken revokes an access token [...] as specified in:
// https://tools.ietf.org/html/rfc7009#section-2.1
RevokeAccessToken(ctx context.Context, requestID string)
```

request.go
```diff
@@ -25,6 +29,10 @@ func NewRequest() *Request {
 	}
 }
 
+func (a *Request) GetID() string {
+	return a.ID
```
arekkas: I think it would be better to set the ID here, on demand:

```
if a.ID == "" {
  a.ID = uuid.New()
}
```

request.go
arekkas: see comment below

revoke_handler.go
```diff
@@ -0,0 +1,101 @@
...
+// of the error by the authorization server as described below.
+//
+// In the next step, the authorization server invalidates the token.
+// The invalidation takes place immediately, and the token cannot be
+// used again after the revocation.
+//
+// * https://tools.ietf.org/html/rfc7009#section-2.2
+// An invalid token type hint value is ignored by the authorization
+// server and does not influence the revocation response.
+func (f *Fosite) NewRevocationRequest(ctx context.Context, r *http.Request) error {
```
arekkas: Please document methods with:

```
// NewRevocationRequest ...
func (f *Fosite) NewRevocationRequest(ctx context.Context, r *http.Request) error {
```

revoke_handler.go
```diff
@@ -0,0 +1,101 @@
...
+// * https://tools.ietf.org/html/rfc7009#section-2.2
+// The authorization server responds with HTTP status code 200 if the
+// token has been revoked successfully or if the client submitted an
+// invalid token.
+//
+// Note: invalid tokens do not cause an error response since the client
+// cannot handle such an error in a reasonable way.  Moreover, the
+// purpose of the revocation request, invalidating the particular token,
+// is already achieved.
+func (f *Fosite) WriteRevocationResponse(rw http.ResponseWriter, err error) {
```
arekkas: Please document methods with:

```
// WriteRevocationResponse ...
func (f *Fosite) WriteRevocationResponse(ctx context.Context, r *http.Request) error {
```

revoke_handler_test.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+		c.mock()
+		ctx := NewContext()
+		fosite.RevocationHandlers = c.handlers
+		err := fosite.NewRevocationRequest(ctx, r)
+		assert.True(t, errors.Cause(err) == c.expectErr, "%d\nwant: %s \ngot: %s", k, c.expectErr, err)
+		if err != nil {
+			t.Logf("Error occured: %v", err)
+		} else {
+			// todo http header codes??
```
arekkas: here is a todo :)

fosite.go
```diff
@@ -49,12 +49,27 @@ func (t *TokenValidators) Append(h TokenValidator) {
 	*t = append(*t, h)
 }
 
+// RevocationHandlers is a list of RevocationHandler
+type RevocationHandlers []RevocationHandler
+
+// Add adds an RevocationHandler to this list. Ignores duplicates based on reflect.TypeOf.
+func (t *RevocationHandlers) Append(h RevocationHandler) {
```
arekkas: This should be:

```
// Append adds an ...
```

handler.go
```diff
@@ -33,3 +33,8 @@ type TokenEndpointHandler interface {
 	// the request, this method should return ErrUnknownRequest and otherwise handle the request.
 	HandleTokenEndpointRequest(ctx context.Context, req *http.Request, requester AccessRequester) error
 }
+
+type RevocationHandler interface {
```
arekkas: It would be awesome if you could give a short description what this interface does

revoke_handler_test.go
jrossiter: Oops!

jrossiter:
Updated the sections you mentioned. I hope I didn't miss anything.

jrossiter:
I this may be caused by the mixture of storage and strategy interface embedded in `TokenRevocationStorage`. 

The example storage is also missing the key example functions so I'll need to add those in tomorrow.

jrossiter:
Okay this commit was a lot more involved :)

- Moved strategies out of TokenRevocationStorage interface
- Added revoke token options to the authorize code workflow on the example store page
- Added revoke functionality to the example store
- Generated mocks and removed old ones (I assume the project had a different structure previously which caused duplicate definitions are running ./generate_mocks.sh)

jrossiter:
Thanks!

arekkas:
@jrossiter you are now a part of the hall of fame :) ec24a6eba332b7960f207a4247e74f8b97786217

PR #113, opened 2016-10-12
oauth2: add token introspection

arekkas:
bc breaks:

* ValidateToken is now IntrospectToken
* TokenValidator is now TokenIntrospector


arekkas:
superseded by #114

PR #121, opened 2016-10-22
handler/oauth2: set JWT ExpiresAt claim per TokenType from the session

cristiangraz:
Fixes https://github.com/ory-am/fosite/issues/120. The JWT strategy now uses the `ExpiresAt` map introduced in `0.5.0` to set expirations by `fosite.TokenType`. Previously the expiration was the same for all token types based on the `JWTClaims.ExpiresAt` field.

Signed-off-by: Cristian Graziano <cristiangraz@users.noreply.github.com>


arekkas:
Awesome, thanks!

PR #126, opened 2016-11-25
Introspection endpoint responds to invalid requests appropriately

kujenga:
This PR improves the behavior of the OAuth2 Introspection endpoint to respond with a more appropriate error message when the request is invalid. Currently, fosite responds to all requests other than those that are classified as "unauthorized" with an HTTP status code of 200 and body of `{"active":"false"}`.

While the spec doesn't define what to do in the case of an invalid request (at least not that I could find), I can't think of a reason that one would want to obscure the difference between an invalid request and some other kind of error. It definitely has the potential to be confusing though when one is developing an application! We ran into this case when accidentally using `GET` instead of `POST` on this endpoint.

[1] https://tools.ietf.org/html/rfc7662#section-2.3

### DIFF
--- a/introspection_response_writer.go
+++ b/introspection_response_writer.go
@@ -7,7 +7,7 @@ import (
 	"strings"
 )
 
-// WriteIntrospectionResponse responds with token metadata discovered by token introspection as defined in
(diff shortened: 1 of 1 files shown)


arekkas:
Yup, that makes sense, thanks!

PR #127, opened 2016-11-29
Introspection responds to invalid requests appropriately

kujenga:
This PR changes the error handling behavior of the introspection handlers to return information to clients informing them when a request is invalid. Since this behavior is more expected and as far as I can tell the OAuth 2 introspection specification [1] doesn't specify what to do in the case of an invalid request, and I don't see a reason to obscure this case.

[1] https://tools.ietf.org/html/rfc7662#section-2.3

### DIFF
--- a/introspection_response_writer.go
+++ b/introspection_response_writer.go
@@ -7,7 +7,7 @@ import (
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


kujenga:
Sorry, I forgot I already submitted this 😄

arekkas:
:D

PR #128, opened 2016-11-29
token/jwt: Allow single element string arrays to be treated as strings

115100:
This commit allows `aud` to be passed in as a single element array
during consent validation on Hydra. This fixes
https://github.com/ory-am/hydra/issues/314.

Signed-off-by: Son Dinh <son.dinh@blacksquaremedia.com>


arekkas:
Thanks!

PR #133, opened 2016-12-29
oauth2/implicit: fix redirect url on error

nikita-v:
Hi!
According [rfc](https://tools.ietf.org/html/rfc6749#section-4.2.2.1) in implicit flow all error parameters should be placed to the fragment component of the redirection url.

### DIFF
--- a/authorize_error.go
+++ b/authorize_error.go
@@ -3,6 +3,7 @@ package fosite
 import (
 	"encoding/json"
 	"net/http"
+	"net/url"
 )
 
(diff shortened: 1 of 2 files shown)


arekkas:
Thanks! I think this is still missing for the various hybrid flows. We could check with `ar.GetResponseTypes().Exact("token") || len(ar.GetResponseTypes()) > 1`

nikita-v:
Fixed

arekkas:
Thanks! I'll go over it in detail by tomorrow :)

Issue #136, opened 2017-01-04
Add support for token revocation by public clients


kujenga:
It appears that the token revocation functionality currently requires a client ID and secret for all situations in order for a revocation to continue.

https://github.com/ory-am/fosite/blob/master/revoke_handler.go#L36-L49

The ability for public clients to revoke a token using just the Client ID is mentioned in the OAuth2 specification: https://tools.ietf.org/html/rfc7009#section-5

Is there a reason this isn't implemented? If not, I'll submit a PR adding support for it.

kujenga:
This issue was resolved in the references PR

PR #137, opened 2017-01-04
Public Client Token Revocation

kujenga:
Allow public clients to revoke tokens with just a Client ID.

This functionality is described in the OAuth2 spec here: https://tools.ietf.org/html/rfc7009#section-5

Relevant issue: #136

### DIFF
--- a/revoke_handler.go
+++ b/revoke_handler.go
@@ -43,9 +43,11 @@ func (f *Fosite) NewRevocationRequest(ctx context.Context, r *http.Request) erro
 		return errors.Wrap(ErrInvalidClient, err.Error())
 	}
 
-	// Enforce client authentication
-	if err := f.Hasher.Compare(client.GetHashedSecret(), []byte(clientSecret)); err != nil {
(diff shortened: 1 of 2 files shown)


arekkas:
Thanks for the PR, are you referring to this?

> The authorization server first validates the client credentials (in
   case of a confidential client) and then verifies whether the token
   was issued to the client making the revocation request.  If this
   validation fails, the request is refused and the client is informed
   of the error by the authorization server as described below.

https://tools.ietf.org/html/rfc7009#section-2.1

arekkas:
To fix the test you will need to add an expected call to `IsPublic` somewhere around [here](https://github.com/ory-am/fosite/blob/master/revoke_handler_test.go#L77).

kujenga:
Ah right, my bad! I've amended the commit with fixed test cases and an additional test for the functionality.

arekkas:
Sorry, completely forgot about this one! Thank you so much for your contribution!

kujenga:
Wonderful! No, thank you!

PR #138, opened 2017-01-07
RFC6749Error: conform to RFC 6749

pschultz:
Section 5.2 specifies the parameters for access error responses; the "error" and "error_description" parameters are misnamed ("name" and "description").

https://tools.ietf.org/html/rfc6749#section-5.2

The patch maintains backwards compatibility by marshaling both the existing fields, and the fields required by the RFC.

### DIFF
--- a/access_error_test.go
+++ b/access_error_test.go
@@ -1,12 +1,16 @@
 package fosite_test
 
 import (
+	"encoding/json"
 	"net/http"
+	"net/http/httptest"
 	"testing"
 
 	"github.com/golang/mock/gomock"
 	. "github.com/ory-am/fosite"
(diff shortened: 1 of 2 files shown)


arekkas:
Nice thanks! Do you think we should keep the old values around, or simply remove them and push 0.7?

pschultz:
I prefer to remove the fields. Keep in mind that this affects HTTP clients and may silently break them. I leave that decision to you.

arekkas:
I second that! Let's remove them. The library is after all still taged pre-release

pschultz:
I amended the commit.

arekkas:
Thank you!

Issue #140, opened 2017-01-20
Support for stateless JWT access tokens


ccbrown:
I've been trying to figure out how to use JWT access tokens statelessly, but all of the examples are actually persisting them to storage as far as I can tell. I have something that works, but I feel like this is a bit too messy and should be made easier by fosite.

Currently, I'm doing this:

* Issuing the access tokens as usual, but my `CreateAccessTokenSession` method is a no-op:

    ```go
    func (s *OAuth2Storage) CreateAccessTokenSession(ctx context.Context, signature string, request fosite.Requester) (err error) {
        return nil
    }
    ```

* In my request handling, decoding the token to the `JWTSession`:

    ```go
    token, err := strategy.Decode(response.GetAccessToken())
    assert.NoError(t, err)
    claims := token.Claims.(jwtgo.MapClaims)
    session = &oauth2.JWTSession{
        Subject: claims["sub"].(string),
    }
    ```

* In my `GetAccessTokenSession` method, building a requester with the decoded session:

    ```go
    func (s *OAuth2Storage) GetAccessTokenSession(ctx context.Context, signature string, session fosite.Session) (request fosite.Requester, err error) {
        return &fosite.Request{Session: session}, nil
    }
    ```

* Passing the decoded session to `IntrospectToken` for validation:

    ```go
    ar, err := provider.IntrospectToken(ctx, response.GetAccessToken(), fosite.AccessToken, session, "")
    assert.NoError(t, err)
    assert.Equal(t, ar.GetSession().GetSubject(), "123")
    ```

Is there a better way to do this? It seems like a pretty common need. After all, that's a very large part of the appeal of using JWT for access tokens.

Hopefully I'm just missing something silly here.

arekkas:
The idea of persisting the JWT in the store is that one is able to revoke access and refresh tokens, and also to make it harder to forge tokens, as one requires access to the database.

ccbrown:
Revocation with stateless JWTs is done with blacklists. The advantage being lower storage requirements, and the drawback being round-trips to a centralized database of revocations. A lot of people use JWTs to avoid hitting a database, and opt for short-lived tokens with small expiration times over instantaneous revocation though. That's what I want in this case.

Refresh tokens simply don't require the _access token_ to be stored. The refresh token, on the other hand, _is_ stored, along with the scope, subject, etc. I don't think refreshing is at all a factor in this issue. I'm only interested in stateless _access tokens_.

[...]

PR #141, opened 2017-01-21
handler/oauth2: allow stateless introspection of jwt access tokens

ccbrown:
Conversation started in #140. A quick draft was requested by @arekkas, which I provided there, but I figured starting a review might be a better way to talk about the code than inline code blocks and patch files. I won't be offended if this isn't consistent with the roadmap. The functionality is either going into my application code or fosite, so it's time well spent either way.

The goal here is to be able to perform validation of JWT access tokens without hitting the storage backend.

Without too much effort, we can define an introspector to do it. No public symbols or behavior are changed. Just added a few things that make this nicer.

### DIFF
--- a/authorize_helper_test.go
+++ b/authorize_helper_test.go
@@ -75,22 +75,22 @@ func TestDoesClientWhiteListRedirect(t *testing.T) {
 			isError: true,
 		},
 		{
-			client:  &DefaultClient{RedirectURIs: []string{"wta://auth"}},
-			url:     "wta://auth",
+			client:   &DefaultClient{RedirectURIs: []string{"wta://auth"}},
+			url:      "wta://auth",
 			expected: "wta://auth",
-			isError: false,
+			isError:  false,
 		},
 		{
-			client:  &DefaultClient{RedirectURIs: []string{"wta:///auth"}},
-			url:     "wta:///auth",
+			client:   &DefaultClient{RedirectURIs: []string{"wta:///auth"}},
+			url:      "wta:///auth",
 			expected: "wta:///auth",
-			isError: false,
+			isError:  false,
 		},
 		{
-			client:  &DefaultClient{RedirectURIs: []string{"wta://foo/auth"}},
-			url:     "wta://foo/auth",
+			client:   &DefaultClient{RedirectURIs: []string{"wta://foo/auth"}},
+			url:      "wta://foo/auth",
 			expected: "wta://foo/auth",
-			isError: false,
+			isError:  false,
 		},
 		{
 			client:  &DefaultClient{RedirectURIs: []string{"https://bar.com/cb"}},
@@ -131,10 +131,10 @@ func TestDoesClientWhiteListRedirect(t *testing.T) {
 }
 
 func TestIsRedirectURISecure(t *testing.T) {
-	for d, c := range []struct{
-		u string
+	for d, c := range []struct {
+		u   string
 		err bool
-	} {
+	}{
 		{u: "http://google.com", err: true},
 		{u: "https://google.com", err: false},
 		{u: "http://localhost", err: false},
--- a/compose/compose_oauth2.go
+++ b/compose/compose_oauth2.go
@@ -87,3 +87,17 @@ func OAuth2TokenIntrospectionFactory(config *Config, storage interface{}, strate
 		ScopeStrategy: fosite.HierarchicScopeStrategy,
 	}
 }
+
+// OAuth2StatelessJWTIntrospectionFactory creates an OAuth2 token introspection handler and
+// registers an access token validator. This can only be used to validate JWTs and does so
+// statelessly, meaning it uses only the data available in the JWT itself, and does not access the
+// storage implementation at all.
+//
+// Due to the stateless nature of this factory, THE BUILT-IN REVOCATION MECHANISMS WILL NOT WORK.
+// If you need revocation, you can validate JWTs statefully, using the other factories.
+func OAuth2StatelessJWTIntrospectionFactory(config *Config, storage interface{}, strategy interface{}) interface{} {
+	return &oauth2.StatelessJWTValidator{
+		JWTAccessTokenStrategy: strategy.(oauth2.JWTAccessTokenStrategy),
+		ScopeStrategy:          fosite.HierarchicScopeStrategy,
+	}
+}
--- a/handler/oauth2/introspector_jwt_test.go
+++ b/handler/oauth2/introspector_jwt_test.go
@@ -1,0 +1,128 @@
+package oauth2
+
+import (
+	"encoding/base64"
+	"strings"
+	"testing"
+
+	"github.com/ory-am/fosite"
+	"github.com/ory-am/fosite/internal"
+	"github.com/ory-am/fosite/token/jwt"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIntrospectJWT(t *testing.T) {
+	strat := &RS256JWTStrategy{
+		RS256JWTStrategy: &jwt.RS256JWTStrategy{
+			PrivateKey: internal.MustRSAKey(),
+		},
+	}
+
+	v := &StatelessJWTValidator{
+		JWTAccessTokenStrategy: strat,
+		ScopeStrategy:          fosite.HierarchicScopeStrategy,
+	}
+
+	for k, c := range []struct {
+		description string
+		token       func() string
+		expectErr   error
+		scopes      []string
+	}{
+		{
+			description: "should fail because jwt is expired",
+			token: func() string {
+				jwt := jwtExpiredCase(fosite.AccessToken)
+				token, _, err := strat.GenerateAccessToken(nil, jwt)
+				assert.NoError(t, err)
+				return token
+			},
+			expectErr: fosite.ErrTokenExpired,
+		},
+		{
+			description: "should pass because scope was granted",
+			token: func() string {
+				jwt := jwtValidCase(fosite.AccessToken)
+				jwt.GrantedScopes = []string{"foo", "bar"}
+				token, _, err := strat.GenerateAccessToken(nil, jwt)
+				assert.NoError(t, err)
+				return token
+			},
+			scopes: []string{"foo"},
+		},
+		{
+			description: "should fail because scope was not granted",
+			token: func() string {
+				jwt := jwtValidCase(fosite.AccessToken)
+				token, _, err := strat.GenerateAccessToken(nil, jwt)
+				assert.NoError(t, err)
+				return token
+			},
+			scopes:    []string{"foo"},
+			expectErr: fosite.ErrInvalidScope,
+		},
+		{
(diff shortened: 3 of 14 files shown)


handler/oauth2/strategy_jwt.go
```diff
@@ -36,66 +37,98 @@ func (h RS256JWTStrategy) AuthorizeCodeSignature(token string) string {
...
+	t, err = h.RS256JWTStrategy.Decode(token)
+
+	if err == nil {
+		err = t.Claims.Valid()
 	}
 
-	// validate the token
-	if err = t.Claims.Valid(); err != nil {
+	if err != nil {
 		if e, ok := err.(*jwtx.ValidationError); ok {
```
ccbrown: This entire block is dead code btw. I didn't change it because I was going for minimal impact and didn't want to change the behavior, but the `RS256JWTStrategy.Decode` method already does all of the validation, and it _always_ returns a new string error.

screen shot 2017-01-21 at 4 45 57 am

handler/oauth2/introspector_jwt_test.go
```diff
@@ -0,0 +1,95 @@
...
+		if err == nil {
+			assert.Equal(t, areq.GetClient().GetID(), "group0")
+			assert.Equal(t, areq.Session.GetSubject(), "peter")
+		}
+
+		t.Logf("Passed test case %d", k)
+	}
+}
+
+func BenchmarkIntrospectJWT(b *testing.B) {
```
ccbrown: There aren't any other benchmarks, so maybe performance is a non-goal of the library, but this was of interest to me. Understood if you want to delete in the interest of staying slim. It's certainly not useful if no one ever runs it.

arekkas:
Thanks for the PR!

I strongly dislike that scopes are currently not supported - it's a core OAuth2 concept and not using it is an anti-pattern, so it shouldn't be encouraged by the code base!

Apart from that, I understand the case now better. One issue I have is that token recovation does not have an effect when the JWT strategy is used. This could be very misleading and mount up to a security issue when the library is configured incorrectly by a developer.

Don't be mistaken though, I would like to support statelessness, I think it's pretty cool.

Also, I think it would be essential to have some sort spec test for the validation strategies - so something that tests the different implementations against the same test data. Maybe you could refactor existing [integration tests](https://github.com/ory-am/fosite/blob/master/integration/introspect_token_test.go) to make that work.

handler/oauth2/strategy_jwt.go
arekkas: Ah right, this should be `errors.Cause(err)` instead

authorize_helper_test.go
```diff
@@ -75,22 +75,22 @@ func TestDoesClientWhiteListRedirect(t *testing.T) {
 			isError: true,
```
ccbrown: Some random `go fmt` changes in this file.

compose/compose_oauth2.go
```diff
@@ -87,3 +87,17 @@ func OAuth2TokenIntrospectionFactory(config *Config, storage interface{}, strate
...
 }
+
+// OAuth2StatelessJWTIntrospectionFactory creates an OAuth2 token introspection handler and
+// registers an access token validator. This can only be used to validate JWTs and does so
+// statelessly, meaning it uses only the data available in the JWT itself, and does not access the
+// storage implementation at all.
+//
+// Due to the stateless nature of this factory, THE BUILT-IN REVOCATION MECHANISMS WILL NOT WORK.
+// If you need revocation, you can validate JWTs statefully, using the other factories.
+func OAuth2StatelessJWTIntrospectionFactory(config *Config, storage interface{}, strategy interface{}) interface{} {
```
ccbrown: Definitely agreed that there's an increased risk of misconfiguration. Hoping renaming this to emphasize the statelessness and adding a cautionary doc block will minimize that risk.

handler/oauth2/strategy_jwt.go
```diff
@@ -105,7 +142,25 @@ func (h *RS256JWTStrategy) generate(tokenType fosite.TokenType, requester fosite
...
+			claims.ExpiresAt = jwtSession.GetExpiresAt(tokenType)
+		}
+
+		if claims.IssuedAt.IsZero() {
+			claims.IssuedAt = time.Now()
+		}
+
+		if claims.Issuer == "" {
+			claims.Issuer = h.Issuer
+		}
```
ccbrown: There seemed to be a lot of cases where these things would be left empty, making it difficult to use JWTs interchangeably in tests.

Noticed the OpenID strategy had this Issuer field, which I added here just because it's handy.

token/jwt/claims_jwt.go
```diff
@@ -31,12 +32,84 @@ func (c *JWTClaims) ToMap() map[string]interface{} {
 	ret["sub"] = c.Subject
 	ret["iss"] = c.Issuer
 	ret["aud"] = c.Audience
-	ret["iat"] = float64(c.IssuedAt.Unix())  // jwt-go does not support int64 as datatype
-	ret["nbf"] = float64(c.NotBefore.Unix()) // jwt-go does not support int64 as datatype
+
+	if !c.IssuedAt.IsZero() {
+		ret["iat"] = float64(c.IssuedAt.Unix()) // jwt-go does not support int64 as datatype
```
ccbrown: Filtering out values of `-62135596800` from JWTs for the optional fields.

ccbrown:
I had to rewrite the history for those last two commits since I just realized my email address in the "signed-off-by" line was wrong. Hope that doesn't make looking at this too weird.

arekkas:
Yup, the `scp` claim is what I'd be doing as well!

arekkas:
How about a recovation strategy that comes with this introspection strategy and that always returns an error on revocation? Then combine them in a composer wrapper. Could that be a solution?

ccbrown:
I did consider making it a runtime error by doing that or something similar, but I think there may be valid use-cases for custom revocation handlers with this introspector. Not really sure if that's a common enough need to consider it though. (Personally, I don't have that need.)

arekkas:
Ok, if we don't find a solution I'd probably accept the refactoring required for your implementation to work, but would not include the implementation itself. You could provide a package `fosite-stateless-validation` and we'd link it in the docs. Then, in the readme, you could point out that token revocation won't work with this code base. What do you think about this? I think it's a fair trade off between your specific use case, a benefit for the open source eco system, and a way to keep the concepts within fosite unisono and repeatable

ccbrown:
Sounds fair to me. 👍 

I'll see if something good comes to mind, and either update the PR with something or remove the `StatelessJWTValidator` struct.

ccbrown:
Went with the return-an-error-on-revocation idea. I think the valid use-cases for revocation with this type of JWT validation would require writing entirely new custom handlers anyways.

handler/oauth2/introspector_jwt_test.go
arekkas: 👍

handler/oauth2/strategy_jwt.go
```diff
@@ -36,66 +38,99 @@ func (h RS256JWTStrategy) AuthorizeCodeSignature(token string) string {
...
 
-	// validate the token
-	if err = t.Claims.Valid(); err != nil {
-		if e, ok := err.(*jwtx.ValidationError); ok {
+	if err != nil {
+		if e, ok := errors.Cause(err).(*jwtx.ValidationError); ok {
 			switch e.Errors {
 			case jwtx.ValidationErrorMalformed:
-				return errors.Wrap(fosite.ErrInvalidTokenFormat, err.Error())
+				err = errors.Wrap(fosite.ErrInvalidTokenFormat, err.Error())
```
arekkas: I'm not sure right now, does golang fallthrough or break per default?

handler/oauth2/strategy_jwt.go
```diff
@@ -105,7 +140,21 @@ func (h *RS256JWTStrategy) generate(tokenType fosite.TokenType, requester fosite
 		return "", "", errors.New("GetTokenClaims() must not be nil")
 	} else {
 		claims := jwtSession.GetJWTClaims()
-		claims.ExpiresAt = jwtSession.GetExpiresAt(tokenType)
+
+		if claims.ExpiresAt.IsZero() {
+			claims.ExpiresAt = jwtSession.GetExpiresAt(tokenType)
```
arekkas: The idea here was to explicitly override the expires_at value. What's the reasoning to move it into `IsZero`?

integration/helper_endpoints_test.go
```diff
@@ -113,33 +114,33 @@ func authCallbackHandler(t *testing.T) func(rw http.ResponseWriter, req *http.Re
...
 }
 
-func tokenEndpointHandler(t *testing.T, oauth2 fosite.OAuth2Provider) func(rw http.ResponseWriter, req *http.Request) {
+func tokenEndpointHandler(t *testing.T, provider fosite.OAuth2Provider) func(rw http.ResponseWriter, req *http.Request) {
 	return func(rw http.ResponseWriter, req *http.Request) {
 		req.ParseForm()
 		ctx := fosite.NewContext()
 
-		accessRequest, err := oauth2.NewAccessRequest(ctx, req, &fosite.DefaultSession{})
+		accessRequest, err := provider.NewAccessRequest(ctx, req, &oauth2.JWTSession{})
```
arekkas: Why is this no longer the default Session?

handler/oauth2/strategy_jwt.go
ccbrown: It definitely feels weird coming from other languages that just fall through, but Go breaks if you don't add a fallthrough statement: https://golang.org/ref/spec#Switch_statements

arekkas:
Thanks, that looks pretty solid, I love that we have scopes in JWTs now!

I've added a few comments. We would also need some clarification in the docs that the JWT Stateless Strategy isn't supporting Token Revocation, and also a compose factory that sets up the stateless strategy appropriately

handler/oauth2/strategy_jwt.go
ccbrown: Hmm. I don't remember if this was intentional or if it was meant to be temporary while I was testing... Doesn't have any impact on tests, so I'd better change it back.

integration/helper_endpoints_test.go
ccbrown: It has to be `JWTSession` for `TestIntrospectToken` to be able to test any of the JWT strategies.

ccbrown:
I think I have those things... The compose factory would be `compose.OAuth2StatelessJWTIntrospectionFactory` and that and `StatelessJWTValidator` both document the lack of revocation. Did you have something else in mind?

arekkas:
Sorry, I probably simply overlooked it! Will review it again later today or latest tomorrow

arekkas:
Also, can you verify that your implementation works as expected with the example: https://github.com/ory-am/fosite-example ? :)

ccbrown:
Yep. Works fine as far as I can tell.

arekkas:
Sorry, I completely forgot about this issue. I will get to it next week!

arekkas:

Issue #152, opened 2017-03-23
exp (ExpiresAt) not set with JWT strategy and Resource Owner Password Credentials grant


wezzle:
I'm using the JWT strategy (compose.NewOAuth2JWTStrategy) in combination with the Resource Owner Password Credentials grant (compose.OAuth2ResourceOwnerPasswordCredentialsFactory) and i've noticed that the exp value is not correctly set on the JWT token.

When using the Client Credentials grant the exp value is correctly set. In the ClientCredentialsGrantHandler's `HandleTokenEndpointRequest` there is a line setting the expires at on the session: `request.GetSession().SetExpiresAt(fosite.AccessToken, time.Now().Add(c.AccessTokenLifespan))`. This is not done in the ResourceOwnerPasswordCredentialsGrantHandler, however adding it there will set the correct exp value on the JWT token.

Should the ResourceOwnerPasswordCredentialsGrantHandler set the expires at on the session as well or is there a reason it's not done there? Isn't this something that should be done on every handler generating tokens?

arekkas:
Yes, this is definitely a bug, thanks for spotting it. Would you mind creating a PR and also adding a check to the tests?

wezzle:
I've sent in a pull request fixing above issue.

However when using the offline scope and receiving the refresh_token the exp is not set either. I could send in another pull request fixing that issue. I think the best way to solve that is to add `RefreshTokenLifespan` to the compose/config and modifying compose/compose_oauth2.go pass it to the `OAuth2ResourceOwnerPasswordCredentialsFactory`. Is that the correct way to go?

arekkas:
Yes, that should work! Thank you for this

wezzle:
Ok, i'll get on this as soon as possible and i'll submit a new pull request.

wezzle:
I've updated the refresh token expiration into above pull request as i believe it's an extension of this issue. Please review the pull request and let me know if you think anything should be changed.

MatthewHartstonge:
Not sure if related, but I've noticed that the `IssueImplicitAccessToken` handler always seems to default to `&expires_in=3600`  even when `exp` is set differently. In this case, I've set mine to 5 minutes for dev.
If I comment out [flow_authorize_implicit line 67](https://github.com/ory/fosite/blob/master/handler/oauth2/flow_authorize_implicit.go#L67) it generates the correct number (300 compared to 3600)

arekkas:
superseded by #180

PR #153, opened 2017-03-31
exp not set with JWT strategy and Password grant

wezzle:
Closes [152](https://github.com/ory/fosite/issues/152)

I've updated the tests for flow_resource_owner and used the flow_client_credentials as an example.

### DIFF
--- a/handler/oauth2/flow_resource_owner_test.go
+++ b/handler/oauth2/flow_resource_owner_test.go
@@ -16,9 +16,10 @@ import (
 func TestResourceOwnerFlow_HandleTokenEndpointRequest(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	store := internal.NewMockResourceOwnerPasswordCredentialsGrantStorage(ctrl)
+	areq := internal.NewMockAccessRequester(ctrl)
 	defer ctrl.Finish()
 
-	areq := fosite.NewAccessRequest(nil)
+	//areq := fosite.NewAccessRequest(nil)
 	httpreq := &http.Request{PostForm: url.Values{}}
 
 	h := ResourceOwnerPasswordCredentialsGrantHandler{
@@ -38,16 +39,19 @@ func TestResourceOwnerFlow_HandleTokenEndpointRequest(t *testing.T) {
 			description: "should fail because not responsible",
 			expectErr:   fosite.ErrUnknownRequest,
 			setup: func() {
-				areq.GrantTypes = fosite.Arguments{"123"}
+				areq.EXPECT().GetGrantTypes().Return(fosite.Arguments{""})
 			},
 		},
 		{
 			description: "should fail because because invalid credentials",
 			setup: func() {
-				areq.GrantTypes = fosite.Arguments{"password"}
-				areq.Client = &fosite.DefaultClient{GrantTypes: fosite.Arguments{"password"}}
 				httpreq.PostForm.Set("username", "peter")
 				httpreq.PostForm.Set("password", "pan")
+
+				areq.EXPECT().GetGrantTypes().Return(fosite.Arguments{"password"})
+				areq.EXPECT().GetClient().Return(&fosite.DefaultClient{
+					GrantTypes: fosite.Arguments{"password"},
+				})
 				store.EXPECT().Authenticate(nil, "peter", "pan").Return(fosite.ErrNotFound)
 			},
 			expectErr: fosite.ErrInvalidRequest,
@@ -55,6 +59,13 @@ func TestResourceOwnerFlow_HandleTokenEndpointRequest(t *testing.T) {
 		{
 			description: "should fail because because error on lookup",
 			setup: func() {
+				httpreq.PostForm.Set("username", "peter")
+				httpreq.PostForm.Set("password", "pan")
+
+				areq.EXPECT().GetGrantTypes().Return(fosite.Arguments{"password"})
+				areq.EXPECT().GetClient().Return(&fosite.DefaultClient{
+					GrantTypes: fosite.Arguments{"password"},
+				})
 				store.EXPECT().Authenticate(nil, "peter", "pan").Return(errors.New(""))
 			},
 			expectErr: fosite.ErrServerError,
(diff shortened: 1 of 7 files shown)


handler/oauth2/flow_resource_owner_test.go
```diff
@@ -38,31 +38,54 @@ func TestResourceOwnerFlow_HandleTokenEndpointRequest(t *testing.T) {
 			description: "should fail because not responsible",
 			expectErr:   fosite.ErrUnknownRequest,
 			setup: func() {
-				areq.GrantTypes = fosite.Arguments{"123"}
+				areq.EXPECT().GetGrantTypes().Return(fosite.Arguments{""})
```
wezzle: No reason, I've copied the setup from the flow_client_credentials tests, thinking it would be better if they used the same checks. I can change it back to 123 but thought the test would need to fail either way. Let me know what you think is best.

arekkas: Let's add "123" again, it makes it clear that a wrong value causes skipping this

wezzle: I've reinserted it back into the test for the resource owner flow. Please review the latest changes and let me know if it's ok like this.

arekkas:
Sorry, I was really busy these last days so I wasn't able to review the code. I remembered something though, I don't think refresh tokens have a life span usually. This is also the reason why the HMAC strategy does not check for the token expiry. I will need to dig through the OAuth2 spec again.

arekkas:
Instead, refresh tokens are one time passwords (vs. time-based one time passwords) - so you get to use them once but it doesn't matter when. Also the current implementation in this PR would be inconsistent because the HMAC startegy is not checking for the refresh token lifetime.

wezzle:
With no expiry on refresh_tokens and the JWTStrategy the oauth server can never revoke a refresh_token and the refresh_token that's given out will be forever valid, seems like a security issue to me?

With HMAC and revoking, leaving the refresh_token valid indefinitely seems like a valid point depending on your use-case. However every implementation i've seen so far does expire refresh tokens usually within periods of a month or 2. I'll look into the spec as well but there should be an option to have a lifespan on refresh tokens in my opinion

arekkas:
Yeah that's some good thinking on JWT strategy. If you find anything in the spec re that post it here :)

I mean it would make sense to limit the time of a refresh token instead of depending on explicit revokation.

wezzle:
Couldn't find anything in the spec about limiting refresh token lifespan. They are supposed to be non expiring. Are you open to an implementation with a default RefreshTokenLifespan of "infinite" and a possibility to specify your own lifespan?

I think that a lot of people could benefit from such an addition. That would mean fosite goes a step further than the spec.

wezzle:
Just to jump back why I implemented RefreshTokenLifespan in the first place:

When using JWTStrategy with RefreshTokenGrant the JWT created as a refresh_token contains the "exp" field but it is set to a negative value (because no expiry is set on the session).

Eventually every token, also in the RefreshTokenGrantHandler, runs through [RS256JWTStrategy.Validate](https://github.com/ory/fosite/blob/14a18a714c419b31d4bf1341e1017159bc17540f/handler/oauth2/strategy_jwt.go#L97) and then through [t.Claims.Valid()](https://github.com/ory/fosite/blob/14a18a714c419b31d4bf1341e1017159bc17540f/handler/oauth2/strategy_jwt.go#L101). The [documentation](https://godoc.org/github.com/dgrijalva/jwt-go#MapClaims.Valid) of that function says the following: `Validates time based claims "exp, iat, nbf". ... As well, if any of the above claims are not in the token, it will still be considered a valid claim.`.

So in order to fix the JWT refresh_token in it's current state it should not have the exp field. Correct?

arekkas:
I slept over this, and I believe you are on the right track here. We should have an expiry for refresh tokens, and we could per default set it to `0` which would mean no expiry. A JWT without expiry date would basically be the highest possible unix timestamp.

What do you think?

wezzle:
Yes i think that would be the way to go aswel. The [JWT Token specification](https://tools.ietf.org/html/rfc7523) does not describe refresh tokens so i guess there is no "correct" way of doing anything. 

I'm in doubt about setting the "exp" to "highest possible unix timestamp", that seems a bit ugly to me. Not setting the "exp" value would seem like a nicer approach? That way we only have to edit the token creation functionality. If RefreshTokenLifespan is set on the config we set the exp value and if not just leaving it out, causing the t.Claims.Valid() to work as is.

arekkas:
> The JWT MUST contain an "exp" (expiration time) claim that
        limits the time window during which the JWT can be used.  The
        authorization server MUST reject any JWT with an expiration time
        that has passed, subject to allowable clock skew between
        systems.  Note that the authorization server may reject JWTs
        with an "exp" claim value that is unreasonably far in the
        future.

- https://tools.ietf.org/html/rfc7523#section-3

If the refresh token is a JWT, it should contain the expiry. Another approach would be to issue JWT access tokens but hmac-sha refresh tokens. Which (when looking at this) actually makes a lot of sense. The refresh token is never used by the client except for exchanging it for an access token so there is no need to read it, right? resource providers also don't deal with refresh tokens

wezzle:
Yes that would be even better, i see no reason for using JWT tokens as refresh tokens either. In my own implementation using fosite i've rewritten the handlers to inject the hmac strategy for refresh tokens and JWT for access tokens.

However using JWT for refresh tokens is still broken in the current implementation. If we want to fix that we should go for either dropping the exp (i know the JWT spec says it should be there, but the spec only describes: "This specification defines the use of a JSON Web Token (JWT) **Bearer
   Token**") or set the exp to a super high unix timestamp.

arekkas:
Ok, let's drop the exp in the refresh token if no expiry time is set, and add it if one is set.

Also, I would be ok to - per default - use HMAC for refresh tokens with JWT!

wezzle:
I'll get on this as soon as possible. I'll let you know what changes and additions i have planned to make sure we're thinking in the same direction.

arekkas:
Let me know if you need any help. You can jump on our gitter channel (see readme) if you need shorter iterations

arekkas:
Any updates on this @wezzle ?

arekkas:
I'm closing this PR due to inactivity, if you find some free time to keep working on this let me know and I'll re-open it :) Thank you for your effort so far!

This is now tracked as #180

PR #156, opened 2017-04-10
Get the token from the access_token query parameter

janekolszak:
As discussed on Gitter some time ago - fosite needs to use _access_token_ query param. (https://tools.ietf.org/html/rfc6750)

### DIFF
--- a/introspect.go
+++ b/introspect.go
@@ -13,10 +13,20 @@ type TokenIntrospector interface {
 }
 
 func AccessTokenFromRequest(req *http.Request) string {
+	// Acording to https://tools.ietf.org/html/rfc6750 you can pass tokens through:
+	// - Form-Encoded Body Parameter. Recomended, more likely to appear. e.g.: Authorization: Bearer mytoken123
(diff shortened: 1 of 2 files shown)


arekkas:

Unfortunately, the tests are failing. Most likely because `req.URL.Query()` has not been mocked properly in the tests. While you're add it, could you please add a test for query-based access token verification as well?

Also, could you please sign our [DOC](https://github.com/ory/fosite/blob/master/CONTRIBUTING.md#developers-certificate-of-origin) :)

arekkas:
Let me know when you're ready for review :) Stupid coveralls...

janekolszak:
It's ready.
I was lazy and tried to use CI for development. In the end had to setup test on my local env. :smi

I'm not sure about `ParseForm`. It seems to parse more than needed, but I can't find anything that parses only URL.

arekkas:
Perfect, thank you! Now one last thing, could you please sign the DOC with `git commit --amend -s`? :)

janekolszak:
Done

arekkas:
Thank you!

PR #159, opened 2017-05-07
Add key id to header

Grillz:
Adding key id to the header of the id token, per https://github.com/ory/hydra/issues/433.

This implies that the most recent active key is always named 'public'.

Signed-off-by: pbarker <pbarker@datapipe.com>

### DIFF
--- a/token/jwt/jwt.go
+++ b/token/jwt/jwt.go
@@ -25,6 +25,8 @@ func (j *RS256JWTStrategy) Generate(claims jwt.Claims, header Mapper) (string, s
 
 	token := jwt.NewWithClaims(jwt.SigningMethodRS256, claims)
 	token.Header = assign(token.Header, header.ToMap())
+	// "public" is understood to be the most recent keyid
+	token.Header["kid"] = "public"
 
 	var sig, sstr string
 	var err error


token/jwt/jwt.go
```diff
@@ -25,6 +25,8 @@ func (j *RS256JWTStrategy) Generate(claims jwt.Claims, header Mapper) (string, s
 
 	token := jwt.NewWithClaims(jwt.SigningMethodRS256, claims)
 	token.Header = assign(token.Header, header.ToMap())
+	// "public" is understood to be the most recent keyid
+	token.Header["kid"] = "public"
```
arekkas: This is actually implicit knowledge from hydra, not all keys have a key id of "public". Instead, we should set the key in the header.

Grillz: So its interesting because the RSA key that is passed into the factory https://github.com/ory/hydra/blob/master/cmd/server/handler_oauth2_factory.go#L66 doesn't have the notion of an ID. I think the token header needs to be added here because it needs to be there before its signed. 

We would likely need to modify the https://github.com/ory/fosite/blob/master/compose/compose_strategy.go#L35 to accept a JWT rather than an RSA key, but that may have a cascading effect

arekkas: It should be enough to add that header field in [`NewSession()`](https://github.com/ory/hydra/blob/master/oauth2/session.go#L21)

arekkas:
Closing becuase this is merged in hydra now

PR #160, opened 2017-05-15
removed the requirement to *http.Request for endpoints and response o…

michaelboke:
While implementing and reviewing the fosite code base, i found a few points of improvements. Primarily on the http.Request who needs to be provided to the endpoints and response structures. The request handlers do extract the needed url params and store them inside the fosite/request.Form structure. So there is no need pass the http.Request object to the endpoints and response constructors. They can refer to the fosite/Request.GetRequestForm for getting the url values.

I had difficulties implementing fosite in frameworks like go-kit and creating different (un)marshallers 
for different protocols (like protobuf). I was always required to create a fake http.Request object and pass them around.

Also i found a few PR's who discussed and introduced the fosite/Request object for the replacement of the http.Request object.

- I modified the endpoints and response signatures and removed the `*http.Request` requirement for the constructor.
- In the endpoints and responses where the code was still referring to the `*http.Request` changed to `fosite.Request.GetRequestForm()`
- Checked and modified all the test so they can work with the new signature. 
- Also found some test referring to `ory-am`

I hope you guys appreciate my modification and make fosite better and simpler. I'm already liking it.
Please let me know what you think

Signed-off-by: Michael Boke <michael.boke@gmail.com>


arekkas:
Thank you for your contribution! Every PR improves fosite :)

As you can imagine and as it's very visible in the PR, there are several breaking changes introduced and several things addressed, which makes code review much more difficult and also very hard to take into account the work people have to do who rely on fosite.

Thus I'd like to ask you to split your PR into multiple ones, each one addressing one of the changes. That way I can review this much easier and also accept some of the changes whilst (maybe) rejecting others.

michaelboke:
At first sight it really does look like a lot.
But it comes down to 4 interfaces i changed (there i removed http.Request argument).

Let me summarise the changes

The following interfaces are changed:
handler.go
- interface AuthorizeEndpointHandler
- interface TokenEndpointHandler
oauth2.go
- interface OAuth2Provider
handler/openid/strategy.go
- interface OpenIDConnectTokenStrategy		

As a result of this i needed to remove all the extra arguments in the test and needed to regenerate the mocks. (27 files autogenerated)

3 Flows needed a change  to use fossit.Request (actual small change in implementation):
- flow_authorize_code_token.go
- flow_refresh.go
- flow_resource_owner.go

For the tests i needed to fix some references from ory-am to ory, also add a missing test dependency in the glide files. 

This is really all of the changes, as you can see its a small change who affects a lot of method signatures and tests.
I do not think splitting this would make it anything better.

I hope this makes it more clear

michaelboke:
To clearify the change to implementation i also updated the fosite-example repo.
Check my  PR (https://github.com/ory/fosite-example/pull/4/files)

So i hope i made it really easy for you now ;)

arekkas:
Awesome, thanks! That change definitely makes sense. Did you find other pieces where an interface change would make sense as well? That way we could break things only one time and make the API cleaner :)

michaelboke:
Not really, i have to say the overall design of the interfaces is pretty clean and straight forward.
This was the only thing that directly caught my attention.

internal/authorize_code_storage.go
```diff
@@ -1,11 +1,10 @@
 // Automatically generated by MockGen. DO NOT EDIT!
-// Source: github.com/ory-am/fosite/handler/oauth2 (interfaces: AuthorizeCodeStorage)
+// Source: github.com/ory/fosite/handler/oauth2 (interfaces: AuthorizeCodeStorage)
 
 package internal
 
 import (
 	context "context"
-
```
michaelboke: thats correct i'm more used to run go fmt. I updated the files with goimports and added the command to the generate-mocks.sh file

arekkas:
Thanks!

Issue #161, opened 2017-05-16
All storage interfaces have a Context except for GetClient


michaelboke:
All the storage interfaces have a Context except for the Storage.GetClient (fosite.Storage / fosite.ClientManager).

function signature ```GetClient(id string) (fosite.Client, error)```

And what is the deal with the separate ClientManager interface?
Seems like a strange duck while there is always a Storage interface defined.

Is this an candidate for a refactor? 
```GetClient(ctx context.Context, id string) (fosite.Client, error)```

Let me know i could do the refactor for you

arekkas:
The client manager is it's own interface so it's easier to compose stores later on. However, adding the context here could be benefitial for some cases.

On the other hand, the context is never used anywhere in the library. I'm not sure if we should keep it at all. What do you think?

michaelboke:
I came across this while implementing the storages with a sql backend. Since go 1.8 the DB has 
 context support for cancelable queries. I could now implement that for all except the client storage.

I would say atleast add it to interface to keep it in line with all the other storage interfaces.

And the ClientManager interface does not really describe a manager functions, it still looks like a storage interface. There is just one getter inside.

arekkas:
Yeah right, I forgot that go 1.8 added that - so let's keep the contexts and I'd also appreciate an update to the GetClient method with context :)

PR #162, opened 2017-05-16
added context to GetClient storage interface, see issue: ory/fosite#161

michaelboke:
added context to client storage GetClient signature.

See issue ory/fosite#161


michaelboke:
👍  great and thanks

PR #163, opened 2017-05-17
Remove forced nonce

grillz:
Hey @arekkas, as you know I'm working on integrating Hydra with Kubernetes' RBAC system. Both Kubernetes and Dex aren't using a nonce when requesting an ID token. I know thats the recommended way of doing it, but its not required and may be substantially more difficult to push these things through upstream.

With this piece unblocked Hydra would be fully compatible with Kubernetes!

Signed-off-by: pbarker <pbarker@datapipe.com>

### DIFF
--- a/handler/openid/strategy_jwt.go
+++ b/handler/openid/strategy_jwt.go
@@ -2,6 +2,7 @@ package openid
 
 import (
 	"encoding/gob"
+	"fmt"
 	"net/http"
 	"time"
 
@@ -143,10 +144,9 @@ func (h DefaultStrategy) GenerateIDToken(_ context.Context, _ *http.Request, req
 
 	nonce := requester.GetRequestForm().Get("nonce")
 	// OPTIONAL. String value used to associate a Client session with an ID Token, and to mitigate replay attacks.
-	// Although optional, this is considered good practice and therefore enforced.
 	if len(nonce) < fosite.MinParameterEntropy {
 		// We're assuming that using less then 8 characters for the state can not be considered "unguessable"
-		return "", errors.WithStack(fosite.ErrInsufficientEntropy)
+		fmt.Println("WARNING: nonce token is missing or contains insufficient entropy")
 	}
 
 	claims.Nonce = nonce


handler/openid/strategy_jwt.go
```diff
@@ -143,10 +144,9 @@ func (h DefaultStrategy) GenerateIDToken(_ context.Context, _ *http.Request, req
 
 	nonce := requester.GetRequestForm().Get("nonce")
 	// OPTIONAL. String value used to associate a Client session with an ID Token, and to mitigate replay attacks.
-	// Although optional, this is considered good practice and therefore enforced.
 	if len(nonce) < fosite.MinParameterEntropy {
 		// We're assuming that using less then 8 characters for the state can not be considered "unguessable"
-		return "", errors.WithStack(fosite.ErrInsufficientEntropy)
+		fmt.Println("WARNING: nonce token is missing or contains insufficient entropy")
```
arekkas: Instead of this, the check should be:

```
if len(nonce) == 0 {
  // skip
} else if len(nonce) < fosite.MinParameterEntropy {
  return "", errors.WithStack(fosite.ErrInsufficientEntropy)
}
```

Grillz:
Sounds good, I'll lean on them upstream and we'll see how it goes 👍

wyattanderson:
This is causing some pain for me; I don't think there's any way to send a nonce with a token request when using the [golang `oauth2` package](https://godoc.org/golang.org/x/oauth2). The [`Exchange` method](https://godoc.org/golang.org/x/oauth2#Config.Exchange) doesn't take any parameters.

Seems as if the nonce is only required for the implicit flow; is there a way to only enforce its presence during the implicit flow?

arekkas:
@wyattanderson `oauth2.GetAuthCodeURL() + "&nonce=1234"`

arekkas:
Any news on that front? Otherwise I'd be good with merging this.

Grillz:
Sorry @arekkas, was getting married 😄 I'll work on it this weekend.

arekkas:
Awesome! Congratulations :)

wyattanderson:
Ran into this again in a different situation; this time in code that I can't easily modify to add the nonce (hooking up an older version of Dex to talk to Hydra as an OIDC connector).

arekkas:
see #185

PR #164, opened 2017-05-18
no expire date is set to session when storing token to storage.

michaelboke:
There is no expire date set on implicit token flows.
The expire date is updated to the session after the store function is called.


arekkas:
Thank you!

PR #166, opened 2017-05-18
authorize code expire time isset after persist

michaelboke:
Authorize code expire time is set after the calling CreateAuthorizeCodeSession.
This way i cannot store the expire time in the database.

This is the same kinda issue as with the implicit code session expire time (other PR)


arekkas:
Thank you!

PR #170, opened 2017-05-22
Enable fosite composing with custom hashers.

MatthewHartstonge:
Makes it a little easier for composing, plus follows the same conventions as storage. Seeks to keep compatibility by setting BCrypt as the default hasher if one isn't set.

Signed-off-by: Matthew Hartstonge <matt@mykro.co.nz>

### DIFF
--- a/compose/compose.go
+++ b/compose/compose.go
@@ -30,14 +30,17 @@ type Factory func(config *Config, storage interface{}, strategy interface{}) int
 //  )
 //
 // Compose makes use of interface{} types in order to be able to handle a all types of stores, strategies and handlers.
-func Compose(config *Config, storage interface{}, strategy interface{}, factories ...Factory) fosite.OAuth2Provider {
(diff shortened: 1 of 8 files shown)


compose/compose.go
```diff
@@ -30,14 +30,17 @@ type Factory func(config *Config, storage interface{}, strategy interface{}) int
 //  )
 //
 // Compose makes use of interface{} types in order to be able to handle a all types of stores, strategies and handlers.
-func Compose(config *Config, storage interface{}, strategy interface{}, factories ...Factory) fosite.OAuth2Provider {
+func Compose(config *Config, storage interface{}, strategy interface{}, hasher interface{}, factories ...Factory) fosite.OAuth2Provider {
```
arekkas: Why is hasher an interface here? It's not being used apart from type asserting `hasher.(fosite.Hasher)`

MatthewHartstonge: Good question! I was blindly following the storage implementation. Fixed with latest commit.

arekkas:
note to self: add this to breaking changes in history.md

Issue #165, opened 2017-05-18
ImplicitGrantStorage never used.


michaelboke:
the ImplicitGrantStorage in handler/oauth/flow_authorize_implicit_storage.go is never used.
The flow_authorize_implicit.go hints in comments it should be using this interface but the `AccessTokenStorage` is used instead.

I do recommend to remove the interface `ImplicitGrantStorage` and  update the comments in flow_authorize_implicit.go that AccessTokenStorage is used to persist the AccessToken

arekkas:
👍

PR #171, opened 2017-05-22
removed implicit storage as its never used, 

michaelboke:
see issue #165, removed the unused storage interface


arekkas:
Thanks!

Issue #169, opened 2017-05-21
Introspect token


michaelboke:
There is no token_type property when introspecting token.
Now i cannot distinguish the introspect token result which type of token it is.

Also the authorize code is introspected and returned as a result. This is probably not the most ideal situation because now a authorize_code could be seen as a valid token.

I suggest dropping introspection of authorize codes and add the token_type to the introspect result

arekkas:
The token type is always bearer, it does not distinguish between refresh and access tokens. See [here](https://tools.ietf.org/html/rfc6749#section-7.1)  for more information.

arekkas:
see PR

michaelboke:
Thanks for your comments, but i still have no idea how to distinguish the token types in the introspect endpoint. You have any idea's how i can ?

arekkas:
You receive the tokens types when performing the auth code flow, so you should know which token is what, right?

PR #172, opened 2017-05-22
get token types in introspect response

michaelboke:
Based on issue #169 i made a little setup to return the token types based on the session values.
It does not work for refresh token but tested it works for access token and authorization codes for now.

I'm not totally comfortable with the hack i used to get the information. I (ab)used the expire time in the session. It works as long as there is an expire time set for the token type. (and for refresh tokens its nil)

Probably we need to store it directly in the session.

What is your opinion?


arekkas:
Closing, if you see this differently let me know

michaelboke:
i totally agree. spitting through the rfc's I also found this.

PR #177, opened 2017-06-07
handler/oauth2: grant scopes before the access token is generated

nikita-v:
Hi!
JWT access token does not contain any granted scopes in "scp" claim. This is because the access token is generated before the scopes are granted.


arekkas:
Thank you!

PR #183, opened 2017-06-20
Update test command in README and CONTRIBUTING

wyattanderson:
Running `go test ./...` fails, even with everything installed. Travis seems to be running the equivalent of `go test $(glide novendor)` and this is what Hydra is doing, so I'm assuming this is more correct.


arekkas:
ty

PR #186, opened 2017-07-04
vendor: separate test dependencies

dackroyd:
* vendor: Move testify to testImport
* test: Move Assert/Require helpers to _test pkg

Signed-off-by: David Ackroyd <daveo.ackroyd@gmail.com>


dackroyd:
With the dependency declared in the main imports, and used by 'non test' code (even if these helpers are only used by tests), users of the library must declare Testify as a regular import rather than a `testImport`, otherwise the following error will occur during a `glide update`:

```
[ERROR] Failed to generate lock file: Generating lock produced conflicting versions of github.com/stretchr/testify. import (~1.1.4), testImport (^1.1.4)
```

Since Testify is a _testing_ library, it should not be declared/used as a non-test import

arekkas:
Thank you

PR #205, opened 2017-07-27
Fixing documentation oauth2 variable + old method

akuma06:
It seems that the documentation was declaring as OAuth2Provider the variable `oauth2Provider` whereas it used a non-declared variable `oauth2`. I renamed `oauth2` into the variable declared `oauth2Provider`.

Furthermore, on line 333, the IntrospectToken method was called without the TokenType argument. I added the fosite.AccessToken type.


arekkas:
Thank you!

Issue #207, opened 2017-08-05
Introspection of refresh token during access token validation


beornf:
According to oauth2 the refresh token is only used to generate a new access token and not read/write content on the resource server. As per the spec RFC 6749 section 1.5
> Unlike access tokens, refresh tokens are intended for use only with authorization servers and are never sent to resource servers.

I am using the introspect token method in http middleware to validate access tokens, however this line allows refresh tokens to pass as well: https://github.com/ory/fosite/blob/master/handler/oauth2/introspector.go#L29.

Is my conclusion right that it would be better to check the passed in `tokenType` explicitly?

arekkas:
Token introspection RFC allows both refresh and access token validation :) I find this confusing as well, but see for yourself: https://tools.ietf.org/html/rfc7662#section-2.1

beornf:
Right so that parameter is the token type hint. I'm still looking for a way for the introspection endpoint to only validate access tokens and ignore refresh tokens as these shouldn't be accepted by the resource server.

Would bringing the warden strategy from hydra into fosite as a custom factory be useful or is too much outside the oauth2 standard?
https://github.com/ory/hydra/blob/master/warden/warden_strategy.go

arekkas:
Hm, maybe make this configurable in the standard validator? The token_hint is not binding, only something to reduce processing time, don't rely on it :)

arekkas:
Addressed and closed in #208

PR #208, opened 2017-08-10
Configure core validator with access only option

beornf:
My attempt to make the core validator configurable refer to #207. If you agree on the direction I'll write tests for the case that `AccessOnly` is turned on.

### DIFF
--- a/compose/config.go
+++ b/compose/config.go
@@ -19,8 +19,8 @@ type Config struct {
 	// HashCost sets the cost of the password hashing cost. Defaults to 12.
 	HashCost int
 
-	// AccessOnly sets whether the introspection endpoint only validates access tokens.
-	AccessOnly bool
+	// DisableRefreshTokenValidation sets the introspection endpoint to disable refresh token validation.
+	DisableRefreshTokenValidation bool
 
 	ScopeStrategy fosite.ScopeStrategy
 }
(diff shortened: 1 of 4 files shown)


compose/config.go
```diff
@@ -61,3 +64,8 @@ func (c *Config) GetHashCost() int {
 	}
 	return c.HashCost
 }
+
+// GetAccessOnly returns whether the introspection endpoint only validates access tokens.
```
arekkas: I don't think this is required as the default value for bool is false anyways.

handler/oauth2/introspector.go
```diff
@@ -10,10 +10,19 @@ import (
 type CoreValidator struct {
 	CoreStrategy
 	CoreStorage
+	AccessOnly    bool
 	ScopeStrategy fosite.ScopeStrategy
 }
 
 func (c *CoreValidator) IntrospectToken(ctx context.Context, token string, tokenType fosite.TokenType, accessRequest fosite.AccessRequester, scopes []string) (err error) {
+	if c.AccessOnly {
+		if err = c.introspectAccessToken(ctx, token, accessRequest, scopes); err == nil {
```
arekkas: this can be simplified to one line

compose/config.go
```diff
@@ -19,6 +19,9 @@ type Config struct {
 	// HashCost sets the cost of the password hashing cost. Defaults to 12.
 	HashCost int
 
+	// AccessOnly sets whether the introspection endpoint only validates access tokens.
+	AccessOnly bool
```
arekkas: I think the name of this variable is misleading, how about `DisableRefreshTokenValidation`?

arekkas:
Perfect, thanks!

PR #210, opened 2017-08-20
Set expires at for password credentials flow

beornf:
Noticed that expires at is not set in the session by default unlike the authorize code, client credential and refresh flows.

### DIFF
--- a/handler/oauth2/flow_resource_owner.go
+++ b/handler/oauth2/flow_resource_owner.go
@@ -2,6 +2,7 @@ package oauth2
 
 import (
 	"fmt"
+	"time"
 
 	"context"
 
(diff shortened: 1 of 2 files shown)


arekkas:
Thank you! Would you mind adding a (failing) test case for this too?

beornf:
Added a failing check to the final test case.

arekkas:
perfect, thank you!

kujenga:
Hmm, I just upgraded our fosite dependency and this PR seems to have broken my app's use of fosite. The expiration time set in the session, which seems like it should be preferential, is now overridden by this value.

Previously, the behavior used the expiration times in the session set here:

https://github.com/ory/fosite/blob/dd9398ea0553b07d63022af50ee2090d1616c5a9/session.go#L46-L69

I can work around this be setting the configuration variable to the access token time that we want, but that seems to defeat the purpose of the custom expiration times in the session.

kujenga:
Unless I was perhaps misusing the `ExpiresAt` map before? I was pre-filling values and expecting those to be respected/used.

arekkas:
This might be https://github.com/ory/fosite/issues/211

beornf:
In the case of pre-filling the session expiration time it might help to wrap all session expiration time setters with a nil check:
`if request.GetSession().GetExpiresAt(fosite.AccessToken).IsZero() ...`

PR #216, opened 2017-10-05
handler/oauth2: set expiration time before the access token is generated

nikita-v:
Hi!
JWT access token has invalid value in "exp" claim if implicit flow is used. This is because the token is generated before it is assigned an expiration time.


arekkas:
Thank you!

PR #227, opened 2017-12-06
Simplifies error contexts

arekkas:
Simplifies how errors are instantiated. Errors now contain all necessary information without relying on `fosite.ErrorToRFC6749Error` any more. `fosite.ErrorToRFC6749Error` is now an internal method and was renamed to `fosite.errorToRFC6749Error`.


access_error.go
```diff
@@ -27,8 +27,8 @@ func (c *Fosite) WriteAccessError(rw http.ResponseWriter, _ AccessRequester, err
 func writeJsonError(rw http.ResponseWriter, err error) {
 	rw.Header().Set("Content-Type", "application/json;charset=UTF-8")
 
-	rfcerr := ErrorToRFC6749Error(err)
-	js, err := json.Marshal(rfcerr)
+	rfcerr := errorToRFC6749Error(err)
+	js, err := json.Marshal(err)
```
kujenga: @arekkas is marshaling the `err` here instead of the `rfcerr` intentional?

arekkas: Nice catch, it should be rfcerr, would you mind creating a PR to fix this? :)

arekkas: I quickly resolved it myself. Noticed that it's not an issue per se, but it does look weird so it's good to have it updated! Thanks for reviewing

kujenga: Got it, sure thing! To clarify, do you have the PR covered then?

Issue #213, opened 2017-09-12
Support for RFC 7636


natebrunson:
I would love to see support for [RFC 7636](https://tools.ietf.org/html/rfc7636) built into Fosite. I currently have some dependency on it for an existing client and it is something would prevent me from using Fosite (via Hydra) for that client until it could be supported. 

If this is something the core team thinks would be a good fit, I'd be willing to contribute a PR.

arekkas:
I think it's a very useful addition and it should be implemented easily. I would love to see a PR. For that, it would be perfect if we could discuss how to approach this first. Have you ever worked with fosite before?

natebrunson:
I have not worked with it before. I've browsed the source a bit, read through the CONTRIBUTING doc, and digging through the go doc now.  I'm not sure I've developed an opinion yet on how to approach it, so I'm all ears.

arekkas:
The only relevant pieces I see are point 4.3 and 4.4 in the spec. I'm quite busy at the moment so I didn't give it a very tough look, but from what I saw we only need to store the code challenge server side and then return it later on.

The [authorize request](https://github.com/ory/fosite/blob/master/authorize_request.go) embeds the [request](https://github.com/ory/fosite/blob/master/request.go) which in turn contains a [list](https://github.com/ory/fosite/blob/master/request.go#L17) of all form values (POST body and GET query params) implying that it will also carry the `code_challenge` if set. Thus, all we really need is a [token endpoint handler](https://github.com/ory/fosite/blob/786b971ca1d36a8f0bd0a5c0bfa798802d5c0c26/handler.go#L26-L34) that looks up if the `code_challenge` is set and executes the appropriate business logic (like setting the right response values) in the request.

I think that should do it!

natebrunson:
Yeah I think it's pretty trivial to implement. Basically if the code_challenge and code_challenge_method come through in the authorization request we need to store that probably alongside any state that is associated with the authorization code. Then when the code is exchanged at the token endpoint the request must also contain the code_verifier, which is used to rebuild the code_challenge using the algorithm defined in the code_challenge_method. I propose fosite only supports the 's256'  challenge method, because the 'plain' method is just a waste of time.

Please excuse the lack of markdown I'm posting late from my phone. Hopefully the underscores did enough to imply that when I'm talking about form parameters. I'll get a PR put together and send it over.

arekkas:
Sounds good, I agree on the s256 part!

PR #244, opened 2018-01-10
feat(oauth): Add RFC7636 aka PKCE support.

Zenithar:
Ref #213 

I have implemented RFC7636 aka PKCE. 
But not really happy about Session usage in order to recevive CodeChallenge on token endpoint, maybe need some refactoring.


access_request_handler.go
```diff
@@ -89,6 +91,30 @@ func (f *Fosite) NewAccessRequest(ctx context.Context, r *http.Request, session
 	}
 	accessRequest.Client = client
 
+	// PKCE Handling
```
arekkas: I really think this should be a custom handler and not hardcoded here

authorize_request.go
```diff
@@ -25,6 +25,11 @@ type AuthorizeRequest struct {
 	State                string    `json:"state" gorethink:"state"`
 	HandledResponseTypes Arguments `json:"handledResponseTypes" gorethink:"handledResponseTypes"`
 
+	// Optional code_challenge as described in rfc7636
+	CodeChallenge string `json:"code_challenge" gorethink:"code_challenge"`
```
arekkas: I really think this can be covered by `Request.Form`

authorize_request_handler.go
```diff
@@ -80,5 +79,41 @@ func (c *Fosite) NewAuthorizeRequest(ctx context.Context, r *http.Request) (Auth
 
 	// Remove empty items from arrays
 	request.SetRequestedScopes(removeEmpty(strings.Split(r.Form.Get("scope"), " ")))
+
+	// Proof Key for Code Exchange by OAuth Public Clients
+	// https://tools.ietf.org/html/rfc7636
+	//
+	// PKCE is only for code response type
+	if request.ResponseTypes.Has("code") {
```
arekkas: Same thing here, this shouldn't be a hardcoded thing but instead an add-in via handlers

authorize_request.go
Zenithar: I consider that CodeChallenge is like State that's why in coded it here.

authorize_request_handler.go
Zenithar: Yep I'm agree. Applying "Do it, do it right, to do better" philosophy ^^

access_request_handler.go
Zenithar: Most of the code are (re)generated mocks ^^

Zenithar: About PKCE settings (RequiredForPublicClient), maybe via a decorator (sort of middleware) in the authorize and token endpoints.

This could be helpful to add / process parameters before handling core OAuth operations.

authorize_request.go
Zenithar: My first problem was the Session usage, due to the fact it's an interface and shared in openid and oauth2 handler (I fix missing implementation on first commit). I had to store code_challenge to compare it on the token retrieval step, maybe Request.Form is a better place if it is shareable like session.

arekkas: So the `Request.Form` thing will contain all request parameters from the `/auth` endpoint, it will thus also contain `code_challenge` if that parameter is given to the `/auth` endpoint. You won't have to deal with any extra sessions or stuff there.

access_request_handler.go
arekkas: The idea of this framework is to have handlers [such as this one for the auth endpoint](https://github.com/ory/fosite/blob/master/handler/oauth2/flow_authorize_code_auth.go) which implement parts of the oauth2 spec. What you would do (probably) is some checks at the `/auth` endpoint and then some checks or updates to the response at the [`/token` endpoint](https://github.com/ory/fosite/blob/master/handler/oauth2/flow_authorize_code_token.go)

arekkas:
@Zenithar can I help you some way to push this forward? :)

Zenithar:
Yes no problem ^^

arekkas:
Ok, so I'm not really fluent yet on PCKE internals but as far as I understood it works like this:

1. Client generates verifier + challenge
2. Client sends challenge to /auth
3. Server responds with auth code code
4. Client sends verifier + auth code to /token
5. Server validates verifier with challenge from auth code

This is very easy to achieve with a new custom handler:

1. `PKCEHandler.HandleAuthorizeEndpointRequest` checks if the client is public and if so, checks if a challenge was set. Depending on wether or not PKCE is forced (this can be a field in `PKCEHandler`) an error is thrown or not.
2. `PKCEHandler. HandleTokenEndpointRequest` checks if the client is public, if yes, checks if a challenge was set (using `AuthorizeRequest.Form["code_challenge"]`. Depending on the wether or not PKCE is forced, an error is returned if code challenge is not set.
3. If code challenge is set, it validates the challenge using the verifier. If no verifier was given an error is returned. If the verifier is invalid, an error is returned as well.

Does this make sense?

Zenithar:
This flow is mainly used by mobile application.

arekkas:
I took a shot at it in #246

PR #245, opened 2018-01-24
Introspection: Decode of Basic Authorization username/password

dolbik:
Signed-off-by: Dmitry Dolbik <dolbik@gmail.com>

Introspection handler does not decode a valid encoded username and password on Basic Authorization process.
For example access_request_handler Basic Authorization parameters decoded properly

Basic Authorization behaviour of endpoints should be the same i think

### DIFF
--- a/introspection_request_handler.go
+++ b/introspection_request_handler.go
@@ -17,6 +17,7 @@ package fosite
 import (
 	"context"
 	"net/http"
+	"net/url"
 	"strings"
 
 	"fmt"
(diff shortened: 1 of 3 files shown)


arekkas:
Thank you for your contribution. Wouldn't it make sense to use https://github.com/ory/fosite/blob/master/access_request_handler.go#L109 and https://github.com/ory/fosite/blob/master/access_request_handler.go#L121 instead of duplicating the code here?

dolbik:
At this case authorization by POST data will be allowed. According RFC it is possible to use different methods of authorization and we need decide whether this is necessary

introspection_request_handler.go
```diff
@@ -122,16 +123,25 @@ func (f *Fosite) NewIntrospectionRequest(ctx context.Context, r *http.Request, s
...
+		if err != nil {
+			return &IntrospectionResponse{Active: false}, errors.WithStack(ErrRequestUnauthorized.WithDebug("HTTP Authorization header missing.WithDebug(malformed or credentials used are invalid"))
+		}
 
 		client, err := f.Store.GetClient(ctx, clientID)
 		if err != nil {
 			return &IntrospectionResponse{Active: false}, errors.WithStack(ErrRequestUnauthorized.WithDebug("HTTP Authorization header missing.WithDebug(malformed or credentials used are invalid"))
 		}
 
+		clientSecret, err := url.QueryUnescape(secret)
```
arekkas: This should be moved before `GetClient` as it might reduce database strain when passwords are formatted incorrectly.

arekkas:
Right, I overlooked that it's actually using post form as a fallback - nice catch!

introspection_request_handler.go
dolbik: updated

arekkas:
Perfect, thank you!

PR #257, opened 2018-04-06
Fix example in README

damz:
Fixes the following:

 * Handlers should just use `Request.Context()` nowadays
 * Missing private key creation
 * Missing trailing comma in structure literal
 * Use `.Username` instead of the invalid `.UserID` in example user-specific action
 * Broken syntax in call to `compose.ComposeAllEnabled`
 * Replace "examplary" ("serving as a desirable model") by "example" ("a thing characteristic of its kind or illustrating a general rule")


arekkas:
Thank you!

PR #262, opened 2018-04-26
Regression fix for request ID in refresh token flow

beornf:
The sanitize request method introduced a change where the generated ID for an access token and refresh token are different for the same response.

This is due to the ID being generated on the sanitized request not being passed back to the handler e.g. https://github.com/ory/fosite/blob/master/handler/oauth2/flow_refresh.go#L110.

To mitigate this the request ID is pre-generated on the original requester. There may be a better approach that would also pass the test.

### DIFF
--- a/request.go
+++ b/request.go
@@ -137,6 +137,7 @@ func (a *Request) Sanitize(allowedParameters []string) Requester {
 	}
 
 	*b = *a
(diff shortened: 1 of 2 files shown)


arekkas:
Nice catch, thanks!

PR #285, opened 2018-05-30
Treat posted data as multipart

mvrhov:
Use Multipart form parse function as some client implementations use it to send form data.
e.g All of our native clients which use CURL

### DIFF
--- a/access_request_handler.go
+++ b/access_request_handler.go
@@ -62,7 +62,7 @@ func (f *Fosite) NewAccessRequest(ctx context.Context, r *http.Request, session 
 
 	if r.Method != "POST" {
 		return accessRequest, errors.WithStack(ErrInvalidRequest.WithDebug("HTTP method is not POST"))
-	} else if err := r.ParseForm(); err != nil {
+	} else if err := r.ParseMultipartForm(1 << 20); err != nil && err != http.ErrNotMultipart {
 		return accessRequest, errors.WithStack(ErrInvalidRequest.WithDebug(err.Error()))
 	}
 
--- a/authorize_request_handler.go
+++ b/authorize_request_handler.go
@@ -39,7 +39,7 @@ func (c *Fosite) NewAuthorizeRequest(ctx context.Context, r *http.Request) (Auth
 		Request:              *NewRequest(),
 	}
 
-	if err := r.ParseForm(); err != nil {
+	if err := r.ParseMultipartForm(1 << 20); err != nil && err != http.ErrNotMultipart {
 		return request, errors.WithStack(ErrInvalidRequest.WithDebug(err.Error()))
 	}
 
--- a/integration/helper_endpoints_test.go
+++ b/integration/helper_endpoints_test.go
@@ -132,7 +132,7 @@ func authCallbackHandler(t *testing.T) func(rw http.ResponseWriter, req *http.Re
 
 func tokenEndpointHandler(t *testing.T, provider fosite.OAuth2Provider) func(rw http.ResponseWriter, req *http.Request) {
 	return func(rw http.ResponseWriter, req *http.Request) {
-		req.ParseForm()
+		req.ParseMultipartForm(1 << 20)
 		ctx := fosite.NewContext()
 
 		accessRequest, err := provider.NewAccessRequest(ctx, req, &oauth2.JWTSession{})
--- a/introspect.go
+++ b/introspect.go
@@ -44,8 +44,7 @@ func AccessTokenFromRequest(req *http.Request) string {
 	if len(split) != 2 || !strings.EqualFold(split[0], "bearer") {
 		// Nothing in Authorization header, try access_token
 		// Empty string returned if there's no such parameter
-		err := req.ParseForm()
-		if err != nil {
+		if err := req.ParseMultipartForm(1 << 20); err != nil && err != http.ErrNotMultipart {
 			return ""
 		}
 		return req.Form.Get("access_token")
--- a/introspection_request_handler.go
+++ b/introspection_request_handler.go
@@ -114,8 +114,11 @@ import (
 func (f *Fosite) NewIntrospectionRequest(ctx context.Context, r *http.Request, session Session) (IntrospectionResponder, error) {
 	if r.Method != "POST" {
 		return &IntrospectionResponse{Active: false}, errors.WithStack(ErrInvalidRequest.WithDebug("HTTP method is not POST"))
-	} else if err := r.ParseForm(); err != nil {
+	} else if err := r.ParseMultipartForm(1 << 20); err != nil && err != http.ErrNotMultipart {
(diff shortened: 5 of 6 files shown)


arekkas:
I don't think multipart is allowed per spec, or am I mistaken?

mvrhov:
The rfc6749 doesn't forbid it it just gives all examples in `application/x-www-form-urlencoded`, because it's more terse and readable.

Hm tests. Why the hell are they failing now.

arekkas:
My guess is that tests are failing because the valid `application/x-www-form-urlencoded` content type has been replaced with multipart which causes the encoding of e.g. the client-id to fail. Multipart should be a fallback, not the default (if at all, I have to check with the specs if that content type makes sense).

arekkas:
By the way, just do `go test ./...` locally to check if tests pass locally!

mvrhov:
I changed this now to ignore the specific error (NotMultiart, just like request.FormValue does) will do.

mvrhov:
All tests except one now pass. And I have no Idea why the last one reports different error

access_request_handler.go
```diff
@@ -62,7 +62,7 @@ func (f *Fosite) NewAccessRequest(ctx context.Context, r *http.Request, session
 
 	if r.Method != "POST" {
 		return accessRequest, errors.WithStack(ErrInvalidRequest.WithDebug("HTTP method is not POST"))
-	} else if err := r.ParseForm(); err != nil {
+	} else if err := r.ParseMultipartForm(2 ^ 24); err != nil && err != http.ErrNotMultipart {
```
arekkas: `ParseMultipartForm` does not parse anything except the content type `multipart`, so the proper encoding has replaced with `multipart` as `ParseForm` has been replaced and if the form is not `multipart`, then the flow fails (so every oauth2 library :D ).

Parsing multipart should be a **fallback**, not the **default**!

Also, do we really need 16MB of data here? Smells like a possible DoS vulnerability!

arekkas:
Hm, I don't think we should allow this, see:

https://tools.ietf.org/html/rfc6750#section-2.2

That explicitly enforces `application/x-www-form-urlencoded`. Will close if you agree.

mvrhov:
I don't agree as the last thing I want to do is maintain my own fork of the library. We have numerous copies of our software out and removing support for old versions is out of the question. 
The other thing is that the implementations in other languages are a lot more forgiven about this and parse such requests without any issues.

arekkas:
Sorry, I linked to a spec that doesn't concern this PR, I'll heck again to see if this may be allowed

arekkas:
I'm still looking for hints in the specs as the are very vague. However, the big ones (like Google OAuth 2.0) are not forgiving when it comes to this parameter, see [the docs](https://developers.google.com/identity/protocols/OAuth2InstalledApp#handlingtheresponse).

Usually, the issue you describe comes from using either outdated OAuth2 clients or self-written code that handles OAuth2. Both should (must) be changed in order to (a) have an OAuth2 compliant flow and (b) remove any potential security issues. Keep in mind that this is probably at the core of your authorization flow, so why rely on potentially insecure code?

Second, you won't have to maintain your own fork. It's very easy to either write a middleware that encodes from `multipart` and decodes to `application/x-www-form-urlencoded` or you can use an API gateway that does that (I think most support this type of operation). It will give you more flexibility (e.g. allow this only for certain clients, define the maximum memory size yourself, ...) while removing the need of having non-standard things in this library.

I know that we're strict and that's the point this library is making. We don't want to compromise integrity, reliability, or security for 1% use cases.

Having said that, I haven't explicitly decided if this should make it or not in the library. I'll see what other big OAuth2 Providers are doing and will try to find information on this in the specs.

access_request_handler.go
mvrhov: I've just seen this review. I've changed this to be 1^10 aka 1M. The other thing being
I don't know which version of go you are running but in 1.10 ParseMultipartForm first calls ParseForm and then it continues to do a multipart parse

arekkas: > I've just seen this review. I've changed this to be 2^20 aka 1M. 

Nice!

> I don't know which version of go you are running but in 1.10 ParseMultipartForm first calls ParseForm and then it continues to do a multipart parse

I think the tests run Go 1.8. We should definitely make this bc compatible with older versions of Go!

> Also in case you missed it. If you use the FormValue or PostFormValue of the request both of them call ParseMultipartForm with the default value of 32M!

I didn't know that. We are however only calling `PostForm` which requires a prior `ParseForm` (or it's nil). This led me to investigate though and it seems that `ParseForm` reads up to 10MB, so also quite a lot (not 32M but still). So 1MB is definitely fine here, and 16MB was probably too! Parsing explicitly only 1MB is even better though!

mvrhov: I've checked the source of request.go in [1.8](https://github.com/golang/go/blob/555eb70db2f7ff4aa22768339425430c1dfdf6ac/src/net/http/request.go#L1203-L1239) and it seems that the ParseMultipartForm code is the same and it also calls ParseForm

arekkas: Ohh, I see, I didn't know that. I was looking at another line where multipart is ignored but apparently it's handled in the outer functions of that call. Thank you for clarifying that! Learned a ton about form parsing in this PR :)

arekkas:
I checked the code and the failing test case is because `ParseForm` returns an error if no the body is empty. Apparently, that's not the case with `ParseMultiForm` which isn't bothered by this. Not sending a body should IMO return ErrInvalidRequest, not an unauthorized error (which is caused by the request lacking a bearer token).

arekkas:
So, I don't think there's actually anything speaking against adding this feature and I was not able to find a resource that would say otherwise. There's however still a failing test :|

mvrhov:
Yeah. But for that we would need to re-read the body.
Example for access_request_handler...

```go
	if body, _ := ioutil.ReadAll(r.Body); len(body) == 0 {
		return accessRequest, errors.WithStack(ErrInvalidRequest.WithDebug("missing form body"))
	}

// or
	if body, _ :=ioutil.ReadAll(io.LimitReader(r.Body, 1)); len(body) == 0 {
		return accessRequest, errors.WithStack(ErrInvalidRequest.WithDebug("missing form body"))
	}

```

arekkas:
Ugh, that sucks, especially because we actually can't re-read the body afaik, as the body is treated as a stream, unless `ParseForm` does some magic there. Another option could be to check if `PostForm` has any data? As in `if len(r.PostForm) == 0`?

mvrhov:
I've added this check only into the revoke and introspect handlers. 
If I also add it to others then a lot of tests fail. Because the ErrInvalidRequest is returned instead of for example ErrInvaildClient.

arekkas:
Perfect, I'll take another look at this today and will merge if nothing else emerges. Thank you for your persistence and work! :)

PR #293, opened 2018-07-19
Return ErrLoginRequired error when `prompt=none` and session doesn't have a subject

mvrhov:
Return ErrLoginRequired error when `prompt=none` and session doesn't have a subject

This would happen if one hits the auth endpoint and the auth endpoint has no information/session about current user. Right now the generic 400 error is returned.
By the standard:

> The Authorization Server MUST NOT display any authentication or consent user interface pages. An error is returned if an End-User is not already authenticated or the Client does not have pre-configured consent for the requested Claims or does not fulfill other conditions for processing the request. The error code will typically be login_required, interaction_required, or another code defined in Section 3.1.2.6. This can be used as a method to check for existing authentication and/or consent.

### DIFF
--- a/handler/openid/validator.go
+++ b/handler/openid/validator.go
@@ -102,6 +102,9 @@ func (v *OpenIDConnectRequestValidator) ValidatePrompt(req fosite.AuthorizeReque
 
 	claims := session.IDTokenClaims()
 	if claims.Subject == "" {
+		if stringslice.Has(prompt, "none") {
+			return errors.WithStack(fosite.ErrLoginRequired.WithDebug("Cannot authenticate user silently because user's session has expired or it has never been authenticated."))
+		}
 		return errors.WithStack(fosite.ErrServerError.WithDebug("Failed to validate OpenID Connect request because session subject is empty."))
 	}
 


handler/openid/validator.go
```diff
@@ -102,6 +102,9 @@ func (v *OpenIDConnectRequestValidator) ValidatePrompt(req fosite.AuthorizeReque
 
 	claims := session.IDTokenClaims()
 	if claims.Subject == "" {
+		if stringslice.Has(prompt, "none") {
+			return errors.WithStack(fosite.ErrLoginRequired.WithDebug("Cannot authenticate user silently because user's session has expired or it has never been authenticated."))
+		}
 		return errors.WithStack(fosite.ErrServerError.WithDebug("Failed to validate OpenID Connect request because session subject is empty."))
```
arekkas: Thank you! This is actually a bit different here. If `claims.Subject` is empty it implies that the fosite framework is used incorrectly. `claims.Subject` has to be hydrated because otherwise the ID token is invalid. That is the point of this check.

I think there is an error `fosite.ErrMisconfiguration` which would make this more obvious than `fosite.ErrServerError`, so from my view it would be ok to change the error.

mvrhov:
So if this is not proper solution for that, then I don't know how I can relay this to the client. Without parsing everything myself in auth handler

arekkas:
I’m not sure why this happens in your case (maybe you could show the auth handler?). If this happens then it’s a bug/misuse of the library in your auth handler - it does not have anything to do with the prompt parameter. That’s why I said that the error message can be improved.

arekkas:
Oh wait, I think I get it now. Please show the auth handler regardless!

mvrhov:
I've thrown out all the unnecessary code

```go
ctx := fosite.NewContext()

ar, err := server.NewAuthorizeRequest(ctx, r)

if err != nil {
    log.Printf("Error occurred in NewAuthorizeRequest: %s\nStack: \n%s", err, err.(stackTracer).StackTrace())
    server.WriteAuthorizeError(w, ar, err)
    return
}

session := newSession(nil, nil, r.Host)
account, ok := r.Context().Value("user").(*account.Account)
if !ok {
    if "none" != ar.GetRequestForm().Get("prompt") {
        //no account and not promted for no UI, redirect to login
        session := r.Context().Value("session").(*scs.Session)
        session.PutString(w, "redirectURI", r.URL.RequestURI())
        http.Redirect(w, r, "/login", http.StatusFound)
        return
    }
} else {
    session = newSession(account, nil, r.Host)
}

response, err := server.NewAuthorizeResponse(ctx, ar, session)

if err != nil {
    log.Printf("Error occurred in NewAuthorizeResponse: %s\nStack: \n%s", err, err.(stackTracer).StackTrace())
    server.WriteAuthorizeError(w, ar, err)
    return
}

// Last but not least, send the response!
server.WriteAuthorizeResponse(w, ar, response)
```

arekkas:
Sorry, completely overlooked the notification. I'll check this today.

arekkas:
Ok so if I understand your code correctly, `account, ok := r.Context().Value("user").(*account.Account)` checks if the user is logged in - right? So what would be the case for ` server.NewAuthorizeResponse(ctx, ar, session)` to not have the session hydrated with an account?

I think the current use of `OpenIDConnectRequestValidator` is in fosite terrible. Reason being that we only check on `WriteAuthorizeResponse`. Instead, this should be executed somewhere else (notably before you decide to log the user in). In ORY Hydra we call [`ValidatePrompt` directly](https://github.com/ory/hydra/blob/master/consent/strategy_default.go#L292-L325) before deciding if the user request is valid or not. While implementing that I did not think too much about usage in Fosite (which causes this mess).

So my point stands, adding this check here does not fix the issue at hand. The problem that needs to be solved here is that fosite should do some type of "pre-check" with the existing user session and throw errors accordingly. I'm currently not sure how to solve this properly but open to ideas.

arekkas:
Oh, totally my bad, I overlooked your check with:

```
    if "none" != ar.GetRequestForm().Get("prompt") {
```

in this case - your code makes total sense. Assuming that `prompt=none` you would expect `NewAuthorizeResponse` to handle the error correctly. That makes total sense.

I think it should work a bit differently, because IMO `"none" != ar.GetRequestForm().Get("prompt")` should not be of the developers concern.

Maybe we could introduce a new method `ValidateAuthorizeRequest`? For example:

```
ctx := fosite.NewContext()
ar, err := server.NewAuthorizeRequest(ctx, r)

// If the session is not set (nil) we assume the user is signed out
session, ok := r.Context().Value("user").(*Session)

if silent, err := server.ValidateAuthorizeRequest(ctx, ar, session); err != nil {
    // An error would be, for example, `ErrLoginRequired` or maybe even `ErrConsentRequired`?
    server.WriteAuthorizeError(w, ar, err)
    return
} else if silent {
 // We can authenticate and authorize the user silently. This would assume that we have access to all of the previous consent decisions made by the user as well.

  // .. we do not show the ui (e.g. because prompt=none)
}

// show login ui + consent ui

// response, err := server.NewAuthorizeResponse(ctx, ar, session)
// ...
```

mvrhov:
That would also work, but I don't know other `prompt` usecases well enough to know what else could be a valid return value from `ValidateAuthorizeRequest` e.g silent is one of them but each of . 
Introducing new function would probably also mean to move a lot of validator.go code to a different place? I'm not sure I'm ready to take such a big responsibility right now as I still don't know the internals and the oidc well enough.

Can we merge this as is. And can leave the refactoring for some other time?

arekkas:
> That would also work, but I don't know other prompt usecases well enough to know what else could be a valid return value from ValidateAuthorizeRequest e.g silent is one of them but each of .

That's exactly the problem, another prompt is `consent` and `select_account`. Those should work as well, specifically the `consent` part.

> Can we merge this as is. And can leave the refactoring for some other time?

Sorry but no. This has to be resolved properly.

mitar:
I stumbled upon this as well, now that I wanted to implement `prompt=none`. It seems fosite does not really support well `prompt` parameter?

mitar:
I think I managed to implement this without a need for `ValidateAuthorizeRequest`. The high-level flow for me is:

```
authorizeRequest := NewAuthorizeRequest()
session := getCurrentSession()
if !isPromptNone(authorizeRequest)) && session == nil {
    showLogin()
}
if !isPromptNone(authorizeRequest)) && session != nil && !consentShown() {
    showGrantConsent()
}
if session != nil && consentShown() {
    authorizeResponse := NewAuthorizeResponse(authorizeRequest, session)
    write(authorizeResponse)
} else if isPromptNone(authorizeRequest) && session == nil {
    returnLoginRequiredError()
} else if isPromptNone(authorizeRequest) && session != nil {
    returnInteractionRequiredError()
}
```

Am I missing something?

aeneasr:
Yeah `prompt=none` isn't really that well implemented here - actually all of `prompt=consent`, `prompt=login`, ...

mitar:
Sure, but I do not see what is missing. Not sure why would one have to validate existing session before showing login if session is not available? To me it is or a) session is present and then you validate things inside `NewAuthorizeResponse`, or b) session is not present and then you return an error if prompt=none, or prompt for login otherwise. Am I missing something? Why would it matter to validate anything if there is already an error because of prompt=none?

aeneasr:
So I checked a bit into the code sample, here's my take:

```
authorizeRequest := NewAuthorizeRequest()
if !hasSession() || isPromptLogin() || isPromptChooseAccount() {
  if isPromptNone() {
     return InteractionRequiredError()
  }

  showLogin()
}

if !hasConsent(authorizeRequest) || isPromptConsent() {
  if isPromptNone() {
     return InteractionRequiredError()
  }

  showConsent()
}

completeRequest()
```

mitar:
I would return "login required" for the first error, but yes, this looks similar to mine. So I do not think any change is necessary to fosite and this PR can be closed without merging?

aeneasr:
I mean what we could do would be a small helper function that returns an ENUM explaining what you need to do?

aeneasr:
https://github.com/ory/fosite/issues/515

PR #300, opened 2018-09-13
WIP - tracing - create spans around hash compare

aaslamin:
Fosite does some non-negligible computation such as bcrypt (_for credential verification_) and hmac (_for token verification_)

This PR adds the opentracing library to fosite and creates a span around the call to bcrypt. 

This is particularly useful for anyone that has used fosite to build an authz service (e.g. Hydra) with an [opentracing compatible tracer](http://opentracing.io/documentation/pages/supported-tracers).

_Review: @aeneasr_

### DIFF
--- a/client_authentication.go
+++ b/client_authentication.go
@@ -28,6 +28,9 @@ import (
 	"net/http"
 	"net/url"
 
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/ext"
+
 	"github.com/dgrijalva/jwt-go"
 	"github.com/pkg/errors"
 	"gopkg.in/square/go-jose.v2"
@@ -200,7 +203,10 @@ func (f *Fosite) AuthenticateClient(ctx context.Context, r *http.Request, form u
 	}
 
 	// Enforce client authentication
+	span, _ := opentracing.StartSpanFromContext(ctx, "hash_compare")
+	defer span.Finish()
 	if err := f.Hasher.Compare(client.GetHashedSecret(), []byte(clientSecret)); err != nil {
+		ext.Error.Set(span, true)
 		return nil, errors.WithStack(ErrInvalidClient.WithDebug(err.Error()))
 	}
 
--- a/client_authentication_test.go
+++ b/client_authentication_test.go
@@ -40,6 +40,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/square/go-jose.v2"
+	"context"
 )
 
 func mustGenerateAssertion(t *testing.T, claims jwt.MapClaims, key *rsa.PrivateKey, kid string) string {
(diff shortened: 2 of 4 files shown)


client_authentication.go
```diff
@@ -200,7 +203,10 @@ func (f *Fosite) AuthenticateClient(ctx context.Context, r *http.Request, form u
 	}
 
 	// Enforce client authentication
+	span, _ := opentracing.StartSpanFromContext(ctx, "hash_compare")
+	defer span.Finish()
 	if err := f.Hasher.Compare(client.GetHashedSecret(), []byte(clientSecret)); err != nil {
```
aaslamin: This should all probably be moved into `Compare` itself so we can add additional useful tags - What do you think @aeneasr ? In order to do so, I will need to change the signature of the functions in [Hasher](https://github.com/ory/fosite/blob/master/hash.go#L28) and pass down context. For example:

```
type Hasher interface {
	// Compare compares data with a hash and returns an error
	// if the two do not match.
	Compare(ctx context.Context, hash, data []byte) error

	// Hash creates a hash from data or returns an error.
	Hash(ctx context.Context, data []byte) ([]byte, error)
}
```

This would give us the opportunity to add tags depending on the hasher used. For example, for bcrypt we would be able to add a tag for the [work factor](https://github.com/ory/fosite/blob/master/hash_bcrypt.go#L31) and the hashing algorithm that is being applied. This would benefit someone that is profiling their oauth svc.

client_authentication.go
```diff
@@ -200,7 +203,10 @@ func (f *Fosite) AuthenticateClient(ctx context.Context, r *http.Request, form u
 	}
 
 	// Enforce client authentication
+	span, _ := opentracing.StartSpanFromContext(ctx, "hash_compare")
```
aaslamin: ℹ️ btw, the second returned value here [is not an error](https://github.com/opentracing/opentracing-go/blob/master/gocontext.go#L43), it is the context object that was built around the created span - which you will use to propagate further if you need.

I am being responsible! 😅

client_authentication_test.go
```diff
@@ -364,7 +365,7 @@ func TestAuthenticateClient(t *testing.T) {
 			store.Clients[tc.client.ID] = tc.client
 			f.Store = store
 
-			c, err := f.AuthenticateClient(nil, tc.r, tc.form)
+			c, err := f.AuthenticateClient(context.TODO(), tc.r, tc.form)
```
aaslamin: https://golang.org/pkg/context/#TODO

> TODO returns a non-nil, empty Context. Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter). TODO is recognized by static analysis tools that determine whether Contexts are propagated correctly in a program.

aaslamin:
TODO: 

Pending your feedback I can:
- Change the interface for Hasher to accept a context
- Generate new mocks for Hasher
- Add test coverage to ensure a span is created with the appropriate tags

aeneasr:
Hhhmm, interesting. I think this is definitely an improvement. On the same hand, it adds some hard dependency in the library which can not be replaced easily. Maybe there is a way to have instrumentation as an abstraction? This would allow other to use different instrumentation tooling.

I think the hasher comment you left where adding `ctx` to hasher is a good start (because we can re-implement hasher easily if we need different instrumentation). We can also provide a default hasher that does not have any instrumentation and then implement a custom one in, for example, hydra.

aaslamin:
@aeneasr you are right. It doesn't make sense for fosite to worry about a concern like this. Having the ability to implement your own tracer is more flexible.

I will raise a PR to update the `Hasher` interface such that its methods take in the go context. Consumers of fosite will then have the ability to provide their own implementations to do any custom work they need (e.g. tracing).

PR #302, opened 2018-09-18
jwt: update JWTStrategy to take in context

aaslamin:
#### What is this:

As per discussed in discussed in https://github.com/ory/hydra/pull/1030#issuecomment-422205587, this PR modifies the `JWTStrategy` interface to take in `context`.

#### Notes:
- The formatting on the diff is the result of running `goimports -w` on all of the modified files.

#### Coming  🔜 : 

- If this PR is accepted, a PR will be followed in Hydra to bump it's fosite version along with its implementation of `JWTStrategy` to adhere to the new interface.

---
_Review: @aeneasr_

### DIFF
--- a/token/jwt/jwt.go
+++ b/token/jwt/jwt.go
@@ -30,17 +30,19 @@ import (
 	"fmt"
 	"strings"
 
+	"context"
+
 	"github.com/dgrijalva/jwt-go"
 	"github.com/ory/fosite"
 	"github.com/pkg/errors"
 )
 
 type JWTStrategy interface {
-	Generate(claims jwt.Claims, header Mapper) (string, string, error)
-	Validate(token string) (string, error)
+	Generate(ctx context.Context, claims jwt.Claims, header Mapper) (string, string, error)
+	Validate(ctx context.Context, token string) (string, error)
+	Hash(ctx context.Context, in []byte) ([]byte, error)
+	Decode(ctx context.Context, token string) (*jwt.Token, error)
 	GetSignature(token string) (string, error)
-	Hash(in []byte) ([]byte, error)
-	Decode(token string) (*jwt.Token, error)
 	GetSigningMethodLength() int
 }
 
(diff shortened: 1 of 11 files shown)


token/jwt/jwt.go
```diff
@@ -30,17 +30,19 @@ import (
...
 )
 
 type JWTStrategy interface {
-	Generate(claims jwt.Claims, header Mapper) (string, string, error)
-	Validate(token string) (string, error)
+	Generate(ctx context.Context, claims jwt.Claims, header Mapper) (string, string, error)
+	Validate(ctx context.Context, token string) (string, error)
+	Hash(ctx context.Context, in []byte) ([]byte, error)
+	Decode(ctx context.Context, token string) (*jwt.Token, error)
 	GetSignature(token string) (string, error)
```
aaslamin: Should this take in `context`?

token/jwt/jwt.go
```diff
@@ -30,17 +30,19 @@ import (
...
-	Hash(in []byte) ([]byte, error)
-	Decode(token string) (*jwt.Token, error)
 	GetSigningMethodLength() int
```
aaslamin: Should this take in context?

aaslamin:
@aeneasr not sure if we also want to do what you suggested in https://github.com/ory/fosite/pull/300#issuecomment-421258645 in this PR as well?

Hopefully I haven't missed anything 🙏

token/jwt/jwt.go
aaslamin: Could you elaborate a bit more on the possible signature computation by a strategy? From my understanding the implementation of this method is just parsing a string and should never result in a db lookup or expensive computation.

aaslamin:
> Thanks, this looks good! We track breaking changes in [HISTORY.md](https://github.com/ory/fosite/blob/master/HISTORY.md). Can you add instructions on what changed with this PR in there? The next version will be `0.22.0`

Sure thing!

HISTORY.md
```diff
@@ -5,8 +5,13 @@ bumps (`0.1.0` -> `0.2.0`).
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
 
+- [0.22.0](#0220)
```
aaslamin: The menu was autogenerated using doctoc via:

```
doctoc --github HISTORY.md
```

aaslamin:
@aeneasr ready for final review!

PR #303, opened 2018-09-22
Modifies Hasher to take in go context

aaslamin:
#### ⚠️ Breaking change

As per discussed in https://github.com/ory/fosite/pull/300, this PR modifies the `Hasher` interface to take in a go context.

All consumers/mocks have been updated to adhere to the new interface.

_Review: @aeneasr_


aaslamin:
**Future thinking for Hydra** 💡:

• Need to bump up Hydra's fosite version
• Need to create an `Hasher` implementation in Hydra that is instrumented with tracing
• I think we will be able to plumb in our instrumented `Hasher` here: https://github.com/ory/hydra/blob/master/cmd/server/handler_oauth2_factory.go#L124

aeneasr:
nice, thanks!

PR #308, opened 2018-09-26
eliminate hanging context.TODO() to properly propagate context [closes #307]

### LINKED ISSUE #307: [Bug] Context not propagating to token generator
I'm testing the newly added context propagation feature of both Hydra + Fosite and noticed my traces are missing chunks where I generate my JWTs:

image

If I'm following this right (and believe me with all the interfaces and such I have no idea if I am), I think the context isn't propagated throughout the code correctly. Specifically, the [jwt strategy handler](https://github.com/ory/fosite/blob/master/handler/oauth2/strategy_jwt.go#L56) doesn't seem to pass the context [down to](https://github.com/ory/fosite/blob/master/handler/oauth2/strategy_jwt.go#L157) the [jwt strategy generating](https://github.com/ory/fosite/blob/master/handler/oauth2/strategy_jwt.go#L176) the token.

This seems like a trivial fix but I'm not sure why it wasn't done beforehand. @aaslamin - was this just an oversight or is this more complicated than it looks? I'd be happy to put in a PR to fix it or leave it to you!

someone1:
Signed-off-by: Prateek Malhotra <someone1@gmail.com>

### DIFF
--- a/handler/oauth2/strategy.go
+++ b/handler/oauth2/strategy.go
@@ -34,7 +34,7 @@ type CoreStrategy interface {
 }
 
 type JWTStrategy interface {
-	ValidateJWT(tokenType fosite.TokenType, token string) (requester fosite.Requester, err error)
+	ValidateJWT(ctx context.Context, tokenType fosite.TokenType, token string) (requester fosite.Requester, err error)
 }
 
 type AccessTokenStrategy interface {
--- a/handler/openid/validator.go
+++ b/handler/openid/validator.go
@@ -51,7 +51,7 @@ func NewOpenIDConnectRequestValidator(prompt []string, strategy jwt.JWTStrategy)
 	}
 }
 
-func (v *OpenIDConnectRequestValidator) ValidatePrompt(req fosite.AuthorizeRequester) error {
+func (v *OpenIDConnectRequestValidator) ValidatePrompt(ctx context.Context, req fosite.AuthorizeRequester) error {
 	// prompt is case sensitive!
 	prompt := stringsx.Splitx(req.GetRequestForm().Get("prompt"), " ")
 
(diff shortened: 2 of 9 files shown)


someone1:
I noticed a few locations where the context wasn't being passed around properly when I looked into #307 - I can bump the version number here as there are breaking changes but since I didn't discuss this with anyone yet, I wanted to know what version to bump to and if this is something you'll accept into fosite. 

I have a PR ready for hydra to match theses changes.

@aaslamin - FYI

aaslamin:
This LGTM Prateek! 🙌 

Are there any other areas you noticed that would make sense for context to be passed thru?

handler/oauth2/strategy.go
```diff
@@ -34,7 +34,7 @@ type CoreStrategy interface {
 }
 
 type JWTStrategy interface {
-	ValidateJWT(tokenType fosite.TokenType, token string) (requester fosite.Requester, err error)
+	ValidateJWT(ctx context.Context, tokenType fosite.TokenType, token string) (requester fosite.Requester, err error)
```
someone1: first for two breaking changes

handler/openid/validator.go
```diff
@@ -51,7 +51,7 @@ func NewOpenIDConnectRequestValidator(prompt []string, strategy jwt.JWTStrategy)
 	}
 }
 
-func (v *OpenIDConnectRequestValidator) ValidatePrompt(req fosite.AuthorizeRequester) error {
+func (v *OpenIDConnectRequestValidator) ValidatePrompt(ctx context.Context, req fosite.AuthorizeRequester) error {
```
someone1: second breaking change

aeneasr:
Please document breaking changes [here](https://github.com/ory/fosite/blob/master/HISTORY.md) under a new version (0.24.0). I use [`doctoc`](https://github.com/thlorenz/doctoc) for TOC generation.

someone1:
@aaslamin - Thanks! I didn't really comb through the code to find other opportunities, these changes were just most apparent to me - though anywhere you started adding contexts as first params to interface signature definitions (e.g. `OAuth2Provider`) I would argue you should have updated all of them, not only the ones currently with underlying interactions requiring contexts.

@aeneasr - Done

aeneasr:
Thanks!

PR #321, opened 2018-10-25
raise bcrypt cost factor lower bound to x/crypto/bcrypt default

adamdecaf:
## Proposed changes

Users of this library could easily create the following:

```Go
hasher := fosite.BCrypt{}
hasher.Hash(..)
```

This is a problem because `WorkFactor` will default to 0 and `x/crypto/bcrypt` will default that to 4 (See https://godoc.org/golang.org/x/crypto/bcrypt). Instead this should be some higher cost factor.

I didn't use `config.GetHashCost()` because that's in another package. 

Callers who need a lower WorkFactor can still lower the cost, if needed.  

## Checklist

- [x] I have read the [contributing guidelines](CONTRIBUTING.md)
vulnerability, I confirm that I got green light (please contact [hi@ory.sh](mailto:hi@ory.sh)) from the maintainers to push the changes.
- [x] I signed the [Developer's Certificate of Origin](https://github.com/ory/keto/blob/master/CONTRIBUTING.md#developers-certificate-of-origin)
by signing my commit(s). You can amend your signature to the most recent commit by using `git commit --amend -s`. If you
amend the commit, you might need to force push using `git push --force HEAD:<branch>`. Please be very careful when using
force push.

### DIFF
--- a/hash_bcrypt.go
+++ b/hash_bcrypt.go
@@ -28,12 +28,17 @@ import (
 	"golang.org/x/crypto/bcrypt"
 )
 
+const DefaultBCryptWorkFactor = 12
+
 // BCrypt implements the Hasher interface by using BCrypt.
 type BCrypt struct {
 	WorkFactor int
 }
 
 func (b *BCrypt) Hash(ctx context.Context, data []byte) ([]byte, error) {
+	if b.WorkFactor == 0 {
+		b.WorkFactor = DefaultBCryptWorkFactor
+	}
 	s, err := bcrypt.GenerateFromPassword(data, b.WorkFactor)
 	if err != nil {
 		return nil, errors.WithStack(err)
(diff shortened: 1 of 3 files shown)


adamdecaf:
FYI I already emailed the security contact about this.

hash_bcrypt.go
```diff
@@ -34,6 +34,9 @@ type BCrypt struct {
 }
 
 func (b *BCrypt) Hash(ctx context.Context, data []byte) ([]byte, error) {
+	if b.WorkFactor == 0 {
+		b.WorkFactor = bcrypt.DefaultCost
```
aeneasr: Since we're defaulting to [`12`](https://github.com/ory/fosite/blob/master/compose/config.go#L111) I think this should also be 12. Potentially as a shared `const`

adamdecaf: I was thinking of that too. Where should that live?

aeneasr:
Thanks!

Issue #309, opened 2018-10-04
auth code and refresh token flows should be treated as a transaction


aaslamin:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

https://github.com/ory/fosite/blob/master/handler/oauth2/flow_refresh.go#L107-L119
https://github.com/ory/fosite/blob/master/handler/oauth2/flow_authorize_code_token.go#L144-L152

Notice how it is possible for any of those db calls to fail which will leave the authz service and client in a bad state. For example, a client's access/refresh tokens could be revoked and the call to create a new token pair could fail leaving the client very confused.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

This is a clear by looking at the code. But to simulate this, you can modify one of the methods in your storage implementation with a long sleep (or return an error intentionally) to inject a fault, re-build your service and go through one of the flows.

For example, Hydra's implementation of the `TokenRevocationStorage` interface is implemented by `FositeSQLStore` - you can modify https://github.com/ory/hydra/blob/master/oauth2/fosite_store_sql.go#L398 to return an error to simulate failure.

**What is the expected behavior?**

These flows should *be all or nothing*. The client should never have to worry about the authorization server messing up.

**Which version of the software is affected?**

All

**Proposal**

Still TBD, but potentially a new method in both https://github.com/ory/fosite/blob/master/handler/oauth2/revocation_storage.go#L30 & https://github.com/ory/fosite/blob/master/handler/oauth2/storage.go#L30 to deal with their respective flows as a TX.

_Cc: @aeneasr @fredbi_

aaslamin:
Sorry haven't had a chance to look into this yet, will pick it up once tracing work is completed.

aeneasr:
No worries, just let me know when you took a glance!

aaslamin:
If I am following correctly, this implies that a storage implementation with transaction support always needs to check first if a transaction object is present (some attribute in the struct) before attempting to fetch a new connection prior to any database operation. Because you won't know if you are dealing with a transaction that may be in progress.

```
type SomeStorageWithTXN interface {
  Begin() (SomeStorage, error)
  Commit() (error)
  Rollback() (error)

  MethodA (Result, error)
}
```


```
// Suppose this struct implements SomeStorageWithTXN
type AmirStore struct {
	DB                  *sqlx.DB
	L                   logrus.FieldLogger
[...]

PR #343, opened 2018-12-11
Addresses #309 - add support for delegating transactions to storage providers

aaslamin:
## Issue: https://github.com/ory/fosite/issues/309

This PR has:

• Added transactional support in the refresh token flow for storage provider implementations that can support transactions
• Added transactional support in the auth code token flow for storage provider implementations that can support transactions
• Added test coverage for new behaviour


handler/oauth2/flow_refresh.go
```diff
@@ -25,6 +25,8 @@ import (
 	"context"
 	"time"
 
+	"github.com/ory/fosite/util"
```
aeneasr: `util` is a go package antipattern :)

> Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains. This makes it harder for clients to use the package and makes it harder for maintainers to keep the package focused. Over time, they accumulate dependencies that can make compilation significantly and unnecessarily slower, especially in large programs. And since such package names are generic, they are more likely to collide with other packages imported by client code, forcing clients to invent names to distinguish them.  - *[Source](https://blog.golang.org/package-names)*

One possible place could be `internal` for this (as it will be hidden from the outside).

aaslamin: Ah cool, TIL 🙌 

Although if I place it in `internal`, the `Transactional` interface won't be visible by a consumer, e.g. `Hydra`. I need that to be exportable.

aeneasr: I think that would be ok, or is there a reason why it should be visible?

aaslamin: The only reason I can think of is for unit testing 🤔 

Mmm...maybe it is not necessary?

aeneasr: I think that should be explicit in unit tests - whereas here we have implicit checking (because both non-tx and tx would work). So I think it's ok to not expose it!

aaslamin:
_Update: Moved `Transactional` interface to a more appropriate package_

aaslamin:
_Update: added mock for `storage.Transactional` (other mocks were touched as a result of executing `./generate-mocks.sh` - I did touch them_

**Coming next:** unit tests to cover the new transactional behaviour of `func (c *RefreshTokenGrantHandler) PopulateTokenEndpointResponse(...)`

aaslamin:
@aeneasr yes, we identified the authorize code flow as well https://github.com/ory/fosite/blob/master/handler/oauth2/flow_authorize_code_token.go#L157-L165

I can do an audit of the other flows as well to see if we can apply this pattern elsewhere.

I am still in the progress of adding unit tests. It's taking a while to get the mocking right, but I am making good progress 😉

aeneasr:
Awesome! :)

aaslamin:
✅ _Update: added test coverage for the new transactional behaviour in the refresh token flow._

aeneasr:
Awesome! Let me know when this is ready for a final 👀

aaslamin:
Ready! 🙏

storage/transactional.go
```diff
@@ -0,0 +1,49 @@
...
+// Implementations for `Commit` & `Rollback` should look for the transaction object inside the supplied context using the same
+// key used by `BeginTX`. If these methods have been called, it is expected that a txn object should be available in the provided
+// context.
+type Transactional interface {
+	BeginTX(ctx context.Context) (context.Context, error)
+	Commit(ctx context.Context) error
+	Rollback(ctx context.Context) error
+}
+
+func MaybeBeginTx(ctx context.Context, storage interface{}) (context.Context, error) {
```
aeneasr: Can you document the exported functions please? :)

aaslamin: Done 👍

aaslamin:
✅_Update: added comments for exported helpers_

aeneasr:
👏

PR #363, opened 2019-05-18
Access token expiry client based

ayushbpl10:
Needed to add client based token expiry for an access token.

For making changes in hydra needed these changes in fosite


aeneasr:
Thank you for your work and changes, but unfortunately these won't be merged for a couple of reasons:

1. Alternating expiry times per client is generally a bad idea. It indicates that you're using Access Tokens as long living credentials for use cases they were not intended for (like API Keys). You can easily refresh Access Tokens which eliminates the need for long-living access tokens.
2. **While it is bad practice**, you can theoretically change the access token expiry on a [per-session basis](https://github.com/ory/fosite/blob/master/session.go#L41) eliminating the need for your changes here.
3. This is a breaking change.

In general, please read the contribution guidelines and don't delete the issue and pull-request templates. It is always a good idea to talk about the intended changes first before going ahead and making changes, as that might save you a lot of time developing something that won't be merged. Especially when it comes to security features or modifications (expiry time is such a case) it is very important to discuss the intended changes first!

Nevertheless, thank you for the contribution and your hard work on this PR!

PR #364, opened 2019-05-20
Add refresh_token_expires_in into the extras when the expiry time is set

mvrhov:
## Checklist

- [x] I have read the [contributing guidelines](CONTRIBUTING.md)
vulnerability, I confirm that I got green light (please contact [hi@ory.sh](mailto:hi@ory.sh)) from the maintainers to push the changes.

### DIFF
--- a/handler/oauth2/flow_authorize_code_token.go
+++ b/handler/oauth2/flow_authorize_code_token.go
@@ -186,6 +186,9 @@ func (c *AuthorizeExplicitGrantHandler) PopulateTokenEndpointResponse(ctx contex
 	responder.SetScopes(requester.GetGrantedScopes())
 	if refresh != "" {
 		responder.SetExtra("refresh_token", refresh)
+		if c.RefreshTokenLifespan > 0 {
+			responder.SetExtra("refresh_token_expires_in", int64(getExpiresIn(requester, fosite.RefreshToken, c.RefreshTokenLifespan, time.Now().UTC())/time.Second))
+		}
 	}
 
 	if err := storage.MaybeCommitTx(ctx, c.CoreStorage); err != nil {
--- a/handler/oauth2/flow_refresh.go
+++ b/handler/oauth2/flow_refresh.go
@@ -169,6 +169,9 @@ func (c *RefreshTokenGrantHandler) PopulateTokenEndpointResponse(ctx context.Con
 	responder.SetExpiresIn(getExpiresIn(requester, fosite.AccessToken, c.AccessTokenLifespan, time.Now().UTC()))
 	responder.SetScopes(requester.GetGrantedScopes())
 	responder.SetExtra("refresh_token", refreshToken)
+	if c.RefreshTokenLifespan > 0 {
+		responder.SetExtra("refresh_token_expires_in", int64(getExpiresIn(requester, fosite.RefreshToken, c.RefreshTokenLifespan, time.Now().UTC())/time.Second))
+	}
 
 	if err := storage.MaybeCommitTx(ctx, c.TokenRevocationStorage); err != nil {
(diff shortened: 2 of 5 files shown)


aeneasr:
This does not appear to be a standardized field?

mvrhov:
No it's not. A few "oauth" providers use it. One of them is linkedin. 
This is a prelude to a next PR.. We need per refresh_token ttl per oauth client. Well At least per internal/external API consumers. And different endpoints won't cut it as people are very resourceful when it comes to getting URLs 👍

aeneasr:
I see, could you lay out the idea / motivation behind that change? We're obviously trying to avoid bespoke things/definitions as much as possible.

aeneasr:
I think for [LinkedIn OAuth2](https://docs.microsoft.com/en-us/linkedin/shared/authentication/programmatic-refresh-tokens?context=linkedin/context) the use case makes sense - Refresh Tokens are valid **in total** for a set amount of time (e.g. one year). Therefore, knowing when that will be kinda makes sense. For Fosite's Refresh Tokens, the TTL is per token. So when a RT is refreshed, the TTL is reset (that's at least how it currently works). The concepts are dissimilar:

- LinkedIn wants their users to conciously re-authorize applications that have had access for longer than X days.
- OAuth2 Core Spec sets a TTL on tokens to reduce the risk of old and unused tokens being "leaked" somewhere but still valid.

mvrhov:
The problem for example is that ATM we have the refresh token TTL of 14 days. And some of our apps don't connect to the API for a longer period of time e.g a couple of days, so they have to refresh the refresh token just before it expires. So now if we change the TTL of refresh token we have to inform all of them that we changed that, so they can shorten/extend the time between refresh requests. Whereas if the the expires is provided in the header we an just say to them, we tell you when it will expire you just have to look at this data and extend.. so the procedure is the same as with access token they just have to look at the different header and everything works automatically.

mvrhov:
What's the verdict on this. IMO we need some sort of notification of refresh token ttl in library.

handler/oauth2/flow_authorize_code_token.go
```diff
@@ -186,6 +186,9 @@ func (c *AuthorizeExplicitGrantHandler) PopulateTokenEndpointResponse(ctx contex
 	responder.SetScopes(requester.GetGrantedScopes())
 	if refresh != "" {
 		responder.SetExtra("refresh_token", refresh)
+		if c.RefreshTokenLifespan > 0 {
+			responder.SetExtra("refresh_token_expires_in", int64(getExpiresIn(requester, fosite.RefreshToken, c.RefreshTokenLifespan, time.Now().UTC())/time.Second))
```
aeneasr: This should just be `SetExpiresIn(getExpiresIn(...))`

handler/oauth2/flow_refresh.go
```diff
@@ -169,6 +169,9 @@ func (c *RefreshTokenGrantHandler) PopulateTokenEndpointResponse(ctx context.Con
 	responder.SetExpiresIn(getExpiresIn(requester, fosite.AccessToken, c.AccessTokenLifespan, time.Now().UTC()))
 	responder.SetScopes(requester.GetGrantedScopes())
 	responder.SetExtra("refresh_token", refreshToken)
+	if c.RefreshTokenLifespan > 0 {
+		responder.SetExtra("refresh_token_expires_in", int64(getExpiresIn(requester, fosite.RefreshToken, c.RefreshTokenLifespan, time.Now().UTC())/time.Second))
```
aeneasr: Same here

aeneasr:
Sorry for the late reply, this kind of slipped my radar.

aeneasr:
Hey there - I still think this is useful - would you be up for amending the requested changes? :)

PR #372, opened 2019-07-02
get rid of gorilla deps

butonic:
hey, I wanted to add a fosit based oidc provider to [our project](https://github.com/cs3org/reva/pull/130) and wondered why gorilla/mux was being pulled in. This PR gets rid of mux as a dependency and uses a simple `ShiftPath` to do the routing as described in https://blog.merovius.de/2017/06/18/how-not-to-use-an-http-router.html. This slims down the dependencies a bit.

You are only using the router for testing, so I think getting rid of `gorilla/mux` is a net win.


aeneasr:
Thank you for your contribution. gorilla/mux is only used for internal testing and should not be pulled when consuming fosite as a library. Using a proper router increases readability in the tests, so I don't think it makes sense to merge this.

tboerger:
But using go modules always add gorilla mux as an additional indirect dependency because of that :(

aeneasr:
Go Modules is still in beta. Not pulling test dependencies is one of the open issues and ongoing discussions. I think that has been solved with Go Module Proxies already (iirc). To conclude: This will be fixed automatically as the toolchain improves.

aeneasr:
Here's the comment explaining how go modules deals with test dependencies: https://github.com/golang/go/issues/26913#issuecomment-411976222

tboerger:
I have already seen this comment, and it sounds like an expected behavior which won't change. So you really want to keep mux just for the tests because it is a little bit more readable compared to the ShiftPath function? :(

aeneasr:
Maybe I misread, but this statement says that the code won't be downloaded:

> The module download is split into two parts: downloading the go.mod and downloading the actual code. If you have dependencies only needed for tests, then they will show up in your go.mod, and go get will download their go.mods, but it will not download their code. The test-only dependencies get downloaded only when you need it, such as the first time you run go test.

PR #381, opened 2019-09-10
allow private client to pass PKCE related parameters and verify them.

damienbr:


aeneasr:
Any reason this was closed? :) We can definitely add this to fosite

damienbr:
I opened it by mistake but I'll will prepare a PR. Some error messages
needed to be updated.

Le lun. 16 sept. 2019 à 11:07, hackerman <notifications@github.com> a
écrit :

PR #386, opened 2019-10-02
fosite.Request: Update Request ID in request.Merge()

silverspace:
Updating `Request.Merge()` to also merge the request ID from the passed in
request, since in all cases where `Merge()` is called, the receiver object
has an uninitialized `ID` value. This fixes an issue where
`introspectAccessToken()` is returning the wrong request ID: 

1. `Fosite.IntrospectToken()` creates a new `ar` object with no initialized ID:
https://github.com/ory/fosite/blob/master/introspect.go#L59
2. `Fosite.IntrospectToken` calls `CoreValidator.IntrospectToken()`:
https://github.com/ory/fosite/blob/master/introspect.go#L61
3. `CoreValidator.IntrospectToken()` calls `CoreValidator.introspectAccessToken()`:
https://github.com/ory/fosite/blob/master/handler/oauth2/introspector.go#L58
4. `introspectAccessToken` reads the access token session from storage. The returned `or` object has the correct RequestID:
https://github.com/ory/fosite/blob/master/handler/oauth2/introspector.go#L83
5. `introspectAccessToken` merges `accessRequest` with `or`, but the call to `Merge()` does not merge the RequestID from `or.GetID()`.
https://github.com/ory/fosite/blob/master/handler/oauth2/introspector.go#L112
6. The returned `accessRequest` object has an uninitialized `ID`, so the next call to `GetID()` will return a random UUID.

Steps to reproduce:
1. Create a new access token for RequestID 'requestId`. 
2. Introspect the token via
```
tt, ar, err := oauth2Provider.IntrospectToken(...)
```
3. Observe that `ar.GetID()` is not the original `requestId` value. 

## Proposed changes

Set the `RequestID` from the passed in `request` object in `Request.Merge()`:
https://github.com/ory/fosite/blob/master/request.go#L168

## Checklist

- [ ] I have read the [security policy](../security/policy)
- [ ] I confirm that this pull request does not address a security vulnerability. If this pull request addresses a security

### DIFF
--- a/request.go
+++ b/request.go
@@ -165,6 +165,7 @@ func (a *Request) Merge(request Requester) {
 		a.GrantAudience(aud)
 	}
 
+	a.ID = request.GetID()
 	a.RequestedAt = request.GetRequestedAt()
 	a.Client = request.GetClient()
 	a.Session = request.GetSession()


aeneasr:
This looks good to me! Could you also add a test so this doesn't break in a future patch?

PR #399, opened 2020-03-03
fosite.Arguments: Nano optimization of .Exact()

redbaron:
Previously Arguments.Exact had vague semantic where
it coudln't distinguish between value with a space and multiple
values. Split it into 2 functions with clear semantic.
   
Old .Exact() remains for compatibility and marked as deprecated

### DIFF
--- a/arguments.go
+++ b/arguments.go
@@ -57,6 +57,23 @@ func (r Arguments) HasOneOf(items ...string) bool {
 	return false
 }
 
+// Deprecated: Use ExactOne, Matches or MatchesExact
 func (r Arguments) Exact(name string) bool {
 	return name == strings.Join(r, " ")
+}
+
+func (r Arguments) ExactOne(name string) bool {
+	return len(r) == 1 && r[0] == name
+}
+
+func (r Arguments) MatchesExact(items ...string) bool {
+	if len(r) != len(items) {
+		return false
+	}
+	for i, item := range items {
+		if item != r[i] {
+			return false
+		}
+	}
+	return false
 }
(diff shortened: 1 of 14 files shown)


aeneasr:
These are functionally different because before it worked for len > 1 and now it doesn’t.

redbaron:
>These are functionally different because before it worked for len > 1 and now it doesn’t.

for len>1 before and after it returns `false`, as it should, no?

redbaron:
Only case I can see them being different is that previously `.Exact()` incorrectly returned `true` for following case:

```
        var v Arguments = []string{"A", "B"}
	v.Exact("A B") 
```

do I miss something?

aeneasr:
```
        var v Arguments = []string{"A", "B"}
	v.Exact("A B") 
```

that is the use case I was thinking about. While it does not appear to be used like that anywhere in fosite, it was originally implemented that way because we might have arguments where order matters / must be preserved. I'm not sure if we import this code somewhere else that depends on this behavior, thus I am hesitant to merge it.

redbaron:
> we might have arguments where order matters 

This should be covered by `func (r Arguments) ExactSlice(names []string) bool`

Using `.Exact([]strings.Join(names," "))` for this purpose is clear misuse

aeneasr:
That might be, but as I explained we have other code that (might) depend on this behavior and even if you have a better naming for it, it doesn't change that fact! Breaking the functionality in such a way without e.g. renaming the function as well is not something that can be merged.

If you want, you can introduce a new function `ExactOne(name string)` and `ExactSlice` and leave the existing one untouched, updating all the code where currently `Exact` is used and mark `Exact` as deprecated using godoc.

redbaron:
Added ExactOne and MatchesExact, also added tests for discussed edge cases

aeneasr:
Awesome, thank you!

Issue #284, opened 2018-05-29
Not compliant with RFC8252 section 7.3 (OAuth 2.0 for Native Apps)


koesie10:
According to the RFC, ports should be ignored for loopback addresses for a public client and both IPv4 and IPv6 should be allowed.

https://tools.ietf.org/html/rfc8252#section-7.3

PR #400, opened 2020-03-03
RFC8252#section-7.3 Loopback Interface Redirection

raffis:
This implements support for a dynamic port request from an oauth2 client.
RFC reference: https://tools.ietf.org/html/rfc8252#section-7.3

## Related issue
#284
https://github.com/ory/hydra/issues/1732

## Checklist

## Further comments
The RFC (https://tools.ietf.org/html/rfc8252#section-7.3) states an URI like `http://127.0.0.1:{port}/{path}` and `http://[::1]:{port}/{path}`. I am not 100% sure if this is a must or just a regex example. 
I have implemented this in fosil that it is now possible to register:

* http://127.0.0.1
* http://[::1]

... and if a registered client has one or both of these redirect uri's a client may request a dynamic port onto the loopback interface. For example http://127.0.0.1:9988 or with a path http://127.0.0.1:9988/callback. Note as with this PR, if both IPv4 and IPv6 should be supported for redirect loopback uris the clients needs to be registered with both.

If the RFC requests that the redirect URI must exactly be `http://127.0.0.1:{port}/{path}` then more code needs to be changed, for example the client registration in hydra would also fail with this (invalid) uri.

### DIFF
--- a/authorize_helper.go
+++ b/authorize_helper.go
@@ -112,21 +112,41 @@ func MatchRedirectURIWithClientRedirectURIs(rawurl string, client Client) (*url.
 // Loopback redirect URIs use the "http" scheme and are constructed with
 // the loopback IP literal and whatever port the client is listening on.
 func isMatchingRedirectURI(uri string, haystack []string) bool {
+	requested, err := url.Parse(strings.ToLower(uri))
+	if err != nil {
+		return false
+	}
+
 	for _, b := range haystack {
 		l := strings.ToLower(b)
-		if l == strings.ToLower(uri) || isLoopbackURI(uri, l) {
+		if l == requested.String() || isLoopbackURI(requested, l) {
 			return true
 		}
 	}
 	return false
 }
 
-func isLoopbackURI(uri string, registeredURI string) bool {
-	if registeredURI != "http://127.0.0.1" && registeredURI != "http://[::1]" {
+func isLoopbackURI(requested *url.URL, registeredURI string) bool {
+	registered, err := url.Parse(registeredURI)
+	if err != nil {
 		return false
 	}
 
-	match, _ := regexp.MatchString("http://(127.0.0.1|\\[::1\\]):?(\\d+)?", uri)
+	if registered.Scheme != "http" || !isLoopbackAddress(registered.Host) {
+		return false
+	}
+
+	if requested.Scheme == "http" && isLoopbackAddress(requested.Host) && registered.Path == requested.Path {
+		return true
+	}
+
+	return false
+}
+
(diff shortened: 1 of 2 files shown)


authorize_helper_test.go
```diff
@@ -154,6 +154,46 @@ func TestDoesClientWhiteListRedirect(t *testing.T) {
...
+			isError: true,
+		},
+		{
+			client:   &DefaultClient{RedirectURIs: []string{"http://127.0.0.1"}},
+			url:      "http://127.0.0.1:1024",
+			isError:  false,
+			expected: "http://127.0.0.1:1024",
+		},
+		{
+			client:   &DefaultClient{RedirectURIs: []string{"http://127.0.0.1"}},
```
aeneasr: I believe that this should fail because the whitelisted redirect URL does not have the `/cb` path segment. The spec only talks about randomly assigned ports but does not leave room for variable paths, if I understand correctly. Otherwise, this looks pretty good!

raffis: Sounds right. I had to change some bits because of that. Using url parse now to extract the paths and compare them.

aeneasr:
You rock, thank you for your contribution!

authorize_helper.go
```diff
@@ -95,6 +96,60 @@ func MatchRedirectURIWithClientRedirectURIs(rawurl string, client Client) (*url.
...
+// https://tools.ietf.org/html/rfc8252#section-7.3
+// Native apps that are able to open a port on the loopback network
+// interface without needing special permissions (typically, those on
+// desktop operating systems) can use the loopback interface to receive
+// the OAuth redirect.
+//
+// Loopback redirect URIs use the "http" scheme and are constructed with
+// the loopback IP literal and whatever port the client is listening on.
+func isMatchingRedirectURI(uri string, haystack []string) bool {
+	requested, err := url.Parse(strings.ToLower(uri))
```
aeneasr: Since we are now using url.Parse, toLower is no longer required and actually might cause issues for servers that have case sensitive paths. I think toLower should be removed.

aeneasr:
```suggestion
	requested, err := url.Parse(uri)
```

raffis:
> You rock, thank you for your contribution!

Thanks, same to you!

authorize_helper.go
raffis: Right, fixed it. (Had to add another fix, with the fix above test 8 would have failed).

Thanks for merging!

aeneasr:
I released a new tag for this so this can now be added to hydra

PR #417, opened 2020-04-03
fix: Not part of any interface

mitar:
#346  Proposed changes

`DeleteAuthorizeCodeSession` and `CreateImplicitAccessTokenSession` not part of any storage interface. So deleting them to not confuse people to think they have to implement them.

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md)
- [x] I have read the [security policy](../security/policy)
- [x] I confirm that this pull request does not address a security vulnerability. If this pull request addresses a security
vulnerability, I confirm that I got green light (please contact [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push the changes.
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation within the code base (if appropriate)

PR #422, opened 2020-04-06
fix: Further cleanup of memory storage

mitar:
Followup to https://github.com/ory/fosite/pull/417.

## Proposed changes

I just realized that I think this is not needed as well.

## Checklist

### DIFF
--- a/storage/memory.go
+++ b/storage/memory.go
@@ -40,7 +40,6 @@ type MemoryStore struct {
 	AuthorizeCodes  map[string]StoreAuthorizeCode
 	IDSessions      map[string]fosite.Requester
 	AccessTokens    map[string]fosite.Requester
-	Implicit        map[string]fosite.Requester
 	RefreshTokens   map[string]fosite.Requester
 	PKCES           map[string]fosite.Requester
 	Users           map[string]MemoryUserRelation
@@ -56,7 +55,6 @@ func NewMemoryStore() *MemoryStore {
 		AuthorizeCodes:         make(map[string]StoreAuthorizeCode),
 		IDSessions:             make(map[string]fosite.Requester),
 		AccessTokens:           make(map[string]fosite.Requester),
-		Implicit:               make(map[string]fosite.Requester),
 		RefreshTokens:          make(map[string]fosite.Requester),
 		PKCES:                  make(map[string]fosite.Requester),
 		Users:                  make(map[string]MemoryUserRelation),


storage/memory.go
```diff
@@ -79,15 +77,15 @@ func NewExampleStore() *MemoryStore {
...
+				ResponseTypes: []string{"id_token", "code", "token"},
 				GrantTypes:    []string{"implicit", "refresh_token", "authorization_code", "password", "client_credentials"},
 				Scopes:        []string{"fosite", "openid", "photos", "offline"},
 			},
 			"encoded:client": &fosite.DefaultClient{
 				ID:            "encoded:client",
 				Secret:        []byte(`$2a$10$A7M8b65dSSKGHF0H2sNkn.9Z0hT8U1Nv6OWPV3teUUaczXkVkxuDS`), // = "encoded&password"
 				RedirectURIs:  []string{"http://localhost:3846/callback"},
-				ResponseTypes: []string{"id_token", "code", "token", "id_token token", "code id_token", "code token", "code id_token token"},
+				ResponseTypes: []string{"id_token", "code", "token"},
```
aeneasr, line 88: This was just merged in another PR, why revert it again?

mitar: Oh, sorry. This is some noise. I will fix this. Only the `Implicit` lines above were meant to be in this PR.

aeneasr: No problem!

mitar: Fixed.

aeneasr: Thanks!

aeneasr:
In case someone looks this up:

Several issues have been introduced by this PR which prevented consumers of this app to compile - resolved with and 0.30.2: https://github.com/ory/fosite/pull/428

mitar:
Oh, sorry? But what was the problem? `Implicit` was not used anywhere? Have I missed something?

aeneasr:
No problem! This was also not for blaming but just people having checked out the commit hash and looking for guidance :)

Yes, it was still being used [here](https://github.com/ory/fosite/pull/428/files#diff-e744e1f8d087662138eddb8ce79489edL72) and [here](https://github.com/ory/fosite/pull/428/files#diff-2232580924688d6d8024941eba639173L102) :)

Oversights like these can happen easily. By the way, I'm using Goland and I know some people hate JVM-based IDEs but the code inspection tools are simply superior, even if the IDE consumes 4GB RAM and is laggish when you have 5+ projects open (but hey what else am I supposed to do with those 16GB RAM :D )

aeneasr:
Oh and the CI was broken for some reason, including linting and testing of subpackages - which is why this wasn't caught by the CI. The reason being that I needed to refresh TravisCI's access to ORY's github repos.

mitar:
Oh yes, I also use Goland (in fact I use InteliJ with Go plugin, because I need multiple languages). They are doing great IDEs. This is how I found this and checked. But then I just used GitHub editor to edit this directly on website and I think I missed updating all places I have seen in the IDE. Bummer.

Yes, I think the future is in integrated CIs with repository hosting (like what GitHub does with Actions, and what GitLab has). I have been using GitLab for this reason for most of my projects now. Because all this permission issues between 3rd party CI and repository hosting is always problematic. Had those issues in the past as well. Same with Docker Hub. :-(

Issue #370, opened 2019-06-26
Add configuration to return refresh tokens without requiring "offline" or "offline_access" scopes


dmcinnes:
**Is your feature request related to a problem? Please describe.**

Hello! I've started working on a PR for this but I should first see if it fits in with your roadmap.

We have a requirement for our service to always return refresh tokens with authorization code grant exchange, however the fosite library requires the `offline` or `offline_access` scopes. For our needs it would be great if the scope is not required.

**Describe the solution you'd like**

An additional bool configuration to the [compose.Config struct](https://github.com/ory/fosite/blob/master/compose/config.go#L31-L89) to turn on always returning refresh tokens. This would need to be passed down to the relevant handlers.

**Describe alternatives you've considered**

An alternative for us would be to modify existing requests to insert the "offline_access" scope, however this is not ideal if we expose the authorizations and their scopes elsewhere.

**Additional context**

I will happily submit a PR if you are are to this idea (I already have one in progress as mentioned above).

aeneasr:
How about a field `RefreshTokenScope: []string` which is per default `offline` / `offline_access` but can be left empty or set to empty to disable the scope. Please keep in mind that this change **must** be backwards compatible.

dmcinnes:
@aeneasr Ah I like it! I'll get that a go, thanks!

mitar:
I think this has been done?

PR #430, opened 2020-05-06
fix: Do not issue refresh tokens that can't be used

pjcdawkins:
## Related issue

No prior discussion, but this is somewhat related to #370, which makes it possible for clients to be issued a refresh token without requesting a scope.

There may be clients which may not be allowed to use the `refresh_token` grant type, but which will currently be issued a refresh token during the authorization code flow. Disallowing the refresh_token grant type might be particularly common when there are public clients (using this flow with PKCE).

It is impossible to use the issued refresh token when the client does not have the grant type, and no other client can use it either. It would be neater to avoid issuing refresh tokens that cannot be used.

A look at the OAuth 2.0 and OpenID Connect specs shows nothing particularly relevant, except RFC 6749's:

> Issuing a refresh token is optional at the discretion of the authorization server.

## Proposed changes

Do not issue refresh tokens to clients without the "refresh_token" grant type.

## Checklist

- [x] I confirm that this pull request does not address a security vulnerability

### DIFF
--- a/handler/oauth2/flow_authorize_code_token.go
+++ b/handler/oauth2/flow_authorize_code_token.go
@@ -118,6 +118,18 @@ func (c *AuthorizeExplicitGrantHandler) HandleTokenEndpointRequest(ctx context.C
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


aeneasr:
Nice find, thanks!

Issue #389, opened 2020-01-08
Discussion: Separated EnforcePKCE settings for public and confidential clients ?


nerocrux:
**Is your feature request related to a problem? Please describe.**

In this PR, set EnforcePKCE to true also forces PKCE to be enabled for confidential clients.
https://github.com/ory/fosite/pull/382

I am totally agree that allow PKCE to be enabled on confidential clients is a good feature.

However I think in some cases, we want PKCE to be only force enabled on public clients, but not confidential clients.

Because confidential clients require client authentication and public clients doesn't, those two types of clients have different security level in the first place.

Of course, enable PKCE on confidential clients will make them more secure, but also make it more difficult to implement.

It will be good if we can set PKCE policy separately on public and confidential clients.

**Describe the solution you'd like**

Separate EnforcePKCE option by client types.
For example (to avoid breaking change)

* EnforcePKCE
* DisableConfidentialClientPKCE

**Describe alternatives you've considered**

**Additional context**

My usecase is basically

* Always enforce PKCE for public clients
* Selectively enforce PKCE for confidential clients (set EnforcePKCE flag to false and validate on the application side)

aeneasr:
Good idea!

nerocrux:
@aeneasr 
Thank you.
If you like it, I will try to write some codes :D

aeneasr:
We could also add a flag to OAuth2 Client definitions saying whether it's enforced or not?

aeneasr:
But I think your suggestion is easier to manage from an admin perspective and solves a very basic use case.

nerocrux:
> We could also add a flag to OAuth2 Client definitions saying whether it's enforced or not?
> But I think your suggestion is easier to manage from an admin perspective and solves a very basic use case.

Yeah, very true.
Add a flag to OAuth2 Client seems to be a better solution because it's more flexible.
I will try to dig more to see which solution is better.

aeneasr:
I think your proposal is good. We can always add the flag later :)

nerocrux:
Thank you!
Then I will try to implement that way.

PR #431, opened 2020-05-07
feat: Add a flag to enforce PKCE only for public clients

pjcdawkins:
## Related issue

Alternative proposal for the issue discussed in #389 and #391, where enforcement of PKCE is wanted only for certain clients.

## Proposed changes

Add a new flag `EnforcePKCEForPublicClients` which enforces PKCE only for public clients. The error hint is slightly different, as it mentions PKCE is enforced for "this client" rather than "clients". (It intentionally does not mention why it's enforced, as I think basing it on public clients is an implementation detail that servers may want to change without adding to the error hints)

I didn't use an enum, as suggested in #391, because I imagine it would be a backwards compatibility problem. I could be convinced otherwise.

## Checklist

- [x] I confirm that this pull request does not address a security vulnerability


aeneasr:
Thank you! I'll release this right after merge if you need this in Hydra :)

PR #434, opened 2020-05-22
fix: double-decoding of client credentials in request body

pjcdawkins:
I noticed that client credentials are URL-decoded after being extracted from the POST body form, which was already URL-decoded by Go. The accompanying error message suggests this was copied and pasted from the HTTP basic authorization header handling, which is the only place where the extra URL-decoding was needed (as per the OAuth 2.0 spec). The result is that client credentials containing `%`-prefixed sequences, whether valid sequences or not, are going to fail validation.

## Proposed changes

Remove the extra URL decoding. Add tests that ensure client credentials work with special characters in both the HTTP basic auth header and in the request body.

## Checklist

### DIFF
--- a/client_authentication_test.go
+++ b/client_authentication_test.go
@@ -66,6 +66,16 @@ func mustGenerateNoneAssertion(t *testing.T, claims jwt.MapClaims, key *rsa.Priv
 	return tokenString
 }
 
+// returns an http basic authorization header, encoded using application/x-www-form-urlencoded
+func clientBasicAuthHeader(clientID, clientSecret string) http.Header {
+	creds := url.QueryEscape(clientID) + ":" + url.QueryEscape(clientSecret)
+	return http.Header{
+		"Authorization": {
+			"Basic " + base64.StdEncoding.EncodeToString([]byte(creds)),
+		},
+	}
+}
+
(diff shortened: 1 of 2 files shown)


pjcdawkins:
Separated the commits just to demonstrate the test works, happy to squash if you need

client_authentication_test.go
```diff
@@ -80,6 +80,10 @@ func TestAuthenticateClient(t *testing.T) {
 	barSecret, err := hasher.Hash(context.TODO(), []byte("bar"))
 	require.NoError(t, err)
 
+	// a secret containing literal characters that would be affected by double-URL-decoding.
+	percentSecret, err := hasher.Hash(context.TODO(), []byte("%66%6F%6F"))
```
aeneasr, line 84: Can we add a special char here as well, just to make sure this works correctly? Something like `<` or `$` or `§`

pjcdawkins: Done - for POST and basic auth

aeneasr:
Awesome, thank you!

aeneasr:
https://github.com/ory/fosite/releases/tag/v0.32.0

pjcdawkins:
@aeneasr thanks for releasing, just a note the release refers to "double-encoding", that should be double-decoding

PR #446, opened 2020-06-25
feat: New compose strategies for ES256

mitar:
I realized those are missing.

This now completes support for ES256.


aeneasr:

PR #454, opened 2020-07-30
refactor: Copy all values when sanitizing

mitar:
Minor improvement when looking at the code. `Get` returns just the first value.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,6 @@
-This is a list of breaking changes. As long as `1.0.0` is not released, breaking changes will be addressed as minor version
-bumps (`0.1.0` -> `0.2.0`).
+**THIS DOCUMENT HAS MOVED**
+
+This file is no longer being updated and kept for historical reasons. Please check the [CHANGELOG](changelog.md) instead!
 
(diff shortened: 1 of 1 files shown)


request.go
```diff
@@ -186,8 +186,8 @@ func (a *Request) Sanitize(allowedParameters []string) Requester {
 	b.ID = a.GetID()
 	b.Form = url.Values{}
 	for k := range a.Form {
```
aeneasr, line 188: Typically only the first value is set in `url.Values` but this is still a good merge. It would be nice though to have a test case that covers the fixed behavior to prevent regression.

mitar:
> Are you by chance following your fosite fork on CircleCI? In those cases, builds do not work in the upstream which is why the CI does not trigger. It would be great to have the CI results though to make sure that everything behaves correctly!

Huh, what? So strange that this matters.

In fact it does not. It seems it is about my master branch being out of date in my fork.

mitar:
Anyway, trying to rebase killed my branch. I will redo this.

mitar:
Done in https://github.com/ory/fosite/pull/455.

Issue #456, opened 2020-08-08
Make error_hint be part of error_description


mitar:
`error_hint` is a non-standard field being returned by fosite. Because it is not standard, standard clients do not expose it at all to developers using our system have hard time understanding what they have to change to fix the error. [RFC6749 says](https://tools.ietf.org/html/rfc6749):

> OPTIONAL.  Human-readable ASCII [USASCII] text providing additional information, used to assist the client developer in understanding the error that occurred.

I think this is exactly what `error_hint` currently is. It seems `error_description` is currently simply a standard description of the `error_code`, but it makes no sense that it is such: anyone can map `error_code` to error descriptions, ideally in the target language. On other hand, `error_description` is not meant to be shown to the user, but to the developer. So English is good there.

Similar is [written here](https://www.oauth.com/oauth2-servers/server-side-apps/possible-errors/):

> Despite the fact that servers return an error_description key, the error description is not intended to be displayed to the user. Instead, you should present the user with your own error message. This allows you to tell the user an appropriate action to take to correct the problem, and also gives you a chance to localize the error messages if you’re building a multi-language website.

So, `error_description` is for developers, `error_code` is for end-users (to be mapped to error message by the client).

So, my proposal would be that when responding, fosite would return in `error_description` what is currently `error_description` and `error_hint` together. We could leave the code mostly as it is and only when marshaling the error we would do something like `<error_description>: <error_hint>` when there is error hint.

aeneasr:
Yeah this makes sense! It would be important to keep the existing behavior around for a while to allow clients to migrate to the new error layout but other than that I have no problem with this.

The error description currently comes from here: https://tools.ietf.org/html/rfc6749#section-4.1.2.1

In general I think that better human readable error responses would greatly help people use this framework and strain support less as no cryptic errors like "invalid_request" without any additional detail is being sent (e.g. "you used application/json but we expected www-form")

mitar:
> Yeah this makes sense! It would be important to keep the existing behavior around for a while to allow clients to migrate to the new error layout but other than that I have no problem with this.

So, let's keep `error_hint` around as a field, but also append it to `error_description` when `error_hint` is available, would that be OK?

What about `error_debug`, should we also append this to `error_description`, when enabled to expose it in the configuration? (And keep it as a separate field?)

> In general I think that better human readable error responses would greatly help people use this framework and strain support less as no cryptic errors like "invalid_request" without any additional detail is being sent (e.g. "you used application/json but we expected www-form")

[...]

PR #460, opened 2020-08-13
feat: error_hint and error_debug are now exposed through error_description

mitar:
## Related issue

Fixes https://github.com/ory/fosite/issues/456.

## Proposed changes

When `error_description` is send out, it includes now `error_hint` (if present) and `error_debug` (if present and enabled to be send). Because `error_description` is standard this allows easier access to that information.

`error_hint` and `error_debug` fields are left as-is for backwards compatibility purposes. `error_verbose` is added to provide only the old error description.

## Checklist

      the changes.


access_error.go
```diff
@@ -34,10 +34,19 @@ func (f *Fosite) WriteAccessError(rw http.ResponseWriter, _ AccessRequester, err
 func (f *Fosite) writeJsonError(rw http.ResponseWriter, err error) {
 	rw.Header().Set("Content-Type", "application/json;charset=UTF-8")
 
-	rfcerr := ErrorToRFC6749Error(err)
+	// We dereference to make a copy and do not change passed err when it is already RFC6749Error.
+	rfcerr := *ErrorToRFC6749Error(err)
```
mitar, line 38: I think this should be done everywhere around the codebase. I see often the pattern like:

```
rfcerr := ErrorToRFC6749Error(err)
rfcerr.Debug = ""
```

This in fact modifies the value in the original `err`. So if you try to log it to Prometheus or something similar afterwards, you do not have `Debug` anymore.

access_error.go
```diff
@@ -34,10 +34,19 @@ func (f *Fosite) WriteAccessError(rw http.ResponseWriter, _ AccessRequester, err
...
 	if !f.SendDebugMessagesToClients {
 		rfcerr.Debug = ""
 	}
+	// We expose both error hint and debug strings through standard error description, too
+	// (they are non-standard fields and some clients do not show them).
+	if rfcerr.Hint != "" {
+		rfcerr.Description += ": " + rfcerr.Hint
```
mitar, line 45: It might be better to add `GetDescription` to rfcerr?

aeneasr: Depends, `GetDescription` will not be json marshalled!

access_error.go
```diff
@@ -34,10 +34,19 @@ func (f *Fosite) WriteAccessError(rw http.ResponseWriter, _ AccessRequester, err
...
+	}
+	if rfcerr.Debug != "" {
+		rfcerr.Description += " (" + rfcerr.Debug + ")"
```
aeneasr, line 48: So debug might contain internal information such as `UNIQUE CONSTRAINT VIOLATION` or other database or sensitive details. Let's keep the debug semantics as is :)

aeneasr: Oh, I see - https://github.com/ory/fosite/pull/460/files#diff-2ac6e970b8a496524d4443819d71fa8cR62 solves this by checking the config. We should do that here as well!

authorize_error.go
```diff
@@ -49,17 +49,21 @@ func (f *Fosite) WriteAuthorizeError(rw http.ResponseWriter, ar AuthorizeRequest
...
-
+	// We expose both error hint and debug strings through standard error description, too
+	// (they are non-standard fields and some clients do not show them).
 	if rfcerr.Hint != "" {
 		query.Add("error_hint", rfcerr.Hint)
+		error_description += ": " + rfcerr.Hint
+	}
+	if f.SendDebugMessagesToClients && rfcerr.Debug != "" {
+		query.Add("error_debug", rfcerr.Debug)
+		error_description += " (" + rfcerr.Debug + ")"
```
aeneasr, line 64: We could make this also more readable by using newlines:

```
${description}

${description}

Debug log: ${debug}
```

access_error.go
```diff
@@ -34,10 +34,19 @@ func (f *Fosite) WriteAccessError(rw http.ResponseWriter, _ AccessRequester, err
...
 
-	rfcerr := ErrorToRFC6749Error(err)
+	// We dereference to make a copy and do not change passed err when it is already RFC6749Error.
+	rfcerr := *ErrorToRFC6749Error(err)
 	if !f.SendDebugMessagesToClients {
 		rfcerr.Debug = ""
 	}
+	// We expose both error hint and debug strings through standard error description, too
+	// (they are non-standard fields and some clients do not show them).
+	if rfcerr.Hint != "" {
```
aeneasr, line 44: See idea below for newlines!

aeneasr (commented):
Great work! I think this will appear in a couple of other places as well. I also had the idea that we could improve all of this a bit more by making RFCError a `JSONUnmarshaller` and add a method `AsQueryString() string` which would do the modifications instead of handling this. The only problem would be that `f.SendDebugMessagesToClients` would need to be passed to these functions somehow.

aeneasr:
Once again, sorry for the late review! By the way, if you allow edits from maintainers I should be able to synch the branches and make the nancy failure go away :)

mitar:
> By the way, if you allow edits from maintainers I should be able to synch the branches and make the nancy failure go away

It should already be on.

aeneasr:
Yes, my bad, I had to change something in the repository settings!

access_error.go
mitar: See also this comment. Should I update this as well around?

access_error.go
mitar: We are, above we set:

```
	if !f.SendDebugMessagesToClients {
		rfcerr.Debug = ""
	}
```

authorize_error.go
mitar: Sure.

access_error.go
mitar: Yea, but you would use it to construct the JSON.

access_error.go
aeneasr: Yes, it makes sense for `ErrorToRFC6749Error` to return by value, not by ref. Please ensure that all method receivers operate on the value as well, except for the ones who modify the values.

mitar:
About `JSONUnmarshaller`, yea, I have no idea how to pass this, unless we do something like:

```
if !config.SendDebugMessagesToClients {
   error = error.WithDebug("")
}
```

And then leave to `JSONUnmarshaller` to serialize.

access_error.go
aeneasr: I don't follow? This is JSON Encoded a few lines later: https://github.com/ory/fosite/pull/460/files#diff-49ae52d32b52cf990340c8dbf949704fR53

access_error.go
aeneasr: Maybe also rename this to `MakeRFC6749ErrorCopy`

access_error.go
aeneasr: Ah I missed that!

mitar:
Or maybe we make `Sanitize` method on the error, which does the above, accepting boolean if it should sanitize or not. And then you call `error.Sanitize(!config.SendDebugMessagesToClients)` or something.

aeneasr:
Hm yeah in that case it probably makes sense to just keep it as is!

aeneasr:
Just to double-check, this would now be good for merge right?

mitar:
No, I have not yet updated it based on your comments.

Not sure why I do not see a way to convert this PR back to draft mode, hm. So I edited the title.

aeneasr:
Ok sounds good :)

access_error.go
mitar: I made `Sanitize` method which does the copy and sanitization.

access_error.go
```diff
@@ -38,12 +38,16 @@ func (f *Fosite) writeJsonError(rw http.ResponseWriter, err error) {
 
 	rfcerr := ErrorToRFC6749Error(err)
 	if !f.SendDebugMessagesToClients {
-		rfcerr.Debug = ""
+		rfcerr = rfcerr.Sanitize()
 	}
 
-	js, err := json.Marshal(rfcerr)
+	js, err := json.MarshalIndent(rfcerr, "", "\t")
```
mitar, line 44: I use `MarshalIndent` here so that both authorize error and access error are represented in the same way.

access_error.go
```diff
@@ -38,12 +38,16 @@ func (f *Fosite) writeJsonError(rw http.ResponseWriter, err error) {
...
 	if err != nil {
-		http.Error(rw, fmt.Sprintf(`{"error": "%s"}`, err.Error()), http.StatusInternalServerError)
+		if !f.SendDebugMessagesToClients {
+			http.Error(rw, fmt.Sprintf(`{\n\t"error": "server_error",\n\t"error_description": "%s"\n}`, err.Error()), http.StatusInternalServerError)
+		} else {
+			http.Error(rw, `{\n\t"error": "server_error"\n}`, http.StatusInternalServerError)
```
mitar, line 49: So this was bad and I fixed it:
* It was not a proper OAuth response (`error` should be one of error constants, not exception string).
* It might leak sensitive data.

authorize_error.go
```diff
@@ -34,47 +34,49 @@ func (f *Fosite) WriteAuthorizeError(rw http.ResponseWriter, ar AuthorizeRequest
...
+
 	if !ar.IsRedirectURIValid() {
-		if !f.SendDebugMessagesToClients {
-			rfcerr.Debug = ""
-		}
+		rw.Header().Set("Content-Type", "application/json;charset=UTF-8")
 
 		js, err := json.MarshalIndent(rfcerr, "", "\t")
 		if err != nil {
-			http.Error(rw, err.Error(), http.StatusInternalServerError)
```
mitar, line 44: This should probably always be JSON, not that sometimes it is just some plain test string.

authorize_error.go
```diff
@@ -34,47 +34,49 @@ func (f *Fosite) WriteAuthorizeError(rw http.ResponseWriter, ar AuthorizeRequest
...
+	// The endpoint URI MUST NOT include a fragment component.
+	redirectURI.Fragment = ""
+
+	query := rfcerr.ToValues()
+	query.Add("state", ar.GetState())
 
+	var redirectURIString string
 	if !(len(ar.GetResponseTypes()) == 0 || ar.GetResponseTypes().ExactOne("code")) && errors.Cause(err) != ErrUnsupportedResponseType {
-		redirectURI.Fragment = query.Encode()
+		redirectURIString = redirectURI.String() + "#" + query.Encode()
```
mitar, line 69: We have to manually URL encode the fragment (by the spec) and append it. Previously, we double escaped stuff here. (There is no `RawFragment` before go 1.15, and `RawFragment` in go 1.15 is different in behavior from `RawQuery`.)

errors.go
```diff
@@ -317,3 +319,72 @@ func (e *RFC6749Error) WithDescription(description string) *RFC6749Error {
...
+	if e.Debug != "" {
+		description += "\n\n" + e.Debug
+	}
+	return description
+}
+
+// RFC6749ErrorJson is a helper struct for JSON encoding/decoding of RFC6749Error.
+type RFC6749ErrorJson struct {
+	Name        string `json:"error"`
+	Verbose     string `json:"error_verbose"`
```
mitar, line 344: Feel free to suggest any other name. I wanted to to keep the original description around so that we can define `UnmarshalJSON`.

introspect.go
```diff
@@ -53,18 +53,18 @@ func AccessTokenFromRequest(req *http.Request) string {
 }
 
 func (f *Fosite) IntrospectToken(ctx context.Context, token string, tokenType TokenType, session Session, scopes ...string) (TokenType, AccessRequester, error) {
-	var found bool = false
+	var found = false
```
mitar, line 56: This is not necessary.

introspect.go
```diff
@@ -53,18 +53,18 @@ func AccessTokenFromRequest(req *http.Request) string {
...
 		tt, err := validator.IntrospectToken(ctx, token, tokenType, ar, scopes)
-		if err := errors.Cause(err); err == nil {
+		if err == nil {
 			found = true
 			foundTokenType = tt
-		} else if err.Error() == ErrUnknownRequest.Error() {
+		} else if err.Error() == ErrUnknownRequest.Error() || errors.Cause(err).Error() == ErrUnknownRequest.Error() {
 			// Nothing to do
-		} else if err != nil {
+		} else {
```
mitar, line 67: This condition was unnecessary and always true.

introspect.go
```diff
@@ -53,18 +53,18 @@ func AccessTokenFromRequest(req *http.Request) string {
...
 func (f *Fosite) IntrospectToken(ctx context.Context, token string, tokenType TokenType, session Session, scopes ...string) (TokenType, AccessRequester, error) {
-	var found bool = false
+	var found = false
 	var foundTokenType TokenType = ""
 
 	ar := NewAccessRequest(session)
 	for _, validator := range f.TokenIntrospectionHandlers {
 		tt, err := validator.IntrospectToken(ctx, token, tokenType, ar, scopes)
-		if err := errors.Cause(err); err == nil {
+		if err == nil {
```
mitar, line 62: I simplified the logic here a bit. `ErrorToRFC6749Error` uses `Cause` internally as well, so I think this now is clearer and easier to read.

mitar:
@aeneasr I finished things from my side. Please review again. When merging, please squash commits. (Or do you want me to squash them?)

aeneasr:
All good, I'll squash them :)

access_error.go
aeneasr: I would expect JSON APIs to respond with machine readable code (minimized) as opposed to human readable (indented, formatted). Any particular reason why we would respond with human readable formats here?

authorize_error.go
aeneasr: Correct!

errors.go
```diff
@@ -265,11 +267,11 @@ func ErrorToRFC6749Error(err error) *RFC6749Error {
 }
 
 type RFC6749Error struct {
-	Name        string `json:"error"`
-	Description string `json:"error_description"`
-	Hint        string `json:"error_hint,omitempty"`
-	Code        int    `json:"status_code,omitempty"`
-	Debug       string `json:"error_debug,omitempty"`
```
aeneasr, line 272: If you want to make it explicit that these are not marshalled, you could add:

`json:“-“`

errors.go
aeneasr: I would probably expect verbose to be more verbose than the other description. Given that we‘re changing the API in a breaking manner and that all OAuth2 clients who rely on this original prameter would need to be updated anyways I think we can simply omit it :)

introspect.go
```diff
@@ -53,18 +53,18 @@ func AccessTokenFromRequest(req *http.Request) string {
...
 			found = true
 			foundTokenType = tt
-		} else if err.Error() == ErrUnknownRequest.Error() {
+		} else if err.Error() == ErrUnknownRequest.Error() || errors.Cause(err).Error() == ErrUnknownRequest.Error() {
```
aeneasr, line 65: How about errors.Is instead?

access_error.go
mitar: It was like that in the authorize error handler already. I just unified it to be the same. I do not really care (except that it will be a bit more work for me to change it) but I just wanted to be uniform. maybe `MarshalIndent` was used for errors so that it is easier to debug when manually inspecting the response when debugging.

errors.go
mitar: I do not care about this message being marshaled. If somebody wants to do it for some other reasons (like storing in the database), I think they should be able. But it is not the structure we expose externally. In fact, I think I should return meaningful JSON names just for this case.

errors.go
mitar: This is something else. We are breaking `Description`, but this field is here so that one can marshal and unmarshal the error object. Without this field it becomes tricky (I would have to unparse the multi-line description, not too hard, but slightly fragile).

Maybe then `ShortDescription`?

I would be able to break everything, I would go for:

* Code
* Name
* Description
* Hint
* StatusCode
* Debug

introspect.go
mitar: Sure.

mitar:
Please see my comments above and tell me your final decisions so that I know what to do.

access_error.go
aeneasr: In that case I would prefer:

```suggestion
	js, err := json.Marshal(rfcerr)
```

access_error.go
```diff
@@ -38,12 +38,16 @@ func (f *Fosite) writeJsonError(rw http.ResponseWriter, err error) {
...
-		rfcerr.Debug = ""
+		rfcerr = rfcerr.Sanitize()
 	}
 
-	js, err := json.Marshal(rfcerr)
+	js, err := json.MarshalIndent(rfcerr, "", "\t")
 	if err != nil {
-		http.Error(rw, fmt.Sprintf(`{"error": "%s"}`, err.Error()), http.StatusInternalServerError)
+		if f.SendDebugMessagesToClients {
+			http.Error(rw, fmt.Sprintf(`{\n\t"error": "server_error",\n\t"error_description": "%s"\n}`, err.Error()), http.StatusInternalServerError)
```
aeneasr, line 47: If `err.Error()` contains a quote this will result in invalid JSON because of missing escaping. You might want to use tidwall/gjson or tidwall/sjson to properly escape the string.

access_error.go
aeneasr:
```suggestion
			http.Error(rw, `{"error":"server_error"}`, http.StatusInternalServerError)
```

authorize_error.go
```diff
@@ -34,47 +34,49 @@ func (f *Fosite) WriteAuthorizeError(rw http.ResponseWriter, ar AuthorizeRequest
...
-		if !f.SendDebugMessagesToClients {
-			rfcerr.Debug = ""
-		}
+		rw.Header().Set("Content-Type", "application/json;charset=UTF-8")
 
 		js, err := json.MarshalIndent(rfcerr, "", "\t")
 		if err != nil {
-			http.Error(rw, err.Error(), http.StatusInternalServerError)
+			if f.SendDebugMessagesToClients {
+				http.Error(rw, fmt.Sprintf(`{\n\t"error": "server_error",\n\t"error_description": "%s"\n}`, err.Error()), http.StatusInternalServerError)
```
aeneasr, line 47: Same problem with potential JSON escaping as above!

```suggestion
				http.Error(rw, fmt.Sprintf(`{"error": "server_error","error_description": "%s"}`, err.Error()), http.StatusInternalServerError)
```

authorize_error.go
```diff
@@ -34,47 +34,49 @@ func (f *Fosite) WriteAuthorizeError(rw http.ResponseWriter, ar AuthorizeRequest
...
+			} else {
+				http.Error(rw, `{\n\t"error": "server_error"\n}`, http.StatusInternalServerError)
```
aeneasr, line 49:
```suggestion
				http.Error(rw, `{"error": "server_error"}`, http.StatusInternalServerError)
```

errors.go
aeneasr: Ok, SGTM

errors.go
aeneasr: I see, in that case the semantics make sense!

aeneasr (commented):
Thank you, almost there :) Still on vacation until September 23rd so response times might be a bit longer!

access_error.go
mitar: So this is tricky, so we get here ONLY if json encoding failed. And so I am not sure what is the point of trying another library at this point. There is something strange happening, so I attempt to do something light.

Maybe I could just do find/replace on `"` into `\"` and `\` into `\\` just to handle the most obvious cases and this is it?

mitar:
I updated the PR.

Enjoy vacations, no hurry with this one.

aeneasr (approved):
Nice! :)

nerocrux:
@mitar @aeneasr 
Thank you for your work!
It's not important but I wonder if it is possible to provide an option for removing `error_verbose` from the error response because the message in `error_verbose` and `error_description` feels duplicated for me.

mitar:
I would suggest we simply make `UnmarshalJSON` always return an error and remove `error_verbose` altogether. So `error_verbose` is there just to support `UnmarshalJSON`, but `UnmarshalJSON` is not really needed at all. I implemented it just for completeness.

mitar:
Also, how people feel about removing `error_hint` and `error_debug`? Or should we just maybe mark them in a comment as 'deprecated pending future removal" and then remove them in half a year or so?

nerocrux:
@mitar 
Thank you for your reply.

> I would suggest we simply make UnmarshalJSON always return an error and remove error_verbose altogether.

I agree with you.

> how people feel about removing error_hint and error_debug?

I'm a user of fosite (not hydra) so my opinion's maybe not complete but I'm totally OK with removing error_hint and error_debug (merging them into error_description).

lazeratops:
Came here to ask about the same thing - `error_verbose` seems redundant. If `error_hint `and `error_debug` are already merged into `error_description` I am fine removing them (as a caveat also just a fosite user, not hydra, not sure exactly how it would affect hydra users). To clarify for sanity purposes, to ensure I am seeing intended behaviour: the only one of these fields that is _not_ intended to be exposed back to the client by default is `error_debug`, correct?

aeneasr:
Those fields will be eventually deprecated. In the meanwhile, simply ignore them!

lazeratops:
We can definitely ask the client to ignore them; the main reason I was asking is to confirm that our auth service is not leaking any info that it shouldn't and is up to spec with the RFC. When we run `WriteAuthorizeError()` via the fosite provider, the response fosite produces to send back to the client does include those `error_verbose` fields and error hints in its current state. I wanted to confirm that we are not meant to be sanitizing any of this away (except for the debug info, which is already sanitized by fosite)

mitar:
Yes, those are fine. There is no extra information provided with these new fields, just merging them into `error_description`.

nerocrux:
Thanks. I'm OK with this. But please allow me to confirm: 

1. There is no way to remove error_verbose from error response (e.g. token endpoint) and the decision is we don't remove it for now.
2. Do you have any plan on the deprecation of these fields?

Thank you.

aeneasr:
Yes, these will be deprecated in the upcoming release!

aeneasr:
For anyone following this thread - I had to introduce some changes to this original PR. You can find the changeset here: https://github.com/ory/fosite/pull/528/commits/e463674f5955671bce664f3e29b0fc83eb659c04

Issue #267, opened 2018-05-16
State parameter length issue when working with spring security oauth2


pragkent:
The default state generator of spring security oauth2 library generates state with length of 6 which is less than the `MinParameterEntropy`(which is 8).  [check code here](https://github.com/spring-projects/spring-security-oauth/blob/master/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/common/util/RandomValueStringGenerator.java) 

I wonder that if it's possible to adjust MinParameterEntropy? or make it configurable?

arekkas:
Making it configurable would be an option

pragkent:
Any thought of how to implement it?
My first thought was making MinParameterEntropy a variable instead of a const.

arekkas:
Yeah, just add a `Fosite.GetMinParameterEntropy()` function which returns either `MinParameterEntropy`, or - if set - the value of e.g. `Fosite.MinParameterEntropy` which can be set when initializing the struct.

mitar:
I made https://github.com/ory/fosite/pull/461 to address this.

PR #461, opened 2020-08-13
feat: Allows configuring MinParameterEntropy

mitar:
## Related issue

Fixes https://github.com/ory/fosite/issues/267.

## Proposed changes

Added `MinParameterEntropy` to `compose.Config` and `fosite.Fosite` (and few other structs) which then controls the minimal length of the parameters which should have entropy.

## Checklist

      the changes.

## Further comments

There are quite a few extra changes to tests to make it a bit easier to control this config variable. I somehow disliked the idea of the global state between various test cases and that previous test case's `MinParameterEntropy` value would influence later ones. So I made functions to create new instances of objects as needed.


errors.go
```diff
@@ -111,7 +111,7 @@ var (
 	}
 	ErrInsufficientEntropy = &RFC6749Error{
 		Name:        errInsufficientEntropyName,
-		Description: fmt.Sprintf("The request used a security parameter (e.g., anti-replay, anti-csrf) with insufficient entropy (minimum of %d characters)", MinParameterEntropy),
+		Description: "The request used a security parameter (e.g., anti-replay, anti-csrf) with insufficient entropy",
```
mitar, line 114: It is not necessary to have the minimal length here because we pass a similar message (with minimal length) as a hint as well.

PR #479, opened 2020-09-16
feat: Errors now wrap underlying errors

### LINKED ISSUE #458: Better propagation of underlying errors
In `client_authentication.go` there is code like:

```
			client, err = f.Store.GetClient(ctx, clientID)
			if err != nil {
				return nil, errors.WithStack(ErrInvalidClient.WithDebug(err.Error()))
			}
```

This means that all the information about the `err` (stack trace, other wrapped errors) is lost at that point. Could `RFC6749Error` implement `WithError` and `Unwrap` to allow chaining errors? So then the above would be:

```
return nil, errors.WithStack(ErrInvalidClient.WithError(err).WithDebug(err.Error()))
```

External facing nothing would change, but if you inspect returned error, you would have access to the underlying wrapped error using (now) standard error wrapping.

mitar:
## Related issue

Fixes #458.

## Proposed changes

All errors wrap now underlying errors, when there is one available. This is not directly exposed through messages, but if you obtain such error in your code you have more information to inspect, log, and act upon.

I also changed around the codebase to remove hard-coded assumptions about the depth of a cause when inspecting errors and use `errors.Is` and `errors.As` instead.

## Checklist

      the changes.

### DIFF
--- a/revoke_handler.go
+++ b/revoke_handler.go
@@ -51,7 +51,7 @@ func (f *Fosite) NewRevocationRequest(ctx context.Context, r *http.Request) erro
 	if r.Method != "POST" {
 		return errors.WithStack(ErrInvalidRequest.WithHintf("HTTP method is \"%s\", expected \"POST\".", r.Method))
 	} else if err := r.ParseMultipartForm(1 << 20); err != nil && err != http.ErrNotMultipart {
-		return errors.WithStack(ErrInvalidRequest.WithHint("Unable to parse HTTP body, make sure to send a properly formatted form request body.").WithDebug(err.Error()))
+		return errors.WithStack(ErrInvalidRequest.WithHint("Unable to parse HTTP body, make sure to send a properly formatted form request body.").WithCause(err).WithDebug(err.Error()))
 	} else if len(r.PostForm) == 0 {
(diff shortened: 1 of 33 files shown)


revoke_handler.go
```diff
@@ -125,7 +124,7 @@ func (f *Fosite) WriteRevocationResponse(rw http.ResponseWriter, err error) {
 
 		rw.WriteHeader(ErrInvalidClient.Code)
 		rw.Write(js)
-	default:
+	} else {
 		// 200 OK
 		rw.WriteHeader(http.StatusOK)
```
titouanfreville, line 129: Shouldn't this return an error code ? If I'm here it means that there is an error so we should return it to the client shouldn't we ?

mitar: So this is not something I changed, but it is like this already in existing code. But you are right, I was also wondering about this. I can fix this as part of this PR as well. What should be returned here? Internal server error?

titouanfreville: 503 from the RFC. But I think they are on it. One of my colleague signal it already.

zepatrik: We received exactly this as a sec vulnerability disclosure yesterday. Working on that already, stay tuned.

aeneasr (approved):

Issue #464, opened 2020-08-15
Double query parameter escaping around code


mitar:
**Describe the bug**

I think this is a similar problem than what was already reported in https://github.com/ory/fosite/pull/434, but elsewhere in the code. I believe all calls to `url.QueryUnescape` should be removed from the code. Go already escapes all query/form parameters automatically and unescaping again can just break things.

Example:

```
func GetRedirectURIFromRequestValues(values url.Values) (string, error) {
	// rfc6749 3.1.   Authorization Endpoint
	// The endpoint URI MAY include an "application/x-www-form-urlencoded" formatted (per Appendix B) query component
	redirectURI, err := url.QueryUnescape(values.Get("redirect_uri"))
	if err != nil {
		return "", errors.WithStack(ErrInvalidRequest.WithHint(`The "redirect_uri" parameter is malformed or missing.`).WithDebug(err.Error()))
	}
	return redirectURI, nil
}
```

mitar:
Done in #480.

PR #480, opened 2020-09-19
fix: redirct_url with query escape character outside of query is failing

ajanthan:
## Related issue
https://github.com/ory/hydra/issues/2055

## Proposed changes
Updated the GetRedirectURIFromRequestValues() method to unescape the query parameters only.

## Checklist

      the changes.


mitar:
Thank you for working on this. I have reported this issue in the past in https://github.com/ory/fosite/issues/464. So this is not the only place where this code logic exist, so it would be great if you could also go around the full codebase (just search for `QueryUnescape`) and fix it everywhere? And add also related tests.

authorize_helper.go
```diff
@@ -41,11 +41,17 @@ import (
 func GetRedirectURIFromRequestValues(values url.Values) (string, error) {
 	// rfc6749 3.1.   Authorization Endpoint
 	// The endpoint URI MAY include an "application/x-www-form-urlencoded" formatted (per Appendix B) query component
-	redirectURI, err := url.QueryUnescape(values.Get("redirect_uri"))
+	redirectURI, err := url.Parse(values.Get("redirect_uri"))
```
mitar, line 44: I think this is correct. We just want to validate the `redirect_uri` value.

authorize_helper.go
```diff
@@ -41,11 +41,17 @@ import (
...
 	if err != nil {
 		return "", errors.WithStack(ErrInvalidRequest.WithHint(`The "redirect_uri" parameter is malformed or missing.`).WithCause(err).WithDebug(err.Error()))
 	}
-	return redirectURI, nil
+	rawQuery, err := url.QueryUnescape(redirectURI.RawQuery)
```
mitar, line 48: But I do not think this is necessary. `values` should already include correctly parsed parameter from the form payload, so no further unescaping of the `redirect_uri` should be necessary. We should just pass it on as-is. We should just validate that the `redirect_uri` is valid URL.

Do you have a test case where `QueryUnescape` is necessary?

mitar (changes requested):

authorize_helper.go
aeneasr: I'm also wondering if `values.Get` isn't already unescaped?

ajanthan: Yes. That is the root of the issue.

authorize_helper.go
ajanthan: Without escaping the query part, one [existing test](https://github.com/ory/fosite/blob/9cecb109a338da8a5c1bb16a10bb00344dba89ba/authorize_helper_test.go#L56) case fails. 
According to this test case, During the client registration, if redirct_uri has an url-encoded query parameter, it is expected to be escaped. I am not sure if it is expected according to the spec.

ajanthan:
> Thank you for working on this. I have reported this issue in the past in #464. So this is not the only place where this code logic exist, so it would be great if you could also go around the full codebase (just search for `QueryUnescape`) and fix it everywhere? And add also related tests.

In other places[1][2][3][4], `QueryUnescape` is used to unescape the special URL characters from the Basic authentication username and password. These special characters won't be escaped during the golang request parsing. It makes sense to keep the logic there. 

WDYT?

[1] https://github.com/ory/fosite/blob/b53f8f58f0b9889d044cf9a8e2604316f0559ff6/introspection_request_handler.go#L140
[2] https://github.com/ory/fosite/blob/b53f8f58f0b9889d044cf9a8e2604316f0559ff6/introspection_request_handler.go#L145
[3] https://github.com/ory/fosite/blob/b53f8f58f0b9889d044cf9a8e2604316f0559ff6/client_authentication.go#L277
[4] https://github.com/ory/fosite/blob/b53f8f58f0b9889d044cf9a8e2604316f0559ff6/client_authentication.go#L279

authorize_helper.go
mitar: But that unescaping should be done by accessing the query string using `Query()`, no? So form payload unparses it once as a whole `redirect_uri` parameter. And then `URL` as result of `url.Parse` has `Query()` method which unescapes when needed. There is no point in doing `redirectURI.RawQuery = url.QueryUnescape(redirectURI.RawQuery)`.

I think the confusion here is from this sentence in the spec:

> The endpoint URI MAY include an "application/x-www-form-urlencoded" formatted (per Appendix B) query component ([RFC3986] Section 3.4), which MUST be retained when adding additional query parameters.

I think this does not mean that you do `urlescape("param1=value1&param2=value2")` but that it has to be `"param1=" + urlescape(value1) + "&param2=" + urlescape(value2)`. So the word there is `formatted`, not `escaped`. But in fact, both of those forms are exactly equivalent and should be processed the same by `url.Query()` call. See [this seciton](https://tools.ietf.org/html/rfc3986#section-3.4) (referenced from [RFC section](https://tools.ietf.org/html/rfc6749#section-3.1.2) referenced from the test you linked above) which says:

> However, as query components are often used to carry identifying information in the form of "key=value" pairs and one frequently used value is a reference to another URI, it is sometimes better for usability to avoid percent-encoding those characters.

So urlescaping also `+` and `=` characters in the query string is fine. But it is not necessary for usability. But both of those are demeed to be `"application/x-www-form-urlencoded" formatted`. In any case, we do not have to double unescape things. We just have to make sure that we append additional query values to existing. Which we should be doing by calling `url.Query()` (which would unescape things), add additional ones, and then encode them back. Which I think we are doing, no?

So, no need for double decoding here.

mitar:
> These special characters won't be escaped during the golang request parsing. It makes sense to keep the logic there.

You are right. Those are the only ones left. And yes, those should be kept in as per RFC itself. So cool, this is then the only place left to fix this.

authorize_helper.go
ajanthan: Yes. I completely agree with you. So, we can get rid of `GetRedirectURIFromRequestValues` method and the associated test case.

mitar: No, we can keep it, just to verify the redirect URI.

You should probably update the test case accordingly.

authorize_helper.go
```diff
@@ -30,24 +30,6 @@ import (
...
 
-// GetRedirectURIFromRequestValues extracts the redirect_uri from values but does not do any sort of validation.
-//
-// Considered specifications
-// * https://tools.ietf.org/html/rfc6749#section-3.1.2
-//   The endpoint URI MAY include an
-//   "application/x-www-form-urlencoded" formatted (per Appendix B) query
-//   component ([RFC3986] Section 3.4), which MUST be retained when adding
-//   additional query parameters.
-func GetRedirectURIFromRequestValues(values url.Values) (string, error) {
```
mitar, line 41: No, as I mentioned. You should still validate that the url is correct. So just parse the URL, check error return value, and this is it.

authorize_helper.go
```diff
@@ -41,11 +41,15 @@ import (
 func GetRedirectURIFromRequestValues(values url.Values) (string, error) {
 	// rfc6749 3.1.   Authorization Endpoint
 	// The endpoint URI MAY include an "application/x-www-form-urlencoded" formatted (per Appendix B) query component
-	redirectURI, err := url.QueryUnescape(values.Get("redirect_uri"))
+	rawRedirectURI := values.Get("redirect_uri")
+	redirectURI, err := url.Parse(rawRedirectURI)
 	if err != nil {
 		return "", errors.WithStack(ErrInvalidRequest.WithHint(`The "redirect_uri" parameter is malformed or missing.`).WithCause(err).WithDebug(err.Error()))
+	} else if rawRedirectURI != "" && (redirectURI.Scheme == "" || redirectURI.Host == "") {
```
mitar, line 48: Any particular reason to add this additional check at this point?

authorize_helper.go
mitar: Sorry, I realized that this function does not have to validate anything, it is written in its description and redirect URI validation is done later in `MatchRedirectURIWithClientRedirectURIs`.

I do agree that it could probably just be removed. But let's wait for @aeneasr to confirm.

authorize_helper.go
ajanthan: 1. According to the [test case](https://github.com/ory/fosite/blob/b53f8f58f0b9889d044cf9a8e2604316f0559ff6/authorize_helper_test.go#L62) empty `redirect_uri` is allowed

2. Golang url.parse doesn't catch [bad URL scheme or host](https://stackoverflow.com/a/55551215) thus the method fails to catch urls like `https//google.com/foo=bar`

aeneasr: That's why we have `url.ParseRequestURI("https//google.com/foo=bar")` :)

https://play.golang.org/p/eblL4ikZTSi

aeneasr: Please keep in mind that this errors if the string is empty, so you need to handle that as a special case!

mitar: But this function should not be doing any validation based on its docstring!

So see other comments. Both @ajanthan and me think now that this function should just be removed and replaced with simply calling `values.Get("redirect_uri")`. The validation happens anyway in `MatchRedirectURIWithClientRedirectURIs`.

aeneasr: Right, overlooked that - makes sense!

aeneasr:
@ajanthan are you still up for the changes? :) If you need any help, let us know!

ajanthan:
@aeneasr updated the pr

mitar (approved):
Looks good to me.

aeneasr (approved):
Awesome 🎉 

PR #481, opened 2020-09-20
fix: Full JSON escaping

mitar:
## Related issue

This is a followup to https://github.com/ory/fosite/pull/460.

## Proposed changes

This makes the escaping complete and not just partial. I think this is good enough because this code path should really never be called. But it does, we should do proper escaping which this now does.

## Checklist

      the changes.

### DIFF
--- a/access_error.go
+++ b/access_error.go
@@ -25,7 +25,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"strings"
 )
 
 func (f *Fosite) WriteAccessError(rw http.ResponseWriter, _ AccessRequester, err error) {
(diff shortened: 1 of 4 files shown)


access_error.go
```diff
@@ -46,8 +46,17 @@ func (f *Fosite) writeJsonError(rw http.ResponseWriter, err error) {
...
+			// See: https://tools.ietf.org/html/std90#section-7
 			errorMessage := err.Error()
+			// Escape reverse solidus.
 			errorMessage = strings.ReplaceAll(errorMessage, `\`, `\\`)
+			// Escape control characters.
+			for r := rune(0); r < ' '; r++ {
+				errorMessage = strings.ReplaceAll(errorMessage, string(r), fmt.Sprintf(`\u%04d`, r))
+			}
+			// Escape quotation mark.
 			errorMessage = strings.ReplaceAll(errorMessage, `"`, `\"`)
```
aeneasr, line 60: Maybe make this a small helper function to avoid code duplication and add a small test? :)

mitar: Done.

aeneasr:
Nice :)

PR #486, opened 2020-09-26
fix: Returning Valid token_type in Introspection Response

ajanthan:
## Related issue
https://github.com/ory/hydra/issues/1762

## Proposed changes

- Added fix to return `bearer` or `mac` in `token_type` of introspection response based on the type of the token being introspected
- If the token type is `access_token` it returns `bearer` always as `mac` is not supported now. 

- Added extra attribute `token_use` to return the token type `access_token` or `refresh_token`. 

## Checklist

      the changes.


ajanthan:
cc @aeneasr

introspection_request_handler.go
```diff
@@ -167,13 +167,15 @@ func (f *Fosite) NewIntrospectionRequest(ctx context.Context, r *http.Request, s
...
 		TokenType:       tt,
+		AccessTokenType: att,
 	}, nil
 }
 
 type IntrospectionResponse struct {
 	Active          bool            `json:"active"`
 	AccessRequester AccessRequester `json:"extra"`
-	TokenType       TokenType       `json:"token_type,omitempty"`
+	TokenType       TokenType       `json:"token_use,omitempty"`
```
aeneasr, line 177: I think we could change this for readability purposes in a breaking fashion so that we have:

```suggestion
	TokenUse       TokenUse       `json:"token_use,omitempty"`
	AccessTokenType Type `json:"token_type,omitempty"`
```

aeneasr (commented):
Thank you for this PR! I looked a bit into Mac-based HTTP based authentication and could not find anything related to OAuth2 except for [this draft](https://tools.ietf.org/html/draft-ietf-oauth-v2-http-mac-05) which expired 2014 while HTTP Bearer token is an [accepted standard](https://tools.ietf.org/html/rfc6750).

I think it's therefore safe to assume that fosite always returns a bearer token, which is great, because then we do not have to break backwards compatibility any more! So all we have to do is update the introspection writer!

I just realized that the request parameters have a token_type_hint parameter which can be access_token or refresh_token (see also RFC7009) but then return token_type in a completely different context (related to the HTTP auth mechanism). Classic OAuth2 ...

handler/oauth2/introspector.go
```diff
@@ -36,32 +36,32 @@ type CoreValidator struct {
 	DisableRefreshTokenValidation bool
 }
 
-func (c *CoreValidator) IntrospectToken(ctx context.Context, token string, tokenType fosite.TokenType, accessRequest fosite.AccessRequester, scopes []string) (fosite.TokenType, error) {
+func (c *CoreValidator) IntrospectToken(ctx context.Context, token string, tokenUse fosite.TokenUse, accessRequest fosite.AccessRequester, scopes []string) (fosite.TokenUse, fosite.AccessTokenType, error) {
```
aeneasr, line 39: Given that `fosite.AccessTokenType` is `bearer` always, we don't have to change backwards compatiilbity here with an additional parameter! :)

handler/oauth2/introspector_jwt.go
```diff
@@ -39,10 +39,10 @@ type StatelessJWTValidator struct {
 	ScopeStrategy fosite.ScopeStrategy
 }
 
-func (v *StatelessJWTValidator) IntrospectToken(ctx context.Context, token string, tokenType fosite.TokenType, accessRequest fosite.AccessRequester, scopes []string) (fosite.TokenType, error) {
+func (v *StatelessJWTValidator) IntrospectToken(ctx context.Context, token string, tokenUse fosite.TokenUse, accessRequest fosite.AccessRequester, scopes []string) (fosite.TokenUse, fosite.AccessTokenType, error) {
```
aeneasr, line 42: Given that `fosite.AccessTokenType` is `bearer` always, we don't have to change backwards compatiilbity here with an additional parameter! :)

introspect.go
```diff
@@ -30,7 +30,7 @@ import (
 )
 
 type TokenIntrospector interface {
-	IntrospectToken(ctx context.Context, token string, tokenType TokenType, accessRequest AccessRequester, scopes []string) (TokenType, error)
+	IntrospectToken(ctx context.Context, token string, tokenUse TokenUse, accessRequest AccessRequester, scopes []string) (TokenUse, AccessTokenType, error)
```
aeneasr, line 33: As said above:

```suggestion
	IntrospectToken(ctx context.Context, token string, tokenUse TokenUse, accessRequest AccessRequester, scopes []string) (TokenUse, error)
```

oauth2.go
```diff
@@ -143,7 +150,7 @@ type OAuth2Provider interface {
 
 	// IntrospectToken returns token metadata, if the token is valid. Tokens generated by the authorization endpoint,
 	// such as the authorization code, can not be introspected.
-	IntrospectToken(ctx context.Context, token string, tokenType TokenType, session Session, scope ...string) (TokenType, AccessRequester, error)
+	IntrospectToken(ctx context.Context, token string, tokenUse TokenUse, session Session, scope ...string) (TokenUse, AccessTokenType, AccessRequester, error)
```
aeneasr, line 153: As discussed above:

```suggestion
	IntrospectToken(ctx context.Context, token string, tokenUse TokenUse, session Session, scope ...string) (TokenUse, AccessRequester, error)
```

aeneasr (commented):
Awesome, thank you for your contribution! This looks pretty good but I have some ideas how to improve it further :)

handler/oauth2/introspector.go
ajanthan: Agree!

handler/oauth2/introspector_jwt.go
ajanthan: Yes, agree

introspect.go
ajanthan: Changed according to the suggestion.

oauth2.go
```diff
@@ -28,13 +28,18 @@ import (
 	"time"
 )
 
+type TokenUse = TokenType
+
+//type TokenUse string
```
aeneasr, line 33:
```suggestion
```

aeneasr (approved):
Awesome 🎉 

PR #489, opened 2020-09-30
feat: Allow configuring redirect secure checker everywhere

mitar:
## Proposed changes

I have been reading this part of the code and found out some things I could improve.

## Checklist

      the changes.

## Further comments

This is changing/cleaning some security logic used by fosite. I do not think any of these changes are changing anything really exploitable, but I still think it will harden and cleanup things a bit. Or at least make it more consistent.


authorize_helper.go
```diff
@@ -131,10 +131,6 @@ func isLoopbackURI(requested *url.URL, registeredURI string) bool {
 		return false
 	}
 
-	if registered.Scheme != "http" || !isLoopbackAddress(registered.Host) {
```
mitar, line 134: This is redundant. So I removed it.

authorize_helper.go
```diff
@@ -172,10 +168,10 @@ func IsValidRedirectURI(redirectURI *url.URL) bool {
 }
 
 func IsRedirectURISecure(redirectURI *url.URL) bool {
-	return !(redirectURI.Scheme == "http" && !IsLocalhost(redirectURI))
+	return redirectURI.Scheme == "https" || (redirectURI.Scheme == "http" && IsLocalhost(redirectURI))
```
mitar, line 171: This is what the spec says is secure, not sure why would any other scheme be secure? Other schemes are not secure. So I moved code from `OpenIDConnectRequestValidator` here, and made `OpenIDConnectRequestValidator` use `IsRedirectURISecure`.

authorize_helper.go
```diff
@@ -172,10 +168,10 @@ func IsValidRedirectURI(redirectURI *url.URL) bool {
...
 }
 
 func IsLocalhost(redirectURI *url.URL) bool {
 	hn := redirectURI.Hostname()
-	return strings.HasSuffix(hn, ".localhost") || hn == "127.0.0.1" || hn == "localhost"
+	return strings.HasSuffix(hn, ".localhost") || hn == "127.0.0.1" || hn == "::1" || hn == "localhost"
```
mitar, line 176: IPv6 localhost is still localhost.

authorize_helper_test.go
```diff
@@ -236,8 +236,11 @@ func TestIsRedirectURISecure(t *testing.T) {
 		{u: "http://localhost", err: false},
 		{u: "http://test.localhost", err: false},
 		{u: "http://127.0.0.1/", err: false},
+		{u: "http://[::1]/", err: false},
+		{u: "http://127.0.0.1:8080/", err: false},
+		{u: "http://[::1]:8080/", err: false},
 		{u: "http://testlocalhost", err: true},
-		{u: "wta://auth", err: false},
+		{u: "wta://auth", err: true},
```
mitar, line 243: So this changes. I do not think a custom scheme is secure. On iPhone, custom app schemes are not unique to the app. I have not found a platform where this would be secure.

handler/openid/validator.go
```diff
@@ -38,6 +39,7 @@ import (
 type OpenIDConnectRequestValidator struct {
 	AllowedPrompt []string
 	Strategy      jwt.JWTStrategy
+	IsRedirectURISecure func(*url.URL) bool
```
mitar, line 42: Adding this for completeness. We are allowing configuring `IsRedirectURISecure` in some places, but not in all. I added it here so that we can configure it here as well.

handler/openid/validator.go
```diff
@@ -75,7 +90,7 @@ func (v *OpenIDConnectRequestValidator) ValidatePrompt(ctx context.Context, req
 		//  be processed as if no previous request had been approved.
 
 		if stringslice.Has(prompt, "none") {
-			if !(req.GetRedirectURI().Scheme == "https" || (fosite.IsLocalhost(req.GetRedirectURI()) && req.GetRedirectURI().Scheme == "http")) {
+			if !v.secureChecker()(req.GetRedirectURI()) {
```
mitar, line 93: So instead of duplicating code, I moved all this code to `IsRedirectURISecure`.

authorize_helper.go
aeneasr: While that's currently true, removing explicit checks for implicit ones opens up a big opportunity for regression issues. Given that this is security sensitive, and that you want to have this removed, please either write a test covering this exact case or point to the test case that already covers this logic. Thanks! :)

aeneasr (changes requested):

authorize_helper.go
mitar: But this is logically redundant. It is literally the opposite than the next case. It is De Morgan law.

Compare:

```
registered.Scheme != "http" || !isLoopbackAddress(registered.Host)
```

and:

```
requested.Scheme == "http" && isLoopbackAddress(requested.Host)
```

authorize_helper.go
mitar: > Secure in this context does not mean "HTTPS" or "encrypted".

Yes, I mean really secure. When I wrote above, "This is what the spec says is secure, not sure why would any other scheme be secure" I meant [RFC 8252](https://tools.ietf.org/html/rfc8252). And I think you are referring to the same one. We just disagree in how we read it. :-) But I even disagree with the spec and I think some things they do not forbid explicitly are insecure.

In any case, I can change the default here because now it is configurable. But I do think it is better to be secure out of the box and if somebody wants to relax security, they can configure a less secure version of "IsRedirectURISecure". I think this is what fosite is all about. I think fosite should see as secure only https (and claimed https redirects) and loopback redirects.

> custom-scheme URLs for mobile apps are loopback, they are considered secure

No, they are not. See [here](https://developer.apple.com/documentation/xcode/allowing_apps_and_websites_to_link_to_your_content/defining_a_custom_url_scheme_for_your_app):

> URL schemes offer a potential attack vector into your app, so make sure to validate all URL parameters and discard any malformed URLs. In addition, limit the available actions to those that don’t risk the user’s data. For example, don’t allow other apps to directly delete content or access sensitive information about the user. When testing your URL-handling code, make sure your test cases include improperly formatted URLs.

If a company who offers custom schemes writes this, I think we should believe them, no?

It is true that the spec [in appendix B.1](https://tools.ietf.org/html/rfc8252#appendix-B.1) says:

> To receive the authorization response, both private-use URI scheme (referred to as "custom URL scheme") redirects and claimed "https" scheme URIs (known as "Universal Links") are viable choices.

But also:

> Claimed "https" scheme URIs are the preferred redirect choice on iOS 9 and above due to the ownership proof that is provided by the operating system.

In [section 8.1](https://tools.ietf.org/html/rfc8252#section-8.1) the spec acknowledges this:

> A limitation of using private-use URI schemes for redirect URIs is that multiple apps can typically register the same scheme, which makes it indeterminate as to which app will receive the authorization code.

And they offer PKCE as a solution to interception. And where I disagree with the spec. PKCE only prevents a malicious app to intercept OIDC request originating from true app. But it **does not** protect against malicious app **both initiating sign-in flow and then receiving the redirect**. In fact, true app might not even be installed on the phone, just a malicious app might be. Which means that it could phish an user to sign-in with my fosite-based authentication service, obtaining an access token they can then misuse.

So given that native apps today have access to "claimed https redirects" on both iOS and Android platforms and that this is secure, why allow anything less? Why open a potential window of attack, given that a more secure alternative exists?

authorize_helper_test.go
mitar: > If you error here, all mobile app auth will fail in prod-systems.

Not true, they just have to use "claimed https redirects".

Or are you saying that current live production systems will start failing with this? Sure. We can have a deprecation period here. But I do think we should deprecate use of anything except for "claimed https redirects" (and loopback IPs).

mitar: We can also offer both secure and less secure version of `IsRedirectURISecure`, where I would go with secure version by default, but changing to less secure would then be one line configuration change people have to do, to keep old behavior.

authorize_helper.go
mitar: Oh, you just want a test case where using `http` scheme returns false and using a `http` scheme without loopback address returns false? I [think those are already covered](https://github.com/ory/fosite/blob/master/authorize_helper_test.go#L229-L246).

aeneasr: Exactly!

That test tests `IsRedirectURISecure` which does not use `isLoopbackURI` so I think this is neither covered explicitly nor implicitly.

authorize_helper.go
aeneasr: > So given that native apps today have access to "claimed https redirects" on both iOS and Android platforms and that this is secure, why allow anything less? Why open a potential window of attack, given that a more secure alternative exists?

Because not doing that will cause breaking changes for existing apps, which in turn means potentially hundreds of people complaining that ORY Hydra no longer works. It is very, very common to use custom redirect URIs in mobile environments.

Your interpretation might be correct or better, but the real world forces our hand here. Because the specification is simply the way it is, it is implemented as such very commonly in the real world. So let's not make our user's life harder than it has to be, OAuth2 sucks enough already.

authorize_helper_test.go
aeneasr: Please revert this, as discussed in the comment above. This can not be changed, forcing our user's users to move to "claimed https redirects" because we changed the default is not acceptable.

aeneasr (changes requested):

mitar:
OK. Are you open to having `IsRedirectURISecureStrict` or something alongside `IsRedirectURISecure` so that one can pick one or the other? I can keep then `IsRedirectURISecure` as a default and explain in `IsRedirectURISecureStrict`'s docstring the reasoning behind it?

mitar:
Addressed review comments.

aeneasr (approved):
Nice, this looks great!

PR #506, opened 2020-10-03
fix: return allowed redirect url with preference

aeneasr:
## Checklist

- [ ] I have read the [security policy](../security/policy)
      the changes.

### DIFF
--- a/authorize_helper.go
+++ b/authorize_helper.go
@@ -84,10 +84,10 @@ func MatchRedirectURIWithClientRedirectURIs(rawurl string, client Client) (*url.
 			// If no redirect_uri was given and the client has exactly one valid redirect_uri registered, use that instead
 			return redirectURIFromClient, nil
 		}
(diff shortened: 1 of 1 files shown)


authorize_helper.go
```diff
@@ -84,10 +84,10 @@ func MatchRedirectURIWithClientRedirectURIs(rawurl string, client Client) (*url.
 			// If no redirect_uri was given and the client has exactly one valid redirect_uri registered, use that instead
 			return redirectURIFromClient, nil
 		}
-	} else if rawurl != "" && isMatchingRedirectURI(rawurl, client.GetRedirectURIs()) {
+	} else if redirectTo, ok := isMatchingRedirectURI(rawurl, client.GetRedirectURIs()); rawurl != "" && ok {
```
mitar, line 87: I think `rawurl != ""` should be first.

aeneasr: It's unfortunately not possible to check before define so unless this is rewritten without `else if` it's the only possibility :/

Issue #470, opened 2020-09-05
Support for response_mode parameter and form post response mode


mitar:
`response_mode` is defined here: https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html

Why I find it useful is that you can then use [Form Post Response mode](https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html) which I think mitigates many issues with query and fragment response modes.

So I would suggest `response_mode` is supported, together with form post response mode.

pjcdawkins:
We use response_mode=postMessage with prompt=none for an iframe "silent refresh" flow

aeneasr:
Contribs welcomed!

mitar:
Is `postMessage` standardized somewhere?

aeneasr:
So I meant contribs for `form_post`!

mitar:
FYI, for `response_mode=postMessage ` I think you are referring to [this draft](https://tools.ietf.org/html/draft-sakimura-oauth-wmrm-00)? Sadly, it has not been standardized yet. But [auth0](https://auth0.com/docs/protocols/protocol-oauth2#authorization-endpoint) does support it as one of response modes. Maybe fosite should, too.

aeneasr:
I don't think that we should support `postMessage`. It looks very legacy and we've never had anyone ask about this before or if they did, the flow could be solved using a much simpler method.

mitar:
Hm, to me it looks like one which is required to go around issues with 3rd party cookies, no? I am not sure what is the difference between `response_mode=postMessage` and [assisted token flow](https://github.com/ory/fosite/issues/408). The latter looks like something modern and useful today.

aeneasr:
Hard to say honestly, I would need to spend some good time reading all the different specs. I think for now - and unless someone asks - we'll just put it on the backburner. Since the OP requested form_post which is currently being worked on I'd probably also just close this unless @pjcdawkins chimes in as to why we would need `postMessage`

aeneasr:
This is now implemented!

Issue #409, opened 2020-03-30
Response type "none" seems to not be supported


mitar:
**Is your feature request related to a problem? Please describe.**

It seems the edge case response type "none" is not implemented/supported? I checked that by looking at the code. Nor I have found any test trying to test for this case.

**Describe the solution you'd like**

That support for "none" response type is implemented.

**Additional context**

Spec: https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#none

aeneasr:
Do you have an actual use case for this?

mitar:
One is described in the spec:

> One scenario is where a user wishes to purchase an application from a market, and desires to authorize application installation and grant the application access to Protected Resources in a single step. However, since the user is not presently interacting with the (not yet active) application, it is not appropriate to return access credentials simultaneously in the authorization step.

I am in fact working on a system which is similar, just that it provides access to datasets, but which might not yet have been uploaded.

aeneasr:
Ok, contributions welcome!

aeneasr:
I am closing this issue as it has not received any engagement from the community or maintainers in a long time. That does not imply that the issue has no merit. If you feel strongly about this issue

- open a PR referencing and resolving the issue;
- leave a comment on it and discuss ideas how you could contribute towards resolving it;
- open a new issue with updated details and a plan on resolving the issue.

We are cleaning up issues every now and then, primarily to keep the 4000+ issues in our backlog in check and to [prevent](https://www.jeffgeerling.com/blog/2020/saying-no-burnout-open-source-maintainer) [maintainer](https://www.jeffgeerling.com/blog/2016/why-i-close-prs-oss-project-maintainer-notes) [burnout](https://docs.brew.sh/Maintainers-Avoiding-Burnout). Burnout in open source maintainership is a [widespread](https://opensource.guide/best-practices/#its-okay-to-hit-pause) and serious issue. It can lead to severe personal and health issues as well as [enabling catastrophic](https://haacked.com/archive/2019/05/28/maintainer-burnout/) [attack vectors](https://www.gradiant.org/en/blog/open-source-maintainer-burnout-as-an-attack-surface/).

Thank you to anyone who participated in the issue! 🙏✌️

PR #509, opened 2020-10-09
feat: Adding form_post support

ajanthan:
## Related issue
https://github.com/ory/fosite/issues/470

## Proposed changes

- Introduces Form attribute to AuthorizeResponse

-  Populates Form attributes in the relevant grant type handlers if the `response_mode`=`form_post` presents in request

- Write the response as defined in the spec if the Form attributes are not empty

## Checklist

      the changes.

## Further comments
I haven't added support for `response_mode`= `query` or `fragment` or any other


ajanthan:
cc @mitar @aeneasr

authorize_helper.go
```diff
@@ -175,3 +182,118 @@ func IsLocalhost(redirectURI *url.URL) bool {
 	hn := redirectURI.Hostname()
 	return strings.HasSuffix(hn, ".localhost") || hn == "127.0.0.1" || hn == "localhost"
 }
+
+func WriteAuthorizeFormPostResponse(redirectURL string, parameters url.Values, rw io.Writer) {
+	t := template.Must(template.New("form_post").Parse(`<html>
```
aeneasr, line 187: This template could be defined as a global variable which means that it would not need to be rendered for every request!

authorize_helper.go
```diff
@@ -175,3 +182,118 @@ func IsLocalhost(redirectURI *url.URL) bool {
...
+
+	_ = t.Execute(rw, struct {
+		RedirURL   string
+		Parameters url.Values
+	}{
+		RedirURL:   redirectURL,
+		Parameters: parameters,
+	})
+}
+func ParseFormPostResponse(redirectURL string, resp io.ReadCloser) (authorizationCode, stateFromServer, iDToken string, token goauth.Token, rFC6749Error RFC6749Error, err error) {
```
aeneasr, line 208: Is this only used for testing?

authorize_error.go
```diff
@@ -66,7 +66,11 @@ func (f *Fosite) WriteAuthorizeError(rw http.ResponseWriter, ar AuthorizeRequest
 	query.Add("state", ar.GetState())
 
 	var redirectURIString string
-	if !(len(ar.GetResponseTypes()) == 0 || ar.GetResponseTypes().ExactOne("code")) && !errors.Is(err, ErrUnsupportedResponseType) {
+	if ar.GetRequestForm().Get("response_mode") == "form_post" {
```
aeneasr, line 69: I think it would make sense to validate this response mode in the authorize handler! And then have `ar.ResponseMode() fosite.ResponseMode` which is an enum of `fosite.ResponseModePost` and `fosite.ResponseModeQuery` (default).

handler/oauth2/flow_authorize_code_auth.go
```diff
@@ -110,10 +110,15 @@ func (c *AuthorizeExplicitGrantHandler) IssueAuthorizeCode(ctx context.Context,
 	if err := c.CoreStorage.CreateAuthorizeCodeSession(ctx, signature, ar.Sanitize(c.GetSanitationWhiteList())); err != nil {
 		return errors.WithStack(fosite.ErrServerError.WithCause(err).WithDebug(err.Error()))
 	}
-
-	resp.AddQuery("code", code)
-	resp.AddQuery("state", ar.GetState())
-	resp.AddQuery("scope", strings.Join(ar.GetGrantedScopes(), " "))
+	if ar.GetRequestForm().Get("response_mode") == "form_post" {
```
aeneasr, line 113: It looks like the two if statements are copies of one another. Can we avoid duplication here?

handler/oauth2/flow_authorize_implicit.go
```diff
@@ -94,12 +94,20 @@ func (c *AuthorizeImplicitGrantTypeHandler) IssueImplicitAccessToken(ctx context
 	if err := c.AccessTokenStorage.CreateAccessTokenSession(ctx, signature, ar.Sanitize([]string{})); err != nil {
 		return errors.WithStack(fosite.ErrServerError.WithCause(err).WithDebug(err.Error()))
 	}
+	if ar.GetRequestForm().Get("response_mode") == "form_post" {
```
aeneasr, line 97: It looks like the two if statements are copies of one another. Can we avoid duplication here?

handler/openid/flow_hybrid.go
```diff
@@ -109,11 +109,14 @@ func (c *OpenIDConnectHybridHandler) HandleAuthorizeEndpointRequest(ctx context.
 		if err := c.AuthorizeExplicitGrantHandler.CoreStorage.CreateAuthorizeCodeSession(ctx, signature, ar.Sanitize(c.AuthorizeExplicitGrantHandler.GetSanitationWhiteList())); err != nil {
 			return errors.WithStack(fosite.ErrServerError.WithCause(err).WithDebug(err.Error()))
 		}
-
-		resp.AddFragment("code", code)
+		if ar.GetRequestForm().Get("response_mode") == "form_post" {
```
aeneasr, line 112: It looks like the two if statements are copies of one another. Can we avoid duplication here?

handler/openid/flow_hybrid.go
```diff
@@ -109,11 +109,14 @@ func (c *OpenIDConnectHybridHandler) HandleAuthorizeEndpointRequest(ctx context.
...
+			resp.AddForm("code", code)
+		} else {
+			resp.AddFragment("code", code)
+		}
 		ar.SetResponseTypeHandled("code")
 
-		hash, err := c.Enigma.Hash(ctx, []byte(resp.GetFragment().Get("code")))
+		hash, err := c.Enigma.Hash(ctx, []byte(resp.GetCode()))
```
aeneasr, line 119: I don't think this is correct, it could be that `GetCode()` is not properly set. Please use `code` here!

aeneasr (changes requested):
Thank you for your hard work on this and the tests and sorry for the late review but it's a bit of a complicated review :)

I think we can improve this PR a lot - it looks like there is a lot of duplication between `AddForm` and `AddFragment`. I think we should have a general `ResponseMode()` method instead of checking for `ar.GetRequestForm().Get("response_mode") == "form_post"`. Since `Form()` and `Fragment` and `Query` are almost all very similar I think we could simplify this also and avoid duplication.

One question is how `response_mode: form_post` behaves when we're dealing with implicit flows. I couldn't find any guidance on this but to me it seems non-sensical to have implicit flows work together with `form_post` as implicit applications are typically javascript SPAs or mobile apps where you can't POST to. Do you have any guidance or sources on this?

authorize_helper.go
ajanthan: Yes. It is only used in tests.

authorize_helper.go
ajanthan: +1

authorize_error.go
ajanthan: +1, One small issue with setting default responseMode in the authorize handler. Depending on the flow the default  responseMode changes.

handler/oauth2/flow_authorize_code_auth.go
ajanthan: yes it can be improved

handler/oauth2/flow_authorize_implicit.go
ajanthan: yes it can be improved

handler/openid/flow_hybrid.go
ajanthan: yes it can be improved

handler/openid/flow_hybrid.go
ajanthan: +1

ajanthan:
> Thank you for your hard work on this and the tests and sorry for the late review but it's a bit of a complicated review :)
> 
> I think we can improve this PR a lot - it looks like there is a lot of duplication between `AddForm` and `AddFragment`. I think we should have a general `ResponseMode()` method instead of checking for `ar.GetRequestForm().Get("response_mode") == "form_post"`. Since `Form()` and `Fragment` and `Query` are almost all very similar I think we could simplify this also and avoid duplication.
> 
> One question is how `response_mode: form_post` behaves when we're dealing with implicit flows. I couldn't find any guidance on this but to me it seems non-sensical to have implicit flows work together with `form_post` as implicit applications are typically javascript SPAs or mobile apps where you can't POST to. Do you have any guidance or sources on this?

No, I couldn't find any guidance around this. My thinking was if the response_mode is in the request process it accordingly otherwise process according to the default response mode for the flow. if it doesn't make sense to support `form_post` to the implicit flow, I don't mind it to restrict it to `fragment` only.

ajanthan:
@aeneasr Did you get a chance to review the latest changes?

authorize_helper.go
aeneasr: In that case I would prefer moving this to a `_test` file!

aeneasr (commented):
Sorry for the slow review, my week was extremely busy. Thank you for the hard work, I have some more ideas for improvement!

authorize_write.go
```diff
@@ -49,27 +40,44 @@ func (f *Fosite) WriteAuthorizeResponse(rw http.ResponseWriter, ar AuthorizeRequ
...
+		rw.Header().Add("Content-Type", "text/html;charset=UTF-8")
+		WriteAuthorizeFormPostResponse(redir.String(), resp.GetParameters(), rw)
+	case ResponseModeQuery, ResponseModeNone:
+		// Explicit grants
+		q := redir.Query()
+		rq := resp.GetParameters()
+		for k := range rq {
+			q.Set(k, rq.Get(k))
+		}
+		redir.RawQuery = q.Encode()
```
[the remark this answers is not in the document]
ajanthan: It is done to preserve the query parameters from the `redirect_uri`. This logic is inherited from the previous implementation. There are test cases for this scenario.

authorize_write.go
```diff
@@ -49,27 +40,44 @@ func (f *Fosite) WriteAuthorizeResponse(rw http.ResponseWriter, ar AuthorizeRequ
...
+		sendRedirect(redir.String(), rw)
+	case ResponseModeFragment:
+		// Implicit grants
+		// The endpoint URI MUST NOT include a fragment component.
+		redir.Fragment = ""
 
-	fr := resp.GetFragment()
-	if len(fr) > 0 {
+		u := redir.String()
```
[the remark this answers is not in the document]
ajanthan: Doing so double encode fragment. Once while doing `resp.GetParameters().Encode()` and then`redir.String()`

authorize_write.go
```diff
@@ -49,27 +40,44 @@ func (f *Fosite) WriteAuthorizeResponse(rw http.ResponseWriter, ar AuthorizeRequ
...
-	// user-agent to the provided client redirection URI using an HTTP
-	// redirection response, or by other means available to it via the
-	// user-agent.
-	wh.Set("Location", u)
+// https://tools.ietf.org/html/rfc6749#section-4.1.1
+// When a decision is established, the authorization server directs the
+// user-agent to the provided client redirection URI using an HTTP
+// redirection response, or by other means available to it via the
+// user-agent.
+func sendRedirect(url string, rw http.ResponseWriter) {
```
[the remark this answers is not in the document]
ajanthan: It requires `*http.Request` as one of the arguments and within `WriteAuthorizeResponse` method it is not available. passing `nil` results in invalid memory access.

authorize_helper.go
ajanthan: This helper function is used in tests in `fosite` package and `integration_test` package. The test helper functions declared in `_test` file are only visible within the same package. Any idea of how a test helper function can be shared between packages?

authorize_error.go
ajanthan: Sorry. I missed it somehow. It is done now.

mitar:
> Do you have any guidance or sources on this?

So [this is the spec](https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html) for form post. In there it says:

> Any technique supported by the User Agent MAY be used to cause the submission of the form, and any form content necessary to support this MAY be included, such as submit controls and client-side scripting commands. However, the Client MUST be able to process the message without regard for the mechanism by which the form submission was initiated. 

Moreover, a non-normative request/response/request example they have in there does have `response_type=id_token&response_mode=form_post` in the authorization request. I think this is implicit flow, no? Where you directly get the `id_token` in the response which is form POSTed to the callback.

I must say that I do not understand why server simply does not do the form POST directly to the callback, why they have this JavaScript doing it? s this done just so that any cookies to the callback URL which browser has go together with this form POST? Not sure why that would be needed, but sure.

I am not sure if it is really possible to talk about implicit/explicit flows here. I think from post is just its own beast. If implemented using the JavaScript post in the client, in fact it can work for SPAs as well if SPA implements a service worker which intercepts the POST (there are maybe other ways, too).

I think that the whole idea of `response_mode` is that it should work with any flow type because it is not really about the flow type. Flow type just configures the default `response_mode`. And one can override it. If it makes sense to override it for a particular flow type? We can leave this to the client.

mitar:
I think it should be allowed for client to specify which response modes it allows. We should maybe have `ResposneModeClient` `interface` with `GetResponseModes` method?

authorize_write.go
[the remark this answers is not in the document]
mitar: Yes, this is ugly but there is no `RawFragment` (and newer versions of go where there is, it does not have the same semantics as `RawQuery`). So this has to be done like @ajanthan made and this is also how it was before.

I would suggest though that `fr.Encode()` + `plusMatch.ReplaceAllString(u, "%20")` is made into a helper function like `FormUrlEncode`. I am not sure though why we are replacing `+` with `%20`, because [`application/x-www-form-urlencoded`](https://www.w3.org/TR/1999/REC-html401-19991224/interact/forms.html#h-17.13.4.1) does use `+` for space (`application/x-www-form-urlencoded` is what fragment should be encoded with by OAuth spec).

authorize_write.go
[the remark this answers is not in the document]
mitar: Yes to @ajanthan.

authorize_request.go
```diff
@@ -25,12 +25,22 @@ import (
 	"net/url"
 )
 
+type ResponseModeType string
+
+const (
+	ResponseModeNone     = ResponseModeType("")
```
mitar, line 31: Rename this `ResponseModeDefault`. Because there is also `ResponseModeNone`: https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#none

mitar: And also implement it please. :-)

authorize_request_handler.go
```diff
@@ -210,6 +210,23 @@ func (f *Fosite) validateResponseTypes(r *http.Request, request *AuthorizeReques
 	request.ResponseTypes = responseTypes
 	return nil
 }
+func (f *Fosite) validateResponseMode(r *http.Request, request *AuthorizeRequest) error {
```
mitar, line 213: This validation function should also respect security considerations from [here](https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Security):

> There are security implications to encoding response values in the query string. The HTTP Referer header includes query parameters, and so any values encoded in query parameters will leak to third parties. Thus, while it is safe to encode an Authorization Code as a query parameter when using a Confidential Client (because it can't be used without the Client Secret, which third parties won't have), more sensitive information such as Access Tokens and ID Tokens MUST NOT be encoded in the query string. In no case should a set of Authorization Response parameters whose default Response Mode is the fragment encoding be encoded using the query encoding.

Also see [description of combinations](https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Combinations).

Based on that it is not really true that all combinations are possible. For example, if default allows fragment, then query cannot be used. Moreover, for errors, it seems that you can use always whichever the client requested.

So this is really that if by default query string is used, you can request that fragment is used. But not the opposite.

authorize_response.go
```diff
@@ -54,24 +52,13 @@ func (a *AuthorizeResponse) AddHeader(key, value string) {
 	a.Header.Add(key, value)
 }
 
-func (a *AuthorizeResponse) GetQuery() url.Values {
-	return a.Query
+func (a *AuthorizeResponse) GetParameters() url.Values {
+	return a.Parameters
 }
 
-func (a *AuthorizeResponse) GetFragment() url.Values {
```
mitar, line 61: I like this cleanup. It was always strange to me that callers had to know where should they put parameters (fragment or query).

authorize_helper.go
```diff
@@ -22,14 +22,34 @@
...
+	"time"
+
+	"golang.org/x/net/html"
+	goauth "golang.org/x/oauth2"
 
 	"github.com/asaskevich/govalidator"
 	"github.com/pkg/errors"
 )
 
+var formPostTemplate = template.Must(template.New("form_post").Parse(`<html>
```
mitar, line 40: Can you make this configurable?

authorize_helper.go
```diff
@@ -22,14 +22,34 @@
...
+   <head>
+      <title>Submit This Form</title>
+   </head>
+   <body onload="javascript:document.forms[0].submit()">
+      <form method="post" action="{{ .RedirURL }}">
```
mitar, line 45: `RedirURL` should be probably escaped before blindly included in the HTML template?

authorize_request_handler.go
mitar: Given that for error any mode can be used, then maybe checking for valid combinations cannot be done here but in write handler?

authorize_helper.go
```diff
@@ -22,14 +22,34 @@
...
+         {{ range $key,$value := .Parameters }}
+		<input type="hidden" name="{{$key}}" value="{{index $value 0}}"/>
```
mitar, line 47: Just for completeness (Because other encoding methods like query and fragment supports that), can you loop here over the value (which is a slice) and repeat `<input` with same `key` multiple times? This should never happen. But you never know if somebody directly modifies `Parameters` to add that (given that you defined it as `url.Values`).

mitar: Also, HTML-escape key and value here.

mitar: And add test for that.

handler/oauth2/flow_authorize_code_auth_test.go
```diff
@@ -122,11 +122,40 @@ func TestAuthorizeCode_HandleAuthorizeEndpointRequest(t *testing.T) {
...
+							GrantedScope:      fosite.Arguments{"a", "b"},
+							Session: &fosite.DefaultSession{
+								ExpiresAt: map[fosite.TokenType]time.Time{fosite.AccessToken: time.Now().UTC().Add(time.Hour)},
+							},
+							RequestedAt: time.Now().UTC(),
+						},
+						State:       "superstate",
+						RedirectURI: parseUrl("https://asdf.de/cb"),
+					},
+					description: "Default responseMode check",
```
aeneasr, line 151: To me it looks like the previous tests all cover the "default responseMode" case. I think we should probably add the explicit query response mode and also the fragment response mode as a test!

handler/oauth2/flow_authorize_implicit.go
```diff
@@ -52,7 +52,9 @@ func (c *AuthorizeImplicitGrantTypeHandler) HandleAuthorizeEndpointRequest(ctx c
 	if !ar.GetResponseTypes().ExactOne("token") {
 		return nil
 	}
-
+	if ar.GetResponseMode() == fosite.ResponseModeNone {
```
aeneasr, line 55: Use `ar.SetDefaultResponseMode`. And also, as per https://github.com/ory/fosite/pull/509/files#r511293832 this handler should not run if `fosite.ResponseModeQuery` and also `FormPost` in my opinion. Please confirm @mitar .

handler/oauth2/flow_authorize_implicit_test.go
```diff
@@ -138,3 +131,49 @@ func TestAuthorizeImplicit_EndpointHandler(t *testing.T) {
 		})
 	}
 }
+func makeAuthorizeImplicitGrantTypeHandler(ctrl *gomock.Controller) (AuthorizeImplicitGrantTypeHandler,
```
aeneasr, line 134: We should have a test which checks if it is allowed to set the response mode to query while asking to perform the implicit flow.

handler/openid/flow_hybrid.go
```diff
@@ -54,7 +54,9 @@ func (c *OpenIDConnectHybridHandler) HandleAuthorizeEndpointRequest(ctx context.
 	if !(ar.GetResponseTypes().Matches("token", "id_token", "code") || ar.GetResponseTypes().Matches("token", "code") || ar.GetResponseTypes().Matches("id_token", "code")) {
 		return nil
 	}
-
+	if ar.GetResponseMode() == fosite.ResponseModeNone {
```
aeneasr, line 57: I think this should not run if `fosite.ResponseModeQuery` and maybe also not with the `FormPost` one. Please confirm @mitar . Also use `ar.SetDefaultResponseMode`.

handler/openid/flow_hybrid_test.go
```diff
@@ -249,9 +249,22 @@ func TestHybrid_HandleAuthorizeEndpointRequest(t *testing.T) {
...
-				assert.NotEmpty(t, aresp.GetFragment().Get("code"))
-				assert.NotEmpty(t, aresp.GetFragment().Get("access_token"))
+				assert.NotEmpty(t, aresp.GetParameters().Get("id_token"))
+				assert.NotEmpty(t, aresp.GetParameters().Get("code"))
+				assert.NotEmpty(t, aresp.GetParameters().Get("access_token"))
+				assert.Equal(t, time.Now().Add(time.Hour).UTC().Round(time.Second), areq.GetSession().GetExpiresAt(fosite.AuthorizeCode))
+			},
+		},
+		{
+			description: "Default responseMode check",
```
aeneasr, line 259: We should add a test for the concrete response modes and their expected results (e.g. error)

handler/openid/flow_implicit.go
```diff
@@ -51,6 +51,10 @@ func (c *OpenIDConnectImplicitHandler) HandleAuthorizeEndpointRequest(ctx contex
 		return nil
 	}
 
+	if ar.GetResponseMode() == fosite.ResponseModeNone {
```
aeneasr, line 54: I think this should not run if `fosite.ResponseModeQuery` and maybe also not with the `FormPost` one. Please confirm @mitar . Also use `ar.SetDefaultResponseMode`.

handler/openid/flow_implicit_test.go
```diff
@@ -241,9 +241,24 @@ func TestImplicit_HandleAuthorizeEndpointRequest(t *testing.T) {
...
-				assert.NotEmpty(t, aresp.GetFragment().Get("id_token"))
-				assert.NotEmpty(t, aresp.GetFragment().Get("state"))
-				assert.NotEmpty(t, aresp.GetFragment().Get("access_token"))
+				assert.NotEmpty(t, aresp.GetParameters().Get("id_token"))
+				assert.NotEmpty(t, aresp.GetParameters().Get("state"))
+				assert.NotEmpty(t, aresp.GetParameters().Get("access_token"))
+			},
+		},
+		{
+			description: "default responseMode check",
```
aeneasr, line 250: Same as for the others, check for other response modes with expected success or errors.

integration/authorize_form_post_test.go
```diff
@@ -0,0 +1,172 @@
...
+	var state string
+	for k, c := range []struct {
+		description  string
+		setup        func()
+		check        func(t *testing.T, stateFromServer string, code string, token goauth.Token, iDToken string, err fosite.RFC6749Error)
+		responseType string
+	}{
+		{
+			description:  "implicit grant test with form_post",
+			responseType: "token",
```
aeneasr, line 77: Should this be allowed @mitar ?

aeneasr (changes requested):
Wow @ajanthan this is a really good PR! I think I found a couple more question marks and suggestions for improvement but we're very close to having this merged. Thank you for all the hard work, this is superb :)

handler/oauth2/flow_authorize_code_auth.go
```diff
@@ -77,6 +77,9 @@ func (c *AuthorizeExplicitGrantHandler) HandleAuthorizeEndpointRequest(ctx conte
 		return nil
 	}
 
+	if ar.GetResponseMode() == fosite.ResponseModeNone {
+		ar.SetResponseMode(fosite.ResponseModeQuery)
```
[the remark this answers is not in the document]
mitar: I think response mode spec does not influence what is being returned, only how. And if anything of what is being returned is sensitive, then how it can be returned is restricted.

handler/oauth2/flow_authorize_implicit.go
mitar: I must say I never really understood why we have handler named "implicit". I would have three authorize handlers: one if you asked for response type "token", one if "code", and one if "id_token". And then based on what you asked for, we allow or not allow some response modes.

I think "implicit" is really just a name to easier think about this, but not really something code/logic should know about. No?

aeneasr: So this particular implicit handler can only handle `token` - `id_token` is only enabled if we have support for openid enabled (which needs more config). The original idea was that your oauth2 server could simply not allow/handle implicit at all, if you wanted. This might also be interesting for future OAuth 2.1/2.2 specs which might remove implicit flows in favor of PKCE!

authorize_helper.go
ajanthan: Done

authorize_request.go
ajanthan: +1 for renaming it to `ResponseModeDefault`
Regarding supporting `none`.  @mitar Isn't the link about `response_type` not the `response_mode`?

mitar: Oh, you are right. Sorry, I got confused here.

authorize_write.go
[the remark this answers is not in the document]
ajanthan: Done

fosite.go
```diff
@@ -105,6 +106,9 @@ type Fosite struct {
 
 	// MinParameterEntropy controls the minimum size of state and nonce parameters. Defaults to fosite.MinParameterEntropy.
 	MinParameterEntropy int
+
+	//FormPostHTMLTemplate sets html template for rendering the authorization response when the request has response_mode=form_post. Defaults to fosite.FormPostDefaultTemplate
```
aeneasr, line 110:
```suggestion
	// FormPostHTMLTemplate sets html template for rendering the authorization response when the request has response_mode=form_post. Defaults to fosite.FormPostDefaultTemplate
```

authorize_helper.go
ajanthan: Done

authorize_helper.go
```diff
@@ -182,3 +202,105 @@ func IsLocalhost(redirectURI *url.URL) bool {
 	hn := redirectURI.Hostname()
 	return strings.HasSuffix(hn, ".localhost") || hn == "127.0.0.1" || hn == "::1" || hn == "localhost"
 }
+
+func WriteAuthorizeFormPostResponse(redirectURL string, parameters url.Values, rw io.Writer) {
+	_ = formPostTemplate.Execute(rw, struct {
```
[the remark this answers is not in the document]
ajanthan: The caller of this method is WriteAuthorizeError and WriteAuthorizeResponse. None of them return an error or have log. Any idea how this error can be handled? Maybe panicking from here?

authorize_helper_test.go
```diff
@@ -257,6 +259,17 @@ func TestIsRedirectURISecure(t *testing.T) {
 	}
 }
 
+func TestWriteAuthorizeFormPostResponse(t *testing.T) {
+	var responseBuffer bytes.Buffer
+	redirectURL := "https://localhost:8080/cb"
+	parameters := url.Values{"code": {"lshr755nsg39fgur"}, "state": {"924659540232"}}
```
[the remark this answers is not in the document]
ajanthan: Done

authorize_request_handler.go
ajanthan: Added the check for using `query` while the default mode is `fragment` and returns an error in the `autheorize_response_writer.go`.

authorize_helper.go
ajanthan: Now the template supports adding multiple values. Also added test case for illegal character encoding.

handler/openid/flow_hybrid.go
ajanthan: added `ar.SetDefaultResponseMode`. Using `query` in this flow is already restricted but do we want to restrict `form_post`? According to the [spec](https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html#Security), it seems not necessary to restrict `form_post` as it more secure than `query` and  `fragment`.

handler/openid/flow_hybrid_test.go
ajanthan: Agree. Added error test case in `authorize_response_writer_test.go`

handler/openid/flow_implicit.go
ajanthan: Added `ar.SetDefaultResponseMode`. Using `query` in this flow is already restricted but do we want to restrict `form_post`?

handler/openid/flow_implicit_test.go
ajanthan: +1 but it doesn't make sense to check success/error for other modes because the validation happens in `authorize_response_writer`. I have added a test case there.

integration/authorize_form_post_test.go
ajanthan: Changed it to `id_token token`.

handler/openid/flow_implicit_test.go
ajanthan: Also added integration test to cover a various combination of response types and modes

handler/oauth2/flow_authorize_code_auth_test.go
ajanthan: +1

handler/oauth2/flow_authorize_implicit_test.go
ajanthan: The validation happens outside of the handler so we can't test here. Added the scenario in integration tests.

aeneasr:
Quick update from my side - this looks great! My plan would be to introduce an automated test pipeline using https://gitlab.com/openid/conformance-suite to ensure that all conformance tests pass as well. Unfortunately, this is quite the amount of work so it might need a bit of time before this is implemented. I have some other stuff to finish but this is next on my list afterwards!

aeneasr:
I started working on this and already gotten pretty far (the set up works and the first tests are passing). I still have to figure out how to do this using automated tools (maybe I'll just use Cypress). But my feeling says this should be workable pretty soon!

mitar:
I like Cypress. BTW, [you can also check here how they did it](https://github.com/panva/node-oidc-provider/tree/master/certification).

aeneasr:
Ah nice, I see how they solved the browser interaction. I was looking for this all night yesterday :D

aeneasr:
I am making good progress here with the automation. Looks like almost everything is working as expected. Bad news is, the recent clean up of `error_description` actually breaks the conformity tests. But that should not block this PR!

I'll try to get all the tests working, then I will do a rewrite to this fosite branch with hydra and to check if `form_post` is implemented correctly. If it is, I will do another final review pass and then this is good to merge. Thank you for your patience!

aeneasr:
Bildschirmfoto 2020-11-03 um 14 57 17

mitar:
Nice!

> Bad news is, the recent clean up of error_description actually breaks the conformity tests.

Oh, what? Maybe open an issue with more information about this. This is surprising.

mitar:
BTW, have you disabled some checks? Because I would assume https://github.com/ory/fosite/issues/409 would fail the testing.

aeneasr:
Certification works in such a way that you choose certain profiles. Also, not everything has to pass (e.g. none signature is not required to be implemented). So "orange" indicators are ok most of the times, black/red indicators are failures.

aeneasr:
With a few fixes, most tests pass - there is one particular test suite which doesn't pass but I have some ideas why that could be. I will be working on this over the next days!

Screenshot_2020-11-05 OIDF Conformance Test Plan

aeneasr:
Ok so as far as I can tell the only thing that doesn't work is properly extracting the response_mode from the request object URL or JWT if one is set. We had this bug actually for `state` as well so I know where and what to fix. Other than that, I think this is pretty much good to go. Once I have confirmed the fix and ran all the tests I will give this one last code review and then I think we can merge this. Thank you for your work and commitment and patience @ajanthan . @vinckr please make sure @ajanthan gets some goodies from us :)

aeneasr:
Please do not be confused by the random pushes, I am trying to debug something...

The PR itself is good to go IMO

authorize_request_handler_oidc_request_test.go
```diff
@@ -166,10 +165,11 @@ func TestAuthorizeRequestParametersFromOpenIDConnectRequest(t *testing.T) {
...
 		{
-			d:          "should pass and set request parameters properly",
-			form:       url.Values{"scope": {"openid"}, "request": {validRequestObject}},
-			client:     &DefaultOpenIDConnectClient{JSONWebKeys: jwks, RequestObjectSigningAlgorithm: "RS256"},
-			expectForm: url.Values{"scope": {"foo openid"}, "request": {validRequestObject}, "foo": {"bar"}, "baz": {"baz"}},
+			d:      "should pass and set request parameters properly",
+			form:   url.Values{"scope": {"openid"}, "response_type": {"code"}, "response_mode": {"none"}, "request": {validRequestObject}},
+			client: &DefaultOpenIDConnectClient{JSONWebKeys: jwks, RequestObjectSigningAlgorithm: "RS256"},
+			// The values from form are overwritten by the request object.
+			expectForm: url.Values{"response_type": {"token"}, "response_mode": {"post_form"}, "scope": {"foo openid"}, "request": {validRequestObject}, "foo": {"bar"}, "baz": {"baz"}},
```
mitar, line 172: Any particular reason why this file contains `post_form` and not correct `form_post`? What is this testing? Why it is not failing?

mitar:
So changelog currently reads:

> This patch introduces support for response_mode=form_post as well as response_mode of none and query and fragment.

I do not think this is correct. This PR did not introduce `response_mode=none`? That I think was just temporary implementation detail which was later renamed to "default" (so if `response_mode` is not provided). I think changelog should be probably fixed and mention of "none" removed?

mitar:
Also, it seems this PR introduced a backwards incompatibility into fosite: if before this change, clients were providing `response_mode`, fosite ignored it and everything worked (if they were not asking for a non-default response mode). With just upgrading to a newer fosite (and when we use a custom `Client` implementation) such requests fail.

I see two solutions:
* `Client` interface should simply get this new `GetResponseModes` method. After upgrading, compilation would fail, I would see what is missing, add it (based on default implementation), it is a simple change, and things will work as expected.
* If client does not implement `GetResponseModes`, then `response_mode` parameter should simply be ignored, and not processed at all.

I think this maybe speaks more about fosite approach to backwards compatibility. I think it is great that compiler helps you point out what has to change to get things upgraded and I would prefer the first approach above. With a good message in the changelog I would know to expect this and would decide to do the upgrade with enough time budgeted to do the code change.

I think the introduction of a new interface makes things look like backwards compatible but in fact it was at the end much more work to discover the problem, get it reported internally, fix it, review the fix, and now me reporting it back here.

I know that I [proposed a new interface](https://github.com/ory/fosite/pull/509#issuecomment-714326385) but I was doing that based on prior practice in fosite. I think, @aeneasr, maybe this practice should be changed and we should embrace changing interfaces more instead of introducing new interfaces?

PR #516, opened 2020-10-20
feat: rfc8693 token exchange impersonation flow (#1218)

mpnunes:
https://github.com/ory/hydra/issues/1218

## Proposed changes

Added token exchange flow for impersonation as described in rfc8693. 

## Checklist

      the changes.

## Further comments

The majority of the implementation was done following a similar structure as for the already existing flows.
After several internal discussions about how to implemented the check of who is allowed to exchange a token (besides requiring the corresponding grant type - see 4.4 in rfc8693) we opted to attach this permission to the subjects client under may_act by adding the client ids of those that should be allowed exchange its token. This decision was made with the purpose of enabling a client to revoke a previously given permission to exchange its token.


client.go
```diff
@@ -50,6 +54,9 @@ type Client interface {
 
 	// GetAudience returns the allowed audience(s) for this client.
 	GetAudience() Arguments
+
+	// GetMayAct returns the allowed exchange actor(s) for this client
+	GetMayAct() Arguments
```
mitar, line 59: I think you should move this to a new interface, like `TokenExchangeClient`?

access_request.go
```diff
@@ -41,3 +44,7 @@ func NewAccessRequest(session Session) *AccessRequest {
 func (a *AccessRequest) GetGrantTypes() Arguments {
 	return a.GrantTypes
 }
+
+func (a *AccessRequest) SetDelegatingClient(client Client) {
```
mitar, line 48: I guess you should move this to an interface as well? `TokenExchangeAccessRequestor`?

oauth2.go
```diff
@@ -236,13 +236,18 @@ type Requester interface {
...
+	// GetDelegatingClient returns the client that provided the original token (subject token) in a token exchange flow.
+	GetDelegatingClient() (client Client)
 }
 
 // AccessRequester is a token endpoint's request context.
 type AccessRequester interface {
 	// GetGrantType returns the requests grant type.
 	GetGrantTypes() (grantTypes Arguments)
 
+	SetDelegatingClient(client Client)
```
mitar, line 249: Not backwards compatible to change interfaces like this.

client.go
mpnunes: Do you mean by extending the Client interface?
What would be the advantage of that?

access_request.go
mpnunes: Do you mean by extending the AccessRequest interface?
What would be the advantage of that?

client.go
mitar: No, not changing the `Client` interface, but creating a new interface. Otherwise any existing implementation of `Client` interface will break because you added an additional method.

mpnunes: I believe you miss understood me, I meant a new interface that extends Client.

mitar: I agree that it is easier to change the interface, but that does bring issues with existing implementations of Client interface. If I upgrade to a new version of fosite with this change, my local Client implementation will not work anymore. Is this acceptable? The fix is trivial though (just implement this extra method).

mpnunes: I agree, the fix is trivial, but will have several implications on the remaining implementation. I'll check on those.

aeneasr (commented):
Just two things that sprung my eye. Unfortunately I had not time to dive into this deeper. I first have to read the RFC and will then revisit. I hope I can get this done next week. Is there a test/conformity suite we can use to test the implementation?

handler/oauth2/flow_refresh.go
```diff
@@ -90,8 +94,27 @@ func (c *RefreshTokenGrantHandler) HandleTokenEndpointRequest(ctx context.Contex
...
+			return errors.WithStack(fosite.ErrUnauthorizedClient.WithHint("The OAuth 2.0 Client is not allowed to perform a token exchange for the given subject token."))
+		}
+	}
+
 	for _, scope := range originalRequest.GetGrantedScopes() {
-		if !c.ScopeStrategy(request.GetClient().GetScopes(), scope) {
+		if !c.ScopeStrategy(request.GetClient().GetScopes(), scope) &&
+			// if it is a refresh of an exchanged token, check scopes against those of the delegating client
+			(delegatingClient == nil || !c.ScopeStrategy(delegatingClient.GetScopes(), scope)) {
 			return errors.WithStack(fosite.ErrInvalidScope.WithHintf("The OAuth 2.0 Client is not allowed to request scope \"%s\".", scope))
```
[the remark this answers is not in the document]
mpnunes: I think there may be a miss understanding.
Here we check that the granted scopes are either available on the requesting client as previously (in the case this is a refresh of an exchanged token, this would be the client that performed the token exchange), or on the scopes of the delegating client (the client that provided the token that  was exchanged).
This was an implementation decision we made in which we allow a client (performing the token exchange) to request scopes available on the delegating client or on its own client.
If there is a delegating client (implies that the refresh flow is being done for an exchanged token) the requested scopes can be declared on either of the clients.
At this point we cannot distinguish, if the requested scopes should be on the delegating client or on the own client.

handler/oauth2/flow_token_exchange.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	if client.IsPublic() {
+		return errors.WithStack(fosite.ErrInvalidGrant.WithHint("The OAuth 2.0 Client is marked as public and is thus not allowed to use authorization grant \"urn:ietf:params:oauth:grant-type:token-exchange\"."))
+	}
+
+	request.GetSession().SetExpiresAt(fosite.AccessToken, time.Now().UTC().Add(c.AccessTokenLifespan))
+	if c.RefreshTokenLifespan > -1 {
+		request.GetSession().SetExpiresAt(fosite.RefreshToken, time.Now().UTC().Add(c.RefreshTokenLifespan).Round(time.Second))
+	}
```
aeneasr, line 126: I assume this is needed because `RefreshTokenGrantHandler` doesn't handle this `grant_type`? However, to me it also looks like `requested_token_type` should be responsible for setting this.

handler/oauth2/flow_token_exchange.go
```diff
@@ -0,0 +1,170 @@
...
+	RefreshTokenStrategy     RefreshTokenStrategy
+	RefreshTokenLifespan     time.Duration
+	RefreshTokenScopes       []string
+	CoreStrategy
+	CoreStorage
+	Store fosite.Storage
+}
+
+// HandleTokenEndpointRequest implements https://tools.ietf.org/html/rfc8693#section-2.1 (currently impersonation only)
+func (c *TokenExchangeGrantHandler) HandleTokenEndpointRequest(ctx context.Context, request fosite.AccessRequester) error {
```
aeneasr, line 51: To me it looks like `actor_token` and `actor_token_type` are currently ignored. If they are not implemented, we should throw an error instead of silently accepting their values (and potentially returning a success). To be honest, I also don't really understand what the value of `actor_token` would be and neither its implications. But as far as I can tell those are values used when performing delegation, and this code currently only supports impersonation.

aeneasr (changes requested):
Thank you! This looks very well thought out and written.

I finally had some time off to dive deep into this PR and the accompanying spec. I left a lot of comments - please don't feel discouraged :) But as you can imagine, this is a pretty serious security component and as such it needs tough and thorough review. Please be prepared to walk through several code review iterations before this can be merged and let me know beforehand if you do not have the time to do them. :)

Besides code comments, I also have some general questions. The spec asys:

>    In the absence of one-time-use or other semantics specific to the
   token type, the act of performing a token exchange has no impact on
   the validity of the subject token or actor token.  Furthermore, the
   exchange is a one-time event and does not create a tight linkage
   between the input and output tokens, so that (for example) while the
   expiration time of the output token may be influenced by that of the
   input token, renewal or extension of the input token is not expected
   to be reflected in the output token's properties.  **It may still be
   appropriate or desirable to propagate token-revocation events.
   However, doing so is not a general property of the STS protocol and
   would be specific to a particular implementation, token type, or
   deployment.**

To me it looks like token revokation would currently not have an impact on impersonated or delegated tokens using this flow. Is this by design, and if so why?

>    An authorization server may be unwilling or unable to fulfill any
   token request, but the likelihood of an unfulfillable request is
   significantly higher when very broad access rights are being
   solicited.  As such, in the absence of specific knowledge about the
   relationship of systems in a deployment, clients should exercise
   discretion in the breadth of the access requested, particularly the
   number of target services.  An authorization server can use the
   "invalid_target" error code, defined in Section 2.2.2, to inform a
   client that it requested access to too many target services
   simultaneously.

The `invalid_target` error should be in the default errors exposed by fosite in my opinion.

access_request.go
mpnunes: This is impersonation only.

aeneasr: Ok, so this would set the OAuth2 Client which the `subject_token` was granted for in the Token Exchange request, correct? If so, maybe we rename it to `SetSubjectTokenClient` to allow further extension (in the future) for `SetActorTokenClient`? And then I think it makes sense to define another interface as @mitar suggested:


```
// in handler/rfc8693/access_request.go

type interface {
  SetDelegatingClient(client Client)
}
```

oauth2.go
aeneasr: Yes, see my last comment in https://github.com/ory/fosite/pull/516/files#r509957103

client.go
mpnunes: Modified as suggested.

mpnunes: sorry, made a mistake with the naming, will correct.

aeneasr:
@mpnunes let me know when this is good for another 👀

handler/oauth2/flow_token_exchange.go
```diff
@@ -0,0 +1,170 @@
...
+	if !delegatingClient.GetMayAct().HasOneOf(client.GetID()) {
+		return errors.WithStack(fosite.ErrUnauthorizedClient.WithHint("The OAuth 2.0 Client is not allowed to perform a token exchange for the given subject token."))
+	}
+	request.SetDelegatingClient(delegatingClient)
+
+	for _, scope := range request.GetRequestedScopes() {
+		if !c.ScopeStrategy(client.GetScopes(), scope) &&
+			!c.ScopeStrategy(or.GetGrantedScopes(), scope) {
+			return errors.WithStack(fosite.ErrInvalidScope.WithHintf("The OAuth 2.0 Client is not allowed to request scope \"%s\".", scope))
+		}
```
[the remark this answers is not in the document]
mpnunes: This would not work. If  you run the tests against this implementation you can see why.

handler/oauth2/flow_token_exchange.go
```diff
@@ -0,0 +1,170 @@
+/*
```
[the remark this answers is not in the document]
mpnunes: Moved.

handler/oauth2/flow_token_exchange.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	var delegatingClient fosite.Client = nil
+	// reload client from storage to ensure that may_act is up to date in case of eventual revocation
+	if or.GetDelegatingClient() == nil {
+		// first exchange request has no delegating client set
+		delegatingClient, err = c.Store.GetClient(ctx, or.GetClient().GetID())
+	} else {
+		delegatingClient, err = c.Store.GetClient(ctx, or.GetDelegatingClient().GetID())
+	}
```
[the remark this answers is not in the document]
mpnunes: Correct, the idea is that client B exchanges a token from A and C can exchange the token from B (who is actually impersonating A)

handler/oauth2/flow_token_exchange.go
```diff
@@ -0,0 +1,170 @@
...
+			!c.ScopeStrategy(or.GetGrantedScopes(), scope) {
+			return errors.WithStack(fosite.ErrInvalidScope.WithHintf("The OAuth 2.0 Client is not allowed to request scope \"%s\".", scope))
+		}
+	}
+
+	if err := c.AudienceMatchingStrategy(client.GetAudience(), request.GetRequestedAudience()); err != nil {
+		return err
+	}
+
+	if client.IsPublic() {
```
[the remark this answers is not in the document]
mpnunes: Moved up.

access_request.go
mpnunes: Extracted interface but kept it in same file/package due to otherwise resulting cyclical dependencies.

oauth2.go
mpnunes: Extracted interface but kept it in same file/package due to otherwise resulting cyclical dependencies, as stated in previous comment.

handler/oauth2/flow_token_exchange.go
```diff
@@ -0,0 +1,170 @@
...
+
+	sig := c.CoreStrategy.AccessTokenSignature(subjectToken)
+	or, err := c.CoreStorage.GetAccessTokenSession(ctx, sig, request.GetSession())
+	if err != nil {
+		return errors.WithStack(fosite.ErrRequestUnauthorized.WithDebug(err.Error()))
+	} else if err := c.CoreStrategy.ValidateAccessToken(ctx, or, subjectToken); err != nil {
+		return err
+	}
+
+	var delegatingClient fosite.Client = nil
```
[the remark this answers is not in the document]
mpnunes: Correct, the delegating client is the "original principal", the client that requested the original token being (re)exchanged.

client.go
```diff
@@ -90,6 +97,7 @@ type DefaultClient struct {
 	Scopes        []string `json:"scopes"`
 	Audience      []string `json:"audience"`
 	Public        bool     `json:"public"`
+	MayAct        []string `json:"may_act"`
```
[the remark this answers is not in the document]
mpnunes: > To me, it is not clear if may_act allows another client to delegate to, or the other way around.

The may_act value(s) of a client define which clients are allowed to exchange the tokens of that client. I'm not sure if that is what you mean.

> I assume that may_act was chosen based on this new claim but to me it looks like that claim would be expected in the subject_token?

Correct, may_act was chosen based on the new claim. `delegatable_by` or `impersonatable_by` would bind it semanticaly to one of the exchange modes (impersonation or delegation) while according to the specification it can be used for both.

> The claim might be used, for example, when a subject_token is
presented to the token endpoint in a token exchange request and may_act claim in the subject
token can be used by the authorization server to determine whether the client (or party
identified in the actor_token) is authorized to engage in the requested delegation or
impersonation.

You are completely right, according to the specification this new claim would be expected in the subject_token.
As I tried to explain in https://github.com/ory/hydra/issues/1218 the claim representation described in the specification does not allow to provide more than one authorized actor within the subject token (please correct me if I'm wrong), something that is necessary for enabling delegation chains as implicitly stated e.g. in the description of the actor claim.

The other aspect why we decided to evaluate this internally over the client was, on the one side the fact that we need anyway to add that information to the clients, but also because we found reasonable to allow clients to revoke a a previously given permission to some client to exchange its tokens.

handler/oauth2/flow_token_exchange.go
```diff
@@ -0,0 +1,170 @@
...
+	} else {
+		delegatingClient, err = c.Store.GetClient(ctx, or.GetDelegatingClient().GetID())
+	}
+
+	if err != nil {
+		return errors.WithStack(fosite.ErrInvalidClient.WithHint("The delegating OAuth2 Client does not exist."))
+	}
+
+	// check if delegating client allows the current client to perform an exchange on its tokens
+	if !delegatingClient.GetMayAct().HasOneOf(client.GetID()) {
```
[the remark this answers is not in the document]
mpnunes: I would rather say, the `delegatingClient` would be that of _user B_, while `client` the one of `resource server A`. User B allows with the `may_act` entries client A to exchange its token (from B) to call C on its behalf.

handler/oauth2/flow_token_exchange.go
```diff
@@ -0,0 +1,170 @@
...
+		if !c.ScopeStrategy(client.GetScopes(), scope) &&
+			!c.ScopeStrategy(or.GetGrantedScopes(), scope) {
+			return errors.WithStack(fosite.ErrInvalidScope.WithHintf("The OAuth 2.0 Client is not allowed to request scope \"%s\".", scope))
+		}
+	}
+
+	if err := c.AudienceMatchingStrategy(client.GetAudience(), request.GetRequestedAudience()); err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
mpnunes: Since this parameter is optional we thought about leaving it to be extended later.

handler/oauth2/flow_refresh.go
```diff
@@ -90,8 +94,27 @@ func (c *RefreshTokenGrantHandler) HandleTokenEndpointRequest(ctx context.Contex
 	request.SetRequestedScopes(originalRequest.GetRequestedScopes())
 	request.SetRequestedAudience(originalRequest.GetRequestedAudience())
 
+	var delegatingClient fosite.Client = nil
+	if originalRequest.GetDelegatingClient() != nil {
```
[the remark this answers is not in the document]
mpnunes: I'm not quite sure what you mean.
Here we get the original client/principal in order to validate its scopes and may_act values.
From your point of view, what would be the reason to keep/follow the whole chain of delegation?

handler/oauth2/flow_token_exchange.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+		if refreshSignature != "" {
+			if err := c.CoreStorage.CreateRefreshTokenSession(ctx, refreshSignature, request.Sanitize([]string{})); err != nil {
+				if rollBackTxnErr := storage.MaybeRollbackTx(ctx, c.CoreStorage); rollBackTxnErr != nil {
+					err = rollBackTxnErr
+				}
+				return errors.WithStack(fosite.ErrServerError.WithDebug(err.Error()))
+			}
+		}
+		response.SetExtra("refresh_token", refresh)
```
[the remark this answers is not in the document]
mpnunes: I'm not so sure if this should be seen as an edge scenario. As a matter of fact it is a scenario that we have.
The logic we use here is pretty much the same as the the one already used e.g. `flow_authorize_code_token`
How are such aspects to be documented here?

handler/oauth2/flow_token_exchange.go
mpnunes: The logic we use here is pretty much the same as the the one already used e.g. `flow_authorize_code_token`.
I disagree, regarding the use of `requested_token_type`.

> requested_token_type
OPTIONAL . An identifier, as described in Section 3, for the type of the requested security
token. ...

In my understanding, a refresh token is not really a security token. I also do not see the point of making an exchange request just to get an refresh token.
Perhaps I miss understood you?

handler/oauth2/flow_token_exchange.go
mpnunes: > If they are not implemented, we should throw an error instead of silently accepting their values

I agree and will add the checks for that.

> To be honest, I also don't really understand what the value of actor_token would be and neither its implications

We had exactly the same problem. After several discussions on that we came up with a scenario for which it may make sense.
E.g. User A want to allow User B to access service D over service C. Assuming Service C has tokens of both users, it exchanges the token from B (actor) with that from A (subject) and calls D with the resulting token.
Unfortunately I could not find further information around to clarify that.

mpnunes:
> To me it looks like token revokation would currently not have an impact on impersonated or delegated tokens using this flow. Is this by design, and if so why?

To be honest I did't thought much about that. Perhaps you can provide some hints about existing mechanisms work in fosite?

> The invalid_target error should be in the default errors exposed by fosite in my opinion.

I will take a look on that.

aeneasr:
> To be honest I did't thought much about that. Perhaps you can provide some hints about existing mechanisms work in fosite?

iirc we have a request id which we track across all the refresh and access tokens. I think impersonated tokens should also be revoked when the original token is revoked. This makes the most sense to me from a security perspective.

I was finally able to run and automate the oidc conformance test suite for #509 so I will be working on this in the next days or so to fix all the pesky bugs and spec violations. Once that's done, I will be able to work FT on this PR as it needs a lot of attention to detail.

Unfortunately, the OIDC Conformity suite doesn't cover RFC8693. But maybe you know a test suite which we could port to fosite from another library that has support for RFC8693 in a different language? Imi a e2e/spec test suite not written by us is mandatory for merging this PR as the stuff that's being touched has a high impact on security.

aeneasr:
Thank you for the call @mpnunes today! I read through the KeyCloak documentation which is the most expansive documentation and implementation of Token Exchange. A few things I observed:

1. They support different token exchange flows - some of them with very custom parameters not specified in the RFC.
1. They do not use the original access token expiry but use the 3600 value - similar to how you solved it in your PR at the moment.
1. They do not really explain how revokation works. I assume they simply create a completely new "strain" of tokens that are separate from the original token.
1. The feature is a preview and needs to be enabled, and has several danger signs.
1. Public clients must not be allowed to exchange tokens.

Regarding my points where we have the original token A, and then do token exchange for token B, and then another token exchange for B to C: This would only be possible if client A has may_act for B, and B has may_act for C. So this is ok. What is absolutely NOT OK is for client A to do token exchange on itself, as this would mean that the client A could just refresh tokens as it may see fit. So please ensure that client A **can not** do token exchange with itself!

mpnunes:
> Regarding my points where we have the original token A, and then do token exchange for token B, and then another token exchange for B to C: This would only be possible if client A has may_act for B, and B has may_act for C. So this is ok.

Currently, for that delegation chain to work, the client A would need to have may_act for B and C. In my opinion it should also stay like that. After the exchanged B and C would be acting as A so I disagree that B should be able to enable C to act as A.

> What is absolutely NOT OK is for client A to do token exchange on itself, as this would mean that the client A could just refresh tokens as it may see fit. So please ensure that client A can not do token exchange with itself!

Good point, I missed that one. Will implement as asap.

aeneasr:
No problem! Let me know if you need further help/guidance :)

aeneasr:
Is there anything I can help you with to get this out? :)

aeneasr:

mpnunes:
> Are you still up for the changes? :) If you need any help, let us know!

I am :) I've just been busy with other stuff.

aeneasr:
This is currently blocked but will eventually be worked on.

Veil:
Hey, I have a vested interest in this RFC's availability in hydra, digging through this pull request am I correct in stating that the may_act semantics only consider a _client's_ permission to exchange rather than that of a _subject's_ permission? For example subject "bob@example.com" is happy for subject "alice@example.com" to impersonate him, regardless of the client performing the exchange?

If I have missed this use case or have entirely misunderstood the RFC, please let me know, but I don't remember seeing this use case handled in the code entering the pull request and I can't see anywhere in the RFC which states that may_act is restricted to just client ids.

Can I also ask what the reasoning behind blanket preventing public clients from performing token exchange is? Surely the decision behind whether a public client is or isn't allowed to perform a token exchange can be made using the same may_act semantics as previously described. There may definitely be valid use cases where one public client may want to delegate to another.

mpnunes:
> Hey, I have a vested interest in this RFC's availability in hydra, digging through this pull request am I correct in stating that the may_act semantics only consider a _client's_ permission to exchange rather than that of a _subject's_ permission? For example subject "[bob@example.com](mailto:bob@example.com)" is happy for subject "[alice@example.com](mailto:alice@example.com)" to impersonate him, regardless of the client performing the exchange?
> 
> If I have missed this use case or have entirely misunderstood the RFC, please let me know, but I don't remember seeing this use case handled in the code entering the pull request and I can't see anywhere in the RFC which states that may_act is restricted to just client ids.
> 
> Can I also ask what the reasoning behind blanket preventing public clients from performing token exchange is? Surely the decision behind whether a public client is or isn't allowed to perform a token exchange can be made using the same may_act semantics as previously described. There may definitely be valid use cases where one public client may want to delegate to another.

Interesting that you bring up that question because I've been also thinking about the current may_act semantic implementation. I believe you are correct in your interpretation of the RFC. Nevertheless I have currently no good solution for two questions around that. The first one is how to configure/set the may_act at when first retrieving a token. The other one is how to allow multiple subjects to exchange the token, one after another as the examples provided in the RFC only "allow" the declaration of one subject within the claim. It looks to me that the specification is somehow inconsistent on that, but perhaps I'm misunderstanding something.
Please fell free to correct me or make a suggestion on how to address those questions.
Furthermore the PR is currently on hold due to organizational reasons but it would be great to ear some ideas on how to address the questions above whenever I can continue the work on it.

handler/rfc8693/flow_token_exchange.go
```diff
@@ -0,0 +1,222 @@
...
+ * @license 	Apache-2.0
+ *
+ */
+
+package rfc8693
+
+import (
+	"context"
+	"github.com/ory/fosite/handler/oauth2"
+	"github.com/ory/fosite/storage"
```
mitar, line 27: Nitpick, the order of imports here is not good.

pike1212:
What is the status of this feature?

aeneasr:
I believe @mpnunes is waiting for approval of legal in order to be able to sign the CLA

mpnunes:
@aeneasr that is correct.

mpnunes:
> I believe @mpnunes is waiting for approval of legal in order to be able to sign the CLA

That is correct

aeneasr:
I was informed that @mpnunes's company is not able to sign the SLA which essentially means that we can not merge these changes into the upstream :/

saxenautkarsh:
@aeneasr We ended up implementing this rfc on our own for the use case. I am thinking of implementing it here too.
The issue is already there, it is okay if I create a PR directly?

aeneasr:
Thank you, that's great! Does the PR follow the discussion we had on this PR? I think there were some valuable insights :)

If you make your PR, can you please make it against this branch: https://github.com/ory/fosite/pull/667

saxenautkarsh:
@aeneasr Yes, I will follow the discussions in this PR. Also, our implementation has some things specific to our use case so I will change them to make it more generic.

> If you make your PR, can you please make it against this branch: https://github.com/ory/fosite/pull/667

sure!

PR #518, opened 2020-10-24
feat: Support passing repeated audience parameter

### LINKED ISSUE #504: audience parameter splitting by space
**Is your feature request related to a problem? Please describe.**

Currently multiple audience values are obtained by:

```
audience := stringsx.Splitx(request.Form.Get("audience"), " ")
```

I think audiences are generally URIs, so I find it surprising that they are simply passed as a space-delimited string. I can understand that spec is like that for `scope` and `grant_type`, but I think there is no spec for `audience` and this is a custom parameter?

The issue here is that what does one do if URI contains a query parameter with a space? I know this might not be realistic, but just makes me think. So I do not have here a real problem I encountered, except that I simply assumed I have to list multiple audiences multiple times, and not with a space in between.

**Describe the solution you'd like**

I think fosite should support also passing audience as multiple parameters. golang already parses this correctly and you can obtain an array of them through `request.Form["audience"]`. Because I think space is rare in the URI we could keep that for backwards compatibility and simply process all values for the `audience` query parameter and split them by space as well.

I also think it should be documented somewhere how you pass this parameter. I have not found it documented anywhere. And the fact that it should be space separated (and if this issue is resolved, also that you can pass it by repeating the parameter).

**Describe alternatives you've considered**

We can leave it as is, but then we should at least document it better.

mitar:
## Related issue

Fixes #504.

## Proposed changes

Added `GetAudiences` helper function which tries to have current behavior and also support multiple/repeated audience parameters. If there are parameter is repeated, then it is not split by space. If there is only one then it is split by space. I think this is the best balance between standard/backwards behavior and allowing repeated parameter and allowing also URIs/audiences with spaces in them (which we probably all agree is probably not something anyone should be doing).

Also added `ExactAudienceMatchingStrategy` which is slightly more suitable to use for audiences which are not URIs. In [OIDC spec](https://openid.net/specs/openid-connect-core-1_0.html) audience is described as:

> Audience(s) that this ID Token is intended for. It MUST contain the OAuth 2.0 client_id of the Relying Party as an audience value. It MAY also contain identifiers for other audiences. In the general case, the aud value is an array of case sensitive strings. In the common special case when there is one audience, the aud value MAY be a single case sensitive string. 

`client_id` is generally not an URI, but some UUID or some other random string.

## Checklist

      the changes.

### DIFF
--- a/audience_strategy.go
+++ b/audience_strategy.go
@@ -46,8 +46,50 @@ func DefaultAudienceMatchingStrategy(haystack []string, needle []string) error {
 	return nil
 }
 
+// ExactAudienceMatchingStrategy does not assume that audiences are URIs, but compares strings as-is and
+// does matching with exact string comparison. It requires that all strings in "needle" are present in
+// "haystack". Use this strategy when your audience values are not URIs (e.g., you use client IDs for
+// audience and they are UUIDs or random strings).
+func ExactAudienceMatchingStrategy(haystack []string, needle []string) error {
+	if len(needle) == 0 {
+		return nil
+	}
+
+	for _, n := range needle {
+		var found bool
+		for _, h := range haystack {
+			if n == h {
+				found = true
+			}
+		}
+
+		if !found {
+			return errors.WithStack(ErrInvalidRequest.WithHintf(`Requested audience "%s" has not been whitelisted by the OAuth 2.0 Client.`, n))
+		}
+	}
+
+	return nil
+}
+
+// GetAudiences allows audiences to be provided as repeated "audience" form parameter,
+// or as a space-delimited "audience" form parameter if it is not repeated.
+// RFC 8693 in section 2.1 specifies that multiple audience values should be multiple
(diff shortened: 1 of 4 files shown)


audience_strategy.go
```diff
@@ -46,8 +46,46 @@ func DefaultAudienceMatchingStrategy(haystack []string, needle []string) error {
 	return nil
 }
 
+func ExactAudienceMatchingStrategy(haystack []string, needle []string) error {
```
aeneasr, line 49: Maybe add a small codedoc to explain why we need this and when it will be used? I think you could use the text from the PR description and maybe add one example when this would be neccessary.

aeneasr (commented):
Looking very good!

audience_strategy.go
mitar: Added.

mitar:
I addressed review comments.

aeneasr (approved):
Good job @mitar , love the contributions! :)

Issue #445, opened 2020-06-18
OAuth2StatelessJWTIntrospectionFactory cannot be used with standard JWT


mitar:
I use the following strategy:

```go
	strategy := compose.CommonStrategy{
		CoreStrategy: compose.NewOAuth2JWTStrategy(
			config.PrivateKey,
			compose.NewOAuth2HMACStrategy(fositeConfig, config.OAuthSecret, nil),
		),
		OpenIDConnectTokenStrategy: compose.NewOpenIDConnectStrategy(fositeConfig, config.PrivateKey),
	}

```

And the following list of factories:

```go
		compose.OAuth2AuthorizeExplicitFactory,
		compose.OAuth2AuthorizeImplicitFactory,
		compose.OAuth2ClientCredentialsGrantFactory,
		compose.OAuth2RefreshTokenGrantFactory,
		compose.OAuth2TokenRevocationFactory,
		compose.OAuth2TokenIntrospectionFactory,

		compose.OpenIDConnectExplicitFactory,
		compose.OpenIDConnectImplicitFactory,
		compose.OpenIDConnectHybridFactory,
		compose.OpenIDConnectRefreshFactory,

		compose.OAuth2PKCEFactory,
```

I would assume that I could just replace `OAuth2TokenRevocationFactory` and `OAuth2TokenIntrospectionFactory` with `OAuth2StatelessJWTIntrospectionFactory`, but this does not seem to be really possible. I get the following error when running it:

```
panic: interface conversion: compose.CommonStrategy is not oauth2.JWTAccessTokenStrategy: missing method ValidateJWT [recovered]
	panic: interface conversion: compose.CommonStrategy is not oauth2.JWTAccessTokenStrategy: missing method ValidateJWT
```

**Expected behavior**

It just works. And instead of introspection looking in the database, it just looks into the token.

**Environment**

- Version: v0.32.1
- Environment: Ubuntu

aeneasr:
Does [this](https://github.com/ory/hydra/blob/master/driver/registry_base.go#L252-L299) help?

lazeratops:
I am seeing the same issue - the link you pasted does not seem to use the stateless introspection factory, so while I have experimented with implementing similar strategies shown there I still run into the same problem when composing with the CommonStrategy as described above: `panic: interface conversion: compose.CommonStrategy is not oauth2.JWTAccessTokenStrategy: missing method ValidateJWT
`

mitar:
[...]

Issue #441, opened 2020-05-29
Support for adding additional fields to introspection result


mitar:
**Is your feature request related to a problem? Please describe.**

So with JWT access token strategy I can add additional custom claims. Similarly, I can add additional custom claims to ID tokens. But if I use HMAC access tokens, it seems the logical way would be that access token is opaque, but then one can use introspection endpoint to access/obtain those additional custom claims. It seems this is not possible.

**Describe the solution you'd like**

Maybe custom claims from the session should be added to the introspection output?

aeneasr:
So we do this in Hydra and I think it makes sense to have it natively in Fosite as well!

Issue #410, opened 2020-04-02
Support introspection of ID tokens


mitar:
**Is your feature request related to a problem? Please describe.**

It looks like introspection of ID tokens does not work through the introspect handler?

**Describe the solution you'd like**

That it does.

**Describe alternatives you've considered**

I could validate ID tokens locally by myself, but it is also useful to have an API endpoint to do so.

**Additional context**

Okta does so: https://developer.okta.com/docs/reference/api/oidc/

aeneasr:
Is there an open standard to this? I don't quite remember but I think some OpenID Connect spec was covering this. Maybe it would make sense to go down that road? The okta solution seems to be "proprietary" to their API.

mitar:
So [rfc7662](https://tools.ietf.org/html/rfc7662) just covers access and refresh tokens, others are out of the scope. But it is dealing only with OAuth anyway. I could not find anything about ID token introspection in OpenID Connect specs. But that is a bit strange, given that OIDC adds a new token type. So not sure what to do about this.

aeneasr:
The rationale is that there is no need to do introspection if the token is not opaque and you have access to the keys. This seems to be really dependent on the provider, Google has the tokeninfo endpoint for that which doesn't follow any RFC or OIDC convention.

mitar:
For my use case, it is more about validation than introspection. One can always decode it, but to check if token was not revoked, you have to call introspection endpoint, no? So it is about trusting the authorization server to do the validation instead of you doing it locally. For some use cases where you do not care so much about extra round trip time, this might be OK.

See [also this spec](https://tools.ietf.org/html/draft-ietf-oauth-jwt-introspection-response-08). This is not exactly the same thing as we are discussing here, but has some of similar rationale for it:

> However, there are use cases where the resource server requires stronger assurance that the authorization server issued the token introspection response for an access token, including cases where the authorization server assumes liability for the content of the token introspection response.

This is similar to what I am thinking. I want the server to tell me that everything is OK. That it is not the client who has the clock wrong, or that the token was revoked.

Is there any other endpoint to check if an ID token has been revoked?

aeneasr:
> For my use case, it is more about validation than introspection. One can always decode it, but to check if token was not revoked, you have to call introspection endpoint, no? So it is about trusting the authorization server to do the validation instead of you doing it locally. For some use cases where you do not care so much about extra round trip time, this might be OK.
>
> Is there any other endpoint to check if an ID token has been revoked?

[...]

PR #519, opened 2020-10-24
fix: Modernized JWT stateless introspection

mitar:
## Related issue

Fixes: https://github.com/ory/fosite/issues/445

## Proposed changes

This makes it so that you can:

```
compose.OAuth2TokenRevocationFactory,
compose.OAuth2TokenIntrospectionFactory,
```

replace with:

```
compose.OAuth2StatelessJWTIntrospectionFactory,
```

and it just works. From looking at the code it looked to me like this was done long time ago where some other current interfaces were not yet around. So I modernized the code to use existing interfaces and removed a specialized interface used just by previous implementation.

All tests still pass. But I am not 100% sure if everything is correct (I have some doubts about original code). I also have few TODOs around the code so input there would be appreciated.

## Checklist

      the changes.

### DIFF
--- a/handler/oauth2/introspector_jwt.go
+++ b/handler/oauth2/introspector_jwt.go
@@ -36,6 +36,56 @@ type StatelessJWTValidator struct {
 	ScopeStrategy fosite.ScopeStrategy
 }
 
+// AccessTokenJWTToRequest tries to reconstruct fosite.Request from a JWT.
+func AccessTokenJWTToRequest(token *jwtx.Token) fosite.Requester {
+	mapClaims := token.Claims.(jwtx.MapClaims)
+	claims := jwt.JWTClaims{}
+	claims.FromMapClaims(mapClaims)
+
+	requestedAt := claims.IssuedAt
+	requestedAtClaim, ok := mapClaims["rat"]
+	if ok {
+		switch requestedAtClaim.(type) {
+		case float64:
+			requestedAt = time.Unix(int64(requestedAtClaim.(float64)), 0).UTC()
+		case int64:
+			requestedAt = time.Unix(requestedAtClaim.(int64), 0).UTC()
+		}
+	}
+
+	clientId := ""
+	clientIdClaim, ok := mapClaims["client_id"]
+	if ok {
+		switch clientIdClaim.(type) {
+		case string:
+			clientId = clientIdClaim.(string)
+		}
+	}
+
+	return &fosite.Request{
+		RequestedAt: requestedAt,
+		Client: &fosite.DefaultClient{
+			ID: clientId,
+		},
+		// We do not really know which scopes were requested, so we set them to granted.
+		RequestedScope: claims.Scope,
+		GrantedScope:   claims.Scope,
+		Session: &JWTSession{
+			JWTClaims: &claims,
+			JWTHeader: &jwt.Headers{
+				Extra: token.Header,
+			},
+			ExpiresAt: map[fosite.TokenType]time.Time{
+				fosite.AccessToken: claims.ExpiresAt,
+			},
+			Subject: claims.Subject,
+		},
+		// We do not really know which audiences were requested, so we set them to granted.
+		RequestedAudience: claims.Audience,
+		GrantedAudience:   claims.Audience,
+	}
+}
+
 func (v *StatelessJWTValidator) IntrospectToken(ctx context.Context, token string, tokenUse fosite.TokenUse, accessRequest fosite.AccessRequester, scopes []string) (fosite.TokenUse, error) {
 	t, err := validate(ctx, v.JWTStrategy, token)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


handler/oauth2/introspector_jwt.go
```diff
@@ -23,38 +23,74 @@ package oauth2
...
+	t, err := validate(ctx, v.JWTStrategy, token)
 	if err != nil {
 		return "", err
 	}
 
-	for _, scope := range scopes {
-		if scope == "" {
-			continue
-		}
+	// TODO: From here we assume it is an access token, but how do we know it is really and that is not an ID token?
```
aeneasr, line 45: Hm yeah good question. I think we generally assume that ID and Access Tokens use different private keys but that check is obviously implicit. I mean we could add the `token_type` to the JWT and set it to `Bearer` (ID Tokens are not HTTP Authorization Bearer Tokens) to check this?

handler/oauth2/introspector_jwt.go
```diff
@@ -23,38 +23,74 @@ package oauth2
...
+			requestedAt = time.Unix(rat.(int64), 0).UTC()
 		}
 	}
 
-	accessRequest.Merge(or)
+	requester := &fosite.Request{
+		ID:          accessRequest.GetID(),
+		RequestedAt: requestedAt,
+		// TODO: Should this client be the client which requested the introspection or the client which obtained the JWT?
+		Client: accessRequest.GetClient(),
```
aeneasr, line 71: The client in the JWT!

handler/oauth2/introspector_jwt.go
```diff
@@ -23,38 +23,74 @@ package oauth2
...
+		// We do not really know which scopes were requested, so we set them to granted.
+		RequestedScope: claims.Scope,
+		GrantedScope:   claims.Scope,
+		Form:           accessRequest.GetRequestForm(),
+		// TODO: Is it OK that this is JWTSession or should we allow custom session objects?
+		Session: &JWTSession{
```
aeneasr, line 77: The session should be the one coming from the store with all the extra information that was set when establishing the session. The introspection basically relays 1:1 the "access token session" and all its affiliated data.

handler/oauth2/introspector_jwt.go
```diff
@@ -23,38 +23,74 @@ package oauth2
...
+			JWTClaims: &claims,
+			JWTHeader: &jwt.Headers{
+				Extra: t.Header,
+			},
+			ExpiresAt: map[fosite.TokenType]time.Time{
+				fosite.AccessToken: claims.ExpiresAt,
+			},
+			Subject: claims.Subject,
+			// TODO: What about username?
```
aeneasr, line 86: We could explicitly support it, or, implicitly support it using the session object. I think one problem though is that the session is set under a key `ext` and can not be "merged" into the root claims - iirc 😅

handler/oauth2/introspector_jwt.go
mitar: > I think we generally assume that ID and Access Tokens use different private keys but that check is obviously implicit.

Ohh. Is this documented anywhere? Because this is the first time I hear this. I was passing same private key to both `NewOAuth2JWTECDSAStrategy` and `NewOpenIDConnectECDSAStrategy`. It [does not help that compose does it as well](https://github.com/ory/fosite/blob/master/compose/compose.go#L101-L104).

> I mean we could add the token_type to the JWT and set it to Bearer

Is this claim standardized anywhere? I can see it some implementations do it, but I could not find a spec?

I see that currently ID tokens have `auth_time` and access tokens do not have it. And access tokens have `scope` and ID tokens do not have it (not sure if any spec controls this). I also have `client_id` in access tokens in my implementation. In my implementation `scope` is always required but I think fosite does not enforce that.

Maybe we could just make it that `scope` claim is always present in access tokens (even if empty string) and never in ID tokens and use that?

mitar: Personally, I would suggest that:
* Fosite introspection endpoint supports both ID tokens and access tokens (https://github.com/ory/fosite/issues/410).
  * Then here we do not have to care which one it is, we just return all claims.
* Caller of `introspect` has to check `aud` claim to verify that the token contains expected audience (which is in fact the main way to differentiate between ID tokens and access tokens, because they have different audience).
  * To make this easier, we add `audience` parameter to introspection endpoint, for caller to be able to specify expected audiences. Similar to how we currently have `scope` parameter already (both would then be a fosite extension, I have not found any spec for `scope` parameter).

handler/oauth2/introspector_jwt.go
mitar: Stateless JWT introspection does not have the store. So we are reconstructing it to the best possible extent from the JWT. Does it then matter if we use `JWTSession` here? I do not think so. Because the only use here I think is that we want to pass claims from here all the way to the output. Is this session really used for anything else?

handler/oauth2/introspector_jwt.go
mitar: What I meant here is that username does not seem to be encoded in the access token JWT by default (subject is). So it is not really possible to retrieve it from the JWT. Also, I have not seen a standard claim which I could use here either.

handler/oauth2/introspector_jwt.go
mitar: I have thought a bit about this and I think it is OK as-is. I do not see use case where this should be customizable.

handler/oauth2/introspector_jwt.go
mitar: Fixed.

mitar:
I made a new implementation of this. I think it addressed all of my TODOs. I think it reconstructs request as much as possible.
I think we should do https://github.com/ory/fosite/issues/410 which addresses the issue of ID token vs. access token as well. Also, when we do https://github.com/ory/fosite/issues/441 it will also address the question of username, because that will simply be mapped through the claims map, however it was encoded originally.

aeneasr (approved):
This looks good from my side! Anything left that need discussion/documentation or can I merge this?

mitar:
I think it is good for merging from my side. I think you should tell me if I can implement #410 which I would also like, even if it is not by the spec.

handler/oauth2/introspector_jwt.go
lazeratops: I actually did not know ID and access tokens are assumed to use different private keys. I believe we have been using the same key for ID and access tokens on our end. I think I need to do some reading up on this. Does anyone know if this is this mentioned somewhere in the OIDC RFC?

aeneasr: They are fundamentally different tokens (proof of authentication, proof of authorization). They should get different keys to prevent mixing them together.

mitar: But how you communicate to consumers that they are using different tokens? Because there is only one `/keys` endpoint where you have to publish both keys? So how consumers know which key is to be used for which tokens?

aeneasr: The access token keys should not exposed for external consumption IMO.

mitar: For JWTs of course they are? That is the whole point that you can validate JWT based access tokens without having to contact introspection endpoint.

aeneasr: Introspection is internal facing, not external.

aeneasr: >    This specification defines a method for a protected resource to query
   an OAuth 2.0 authorization server to determine the active state of an
   OAuth 2.0 token and to determine meta-information about this token.
   OAuth 2.0 deployments can use this method to convey information about
   the authorization context of the token from the authorization server
   to the protected resource.

mitar: I mean, it is a protected resource. But that does not mean it has to be internal facing? It just has to be protected to prevent scanning attacks.

PR #520, opened 2020-10-28
feat: Allow extra fields in introspect response

### LINKED ISSUE #441: Support for adding additional fields to introspection result

mitar:
## Related issue

Fixes #441.

## Proposed changes

Sessions can now implement `GetExtraClaims` to control extra claims in token's introspect output.

## Checklist

      the changes.

## Further comments

I am not completely satisfied with tests here. For example, there is no test which would check which all fields are really output from introspect for regular access token and for JWT access token. Any suggestion where to add it is welcome.

### DIFF
--- a/handler/oauth2/strategy_jwt_session.go
+++ b/handler/oauth2/strategy_jwt_session.go
@@ -103,3 +103,14 @@ func (s *JWTSession) Clone() fosite.Session {
 
 	return deepcopy.Copy(s).(fosite.Session)
 }
+
+// GetExtraClaims implements ExtraClaimsSession for JWTSession.
+// The returned value is a copy of JWTSession claims.
+func (s *JWTSession) GetExtraClaims() map[string]interface{} {
+	if s == nil {
+		return nil
+	}
+
+	// We make a clone so that WithScopeField does not change the original value.
+	return s.Clone().(*JWTSession).GetJWTClaims().WithScopeField(jwt.JWTScopeFieldString).ToMapClaims()
+}
--- a/handler/oauth2/strategy_jwt_test.go
+++ b/handler/oauth2/strategy_jwt_test.go
@@ -55,7 +55,6 @@ var jwtValidCase = func(tokenType fosite.TokenType) *fosite.Request {
 			JWTClaims: &jwt.JWTClaims{
 				Issuer:    "fosite",
 				Subject:   "peter",
-				Audience:  []string{"group0"},
 				IssuedAt:  time.Now().UTC(),
 				NotBefore: time.Now().UTC(),
 				Extra:     map[string]interface{}{"foo": "bar"},
@@ -71,6 +70,8 @@ var jwtValidCase = func(tokenType fosite.TokenType) *fosite.Request {
 	r.SetRequestedScopes([]string{"email", "offline"})
 	r.GrantScope("email")
 	r.GrantScope("offline")
+	r.SetRequestedAudience([]string{"group0"})
+	r.GrantAudience("group0")
 	return r
 }
 
@@ -83,7 +84,6 @@ var jwtValidCaseWithZeroRefreshExpiry = func(tokenType fosite.TokenType) *fosite
 			JWTClaims: &jwt.JWTClaims{
 				Issuer:    "fosite",
 				Subject:   "peter",
-				Audience:  []string{"group0"},
 				IssuedAt:  time.Now().UTC(),
 				NotBefore: time.Now().UTC(),
 				Extra:     map[string]interface{}{"foo": "bar"},
@@ -100,6 +100,8 @@ var jwtValidCaseWithZeroRefreshExpiry = func(tokenType fosite.TokenType) *fosite
 	r.SetRequestedScopes([]string{"email", "offline"})
 	r.GrantScope("email")
 	r.GrantScope("offline")
+	r.SetRequestedAudience([]string{"group0"})
+	r.GrantAudience("group0")
 	return r
 }
 
@@ -112,7 +114,6 @@ var jwtValidCaseWithRefreshExpiry = func(tokenType fosite.TokenType) *fosite.Req
 			JWTClaims: &jwt.JWTClaims{
 				Issuer:    "fosite",
 				Subject:   "peter",
-				Audience:  []string{"group0"},
 				IssuedAt:  time.Now().UTC(),
 				NotBefore: time.Now().UTC(),
 				Extra:     map[string]interface{}{"foo": "bar"},
@@ -129,6 +130,8 @@ var jwtValidCaseWithRefreshExpiry = func(tokenType fosite.TokenType) *fosite.Req
 	r.SetRequestedScopes([]string{"email", "offline"})
 	r.GrantScope("email")
 	r.GrantScope("offline")
+	r.SetRequestedAudience([]string{"group0"})
+	r.GrantAudience("group0")
 	return r
 }
 
@@ -144,11 +147,10 @@ var jwtExpiredCase = func(tokenType fosite.TokenType) *fosite.Request {
 			JWTClaims: &jwt.JWTClaims{
 				Issuer:    "fosite",
 				Subject:   "peter",
-				Audience:  []string{"group0"},
 				IssuedAt:  time.Now().UTC(),
 				NotBefore: time.Now().UTC(),
 				ExpiresAt: time.Now().UTC().Add(-time.Minute),
-				Extra:     make(map[string]interface{}),
+				Extra:     map[string]interface{}{"foo": "bar"},
 			},
 			JWTHeader: &jwt.Headers{
 				Extra: make(map[string]interface{}),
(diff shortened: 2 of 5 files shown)


introspection_response_writer.go
```diff
@@ -204,35 +205,46 @@ func (f *Fosite) WriteIntrospectionResponse(rw http.ResponseWriter, r Introspect
...
+	response := map[string]interface{}{
+		"active": true,
+	}
+
+	extraClaimsSession, ok := r.GetAccessRequester().GetSession().(ExtraClaimsSession)
+	if ok {
+		extraClaims := extraClaimsSession.GetExtraClaims()
+		if extraClaims != nil {
+			for name, value := range extraClaims {
+				if !reflect.ValueOf(value).IsZero() {
```
aeneasr, line 217: If I am not mistaken, this would remove zero values. But what if we have a parameter which is e.g. an enum and has a valid number of `0`? Maybe we could check for not nil instead?

introspection_response_writer.go
```diff
@@ -204,35 +205,46 @@ func (f *Fosite) WriteIntrospectionResponse(rw http.ResponseWriter, r Introspect
...
+		response["scope"] = strings.Join(r.GetAccessRequester().GetGrantedScopes(), " ")
+	}
+	if !r.GetAccessRequester().GetRequestedAt().IsZero() {
+		response["iat"] = r.GetAccessRequester().GetRequestedAt().Unix()
+	}
+	if r.GetAccessRequester().GetSession().GetSubject() != "" {
+		response["sub"] = r.GetAccessRequester().GetSession().GetSubject()
+	}
+	if len(r.GetAccessRequester().GetGrantedAudience()) > 0 {
+		response["aud"] = r.GetAccessRequester().GetGrantedAudience()
```
aeneasr, line 240: In my opinion, `GetSession()` should not be able to overwrite these standard claims. It would allow developers to really mess with the logic by abusing the session and it could lead to security implications around subject values, expiry times, and more. What do you think?

introspection_response_writer_test.go
```diff
@@ -117,8 +120,23 @@ func TestWriteIntrospectionResponseBody(t *testing.T) {
 				sess.SetExpiresAt(ires.TokenUse, time.Time{})
 				ires.AccessRequester = NewAccessRequest(sess)
 			},
-			active: true,
-			hasExp: false,
+			active:   true,
+			hasExp:   false,
+			hasExtra: false,
+		},
+		{
```
aeneasr, line 127: Could you please add a test which checks that the standard claims are not overwritten by the session? So things like `sub`, `aud`, `expires_at`?

aeneasr (changes requested):
Thank you! I've added a few comments with ideas how to improve the patch :)

introspection_response_writer.go
mitar: So this is why I first populate it from `GetExtraClaims` and then from standard API methods. So not sure why you are saying that session can override standard claims?

introspection_response_writer.go
mitar: I think enumerations will not encode to JSON well anyway, if they are numeric.

We can maybe check for empty strings and nil only.

We can also remove this check and be more careful what we get to here in the first place. This is why I changed `ToMap` functions to not add claims if values are zero.

aeneasr: I meant enums of type int, so `0,1,2,3`. `acr_values` is such an enum which has a `0` value that must be set if requested.

introspection_response_writer.go
aeneasr: Because the code skips those values if they are not set. This can lead to hacks where people try to deliberately null these to set something else. The session should not be able to ever change these values, regardless of the actual value.

introspection_response_writer.go
mitar: So yes, which of two options I listed you would prefer?

aeneasr: I think removing the check is legit as long as we add an option to `ExtraClaims` to also delete a value! Otherwise I would check for `nil`.

introspection_response_writer.go
mitar: OK, I can filter them out.

introspection_response_writer.go
mitar: > to also delete a value!

What do you mean? Why would you be deleting a value? We just now established that `GetExtraClaims` should be allowed just to add non-standard claims. But then who else would be setting it to make deletion useful? You simply should not return value from `GetExtraClaims` if you do not want to set it. And we should make sure default implementation does not return such values.

aeneasr: > What do you mean? Why would you be deleting a value

If your session pipeline is complex :) But maybe I'm getting ahead of myself. Let's just use all the values from the map and not do any check.

mitar:
@aeneasr I think I addressed all your comments.

handler/oauth2/strategy_jwt_session.go
```diff
@@ -103,3 +103,14 @@ func (s *JWTSession) Clone() fosite.Session {
...
+
+// GetExtraClaims implements ExtraClaimsSession for JWTSession.
+// The returned value is a copy of JWTSession claims.
+func (s *JWTSession) GetExtraClaims() map[string]interface{} {
+	if s == nil {
+		return nil
+	}
+
+	// We make a clone so that WithScopeField does not change the original value.
+	return s.Clone().(*JWTSession).GetJWTClaims().WithScopeField(jwt.JWTScopeFieldString).ToMapClaims()
```
aeneasr, line 115: Why is the scope field hardcoded here?

handler/oauth2/strategy_jwt_test.go
```diff
@@ -55,7 +55,6 @@ var jwtValidCase = func(tokenType fosite.TokenType) *fosite.Request {
 			JWTClaims: &jwt.JWTClaims{
 				Issuer:    "fosite",
 				Subject:   "peter",
-				Audience:  []string{"group0"},
```
aeneasr, line 58: Would this still work though? If so can we add a test to cover this case?

introspection_response_writer.go
```diff
@@ -204,35 +204,50 @@ func (f *Fosite) WriteIntrospectionResponse(rw http.ResponseWriter, r Introspect
...
+	}
+
+	extraClaimsSession, ok := r.GetAccessRequester().GetSession().(ExtraClaimsSession)
+	if ok {
+		extraClaims := extraClaimsSession.GetExtraClaims()
+		if extraClaims != nil {
+			for name, value := range extraClaims {
+				switch name {
+				// We do not allow these to be set through extra claims.
+				case "exp", "client_id", "scope", "iat", "sub", "aud", "username":
```
aeneasr, line 218: Must include `iss`, `jti`, `nbf`

introspection_response_writer.go
```diff
@@ -204,35 +204,50 @@ func (f *Fosite) WriteIntrospectionResponse(rw http.ResponseWriter, r Introspect
...
+					break
```
aeneasr, line 219: Shouldn't this be a continue instead of a break? Or even better an error so we know what is happening and don't have to debug a long time to find this place.

aeneasr (changes requested):
Sorry for the long wait time, but it is finally possible to run oidc conformity tests in fosite which I wanted to address before continuing review here.

I added a few more comments which I think need to be addressed before merge!

handler/oauth2/strategy_jwt_session.go
mitar: Because in fact the only standard approach for scope field is to do `JWTScopeFieldString` and not `JWTScopeFieldList`. The only reason why `JWTScopeFieldList` exists is for backwards compatibility. Because we are adding new output here, I do not think we have to care about backwards compatibility. Thus I just picked the standard behavior.

introspection_response_writer.go
mitar: You are right. `continue` should be here.

We cannot return an error here because introspect endpoint does not really return any errors.

introspection_response_writer.go
mitar: This is not really possible to limit. Because currently regular claims from JWT tokens are passed through to introspection endpoint using `GetExtraClaims`. I tried to not do it like that, but I get import cycle. See: https://github.com/ory/fosite/pull/579/commits/04cd6255188941af972552da984f65f9cf6cde7e

mitar:
I made a followup PR: https://github.com/ory/fosite/pull/579

handler/oauth2/strategy_jwt_test.go
mitar: No, it cannot work because we are filtering out `aud` from `GetExtraClaims`. Before the test was in fact buggy, I would claim, and didn't really check if audience is set in the generated (and tested) token. So setting this value does nothing.

PR #522, opened 2020-10-28
refactor: Generate claims in the same way

mitar:
## Proposed changes

Just a bit of refactoring/code cleanup. Made sure that both OAuth JWT and OpenID JWT claims are generated in the same way. Also made order of field processing in the code the same so it is easier to compare.

## Checklist

      the changes.

### DIFF
--- a/token/jwt/claims_id_token.go
+++ b/token/jwt/claims_id_token.go
@@ -49,14 +49,37 @@ type IDTokenClaims struct {
 // ToMap will transform the headers to a map structure
 func (c *IDTokenClaims) ToMap() map[string]interface{} {
 	var ret = Copy(c.Extra)
-	ret["sub"] = c.Subject
-	ret["iss"] = c.Issuer
-	ret["jti"] = c.JTI
+
+	if c.Subject != "" {
+		ret["sub"] = c.Subject
+	}
+
+	if c.Issuer != "" {
+		ret["iss"] = c.Issuer
+	}
+
+	if c.JTI != "" {
+		ret["jti"] = c.JTI
+	} else {
+		ret["jti"] = uuid.New()
+	}
 
 	if len(c.Audience) > 0 {
 		ret["aud"] = c.Audience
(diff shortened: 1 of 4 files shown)


token/jwt/claims_id_token.go
```diff
@@ -87,9 +106,6 @@ func (c *IDTokenClaims) ToMap() map[string]interface{} {
 		ret["amr"] = c.AuthenticationMethodsReference
 	}
 
-	ret["iat"] = float64(c.IssuedAt.Unix())
```
mitar, line 90: So `iat` in ID tokens was always added but in access tokens only if set.

I made it so that all fields are not set if empty. I think this is cleaner.

token/jwt/claims_jwt_test.go
```diff
@@ -75,7 +75,7 @@ func TestAssert(t *testing.T) {
 		ToMapClaims().Valid())
 	assert.NotNil(t, (&JWTClaims{NotBefore: time.Now().UTC().Add(time.Hour)}).
 		ToMapClaims().Valid())
-	assert.NotNil(t, (&JWTClaims{NotBefore: time.Now().UTC().Add(-time.Hour)}).
+	assert.Nil(t, (&JWTClaims{NotBefore: time.Now().UTC().Add(-time.Hour)}).
```
mitar, line 78: This changes because `ExpiredAt` is not 0 anymore.

token/jwt/claims_id_token.go
```diff
@@ -49,16 +49,39 @@ type IDTokenClaims struct {
 // ToMap will transform the headers to a map structure
```
aeneasr, line 49: I have similar concerns about this as [here](https://github.com/ory/fosite/pull/520#discussion_r513301990).

aeneasr:
How should we proceed with this PR?

mitar:
I will get back to this one shortly.

PR #523, opened 2020-10-28
fix: Resetting more ID token claims on refresh

mitar:
## Proposed changes

I found this during code review. I am not sure if it is correct. I could not find any test testing this. Even if I remove existing lines resetting `Nonce` and `ExpiresAt` nothing fails. So where is a test which would make sure that refreshed tokens do not have same JIT?

## Checklist

      the changes.


aeneasr (approved):

aeneasr:
Yes we definitely should add a test covering this. I would expect this to be covered in the [refresh integration tests](https://github.com/ory/fosite/blob/master/integration/refresh_token_grant_test.go).

Issue #452, opened 2020-07-22
Do not require that key ID is provided with the client's public key


mitar:
**Is your feature request related to a problem? Please describe.**

It seems clients are bad at managing their keys and making sure key IDs are unique for keys they provide. Also sometimes when they generate a key it simply does not contain the key ID. So they have to manually add it and stuff.

But I think key ID in client's public key should not even be required. RFC says that [`kid` header parameter is optional](https://tools.ietf.org/html/rfc7515#section-4.1.4), [key identification section](https://tools.ietf.org/html/rfc7515#section-6) also talks that any way of figuring out the key can be used, and [appendix D](https://tools.ietf.org/html/rfc7515#appendix-D) explicitly mentions simply trying all possibly viable keys.

**Describe the solution you'd like**

I think `kid` should be optional. It is just an optimization. We could go over all keys client has and try them in order. If `kid` is provided, we can make that faster. But going over all keys should be enough.

Or at least, if there is only one key in the list of client's keys, then the `kid` should be optional. We should just return that one key.

**Describe alternatives you've considered**

There are of course so many other fields we could be selecting keys on, like `x5t`. So making `kid` required is really a very opinionated decision. It could even be `x5t` by which you select the key. So if you want to keep that we do not traverse all keys client has configured (which I doubt is expensive because those lists tend to be short), we could ideally at least use many more mechanisms to match the one key (`kid`, having only one key in the list, `x5t`, and so on). I think this is a bit of an overkill to support all those and I would leave to others to implement that if somebody will want. But making `kid` optional is something I would implement if you ACK it.

There is a bit of an issue that that the [library currently used](https://github.com/dgrijalva/jwt-go) requires only one key to be provided for validation. I opened https://github.com/dgrijalva/jwt-go/issues/416 to address this. Because of this I think that for now the most reasonable way is to allow `kid` to not be provided if there is only one key listed for the client. If so, just return that key if `kid` is not provided.

aeneasr:
Checking against possibly hundreds of RSA keypairs every couple of seconds (assuming medium scale) is a massive CPU drain. We had this problem in ORY Oathkeeper and had to solve it with caching even though we knew the `kid`.

mitar:
So you are saying one particular client would have possibly hundreds of RSA keypairs? So we already know which client it is, we just have to check inside keys of that client which key it is. And I really do not think it is too expensive to check if there is only one key for that user and just returning it, if `kid` is not provided?

aeneasr:
[...]

PR #526, opened 2020-11-04
fix: only use allowed characters in error_description

### LINKED ISSUE #525: Non-conformant error responses
**Describe the bug**

We recently changed the way `error_description` and other error fields are generated. This unfortunately caused two warnings and one failure in the OpenID Connect Conformity Test Suite, specifically the  `oidcc-response-type-missing` test.

**To Reproduce**

The full test log can be seen here: [test-log-oidcc-response-type-missing-client_secret_basic-discovery-code-default-dynamic_client-8XprPDGkiQ09O8v.zip](https://github.com/ory/fosite/files/5486535/test-log-oidcc-response-type-missing-client_secret_basic-discovery-code-default-dynamic_client-8XprPDGkiQ09O8v.zip)

**WARNING: `error_hint` is unexpected**

*I think this can be ignored*

> error response includes unexpected parameters. This may be because the server supports extensions the test suite is unaware of, or the server may be returning values it should not.

**WARNING: `error_hint` is unexpected**

*I think this can be ignored*

> error response includes unexpected parameters. This may be because the server supports extensions the test suite is unaware of, or the server may be returning values it should not.

**WARNING error_description has CR LF or TAB**

*This should be fixed*

> 'error_description' field includes characters CR, LF or TAB, these are not recommended to include
>
> * see: https://bitbucket.org/openid/connect/issues/1147/certification-rfc6749-must-for
> * error_description: The authorization server does not support obtaining a token using this method The request is missing the "response_type"" parameter.

**ERROR invalid characters in error_description**

> 'error_description' field MUST NOT include characters outside the set %09-0A (Tab and LF) / %x0D (CR) / %x20-21 / %x23-5B / %x5D-7E
> * error_description: The authorization server does not support obtaining a token using this method The request is missing the "response_type"" parameter.

**Expected behavior**

The test should pass.

**Environment**

ORY Fosite `v0.34.1`

**Additional context**

Bildschirmfoto 2020-11-04 um 09 49 20

/cc @mitar could you take a look maybe?

aeneasr:
Replace LF and quotes with `.` and `'` to match allowed and recommended character set defined in various RFCs.

Closes #525

## Checklist

      the changes.

### DIFF
--- a/audience_strategy.go
+++ b/audience_strategy.go
@@ -18,14 +18,14 @@ func DefaultAudienceMatchingStrategy(haystack []string, needle []string) error {
 	for _, n := range needle {
 		nu, err := url.Parse(n)
 		if err != nil {
-			return errors.WithStack(ErrInvalidRequest.WithHintf(`Unable to parse requested audience "%s".`, n).WithCause(err).WithDebug(err.Error()))
+			return errors.WithStack(ErrInvalidRequest.WithHintf("Unable to parse requested audience '%s'.", n).WithCause(err).WithDebug(err.Error()))
 		}
 
 		var found bool
 		for _, h := range haystack {
 			hu, err := url.Parse(h)
 			if err != nil {
-				return errors.WithStack(ErrInvalidRequest.WithHintf(`Unable to parse whitelisted audience "%s".`, h).WithCause(err).WithDebug(err.Error()))
+				return errors.WithStack(ErrInvalidRequest.WithHintf("Unable to parse whitelisted audience '%s'.", h).WithCause(err).WithDebug(err.Error()))
 			}
 
 			allowedPath := strings.TrimRight(hu.Path, "/")
@@ -39,7 +39,7 @@ func DefaultAudienceMatchingStrategy(haystack []string, needle []string) error {
 		}
 
 		if !found {
-			return errors.WithStack(ErrInvalidRequest.WithHintf(`Requested audience "%s" has not been whitelisted by the OAuth 2.0 Client.`, n))
+			return errors.WithStack(ErrInvalidRequest.WithHintf("Requested audience '%s' has not been whitelisted by the OAuth 2.0 Client.", n))
 		}
 	}
 
--- a/authorize_request_handler.go
+++ b/authorize_request_handler.go
@@ -34,6 +34,14 @@ import (
 	"github.com/ory/go-convenience/stringslice"
 )
 
+func wrapSigningKeyFailure(outer *RFC6749Error, inner error) *RFC6749Error {
+	outer = outer.WithCause(inner).WithDebug(inner.Error())
+	if e := new(RFC6749Error); errors.As(inner, &e) {
+		return outer.WithHintf("%s %s", outer.Reason(), e.Reason())
+	}
+	return outer
+}
+
 func (f *Fosite) authorizeRequestParametersFromOpenIDConnectRequest(request *AuthorizeRequest) error {
 	var scope Arguments = RemoveEmpty(strings.Split(request.Form.Get("scope"), " "))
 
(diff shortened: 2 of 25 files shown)


audience_strategy.go
```diff
@@ -18,14 +18,14 @@ func DefaultAudienceMatchingStrategy(haystack []string, needle []string) error {
 	for _, n := range needle {
 		nu, err := url.Parse(n)
 		if err != nil {
-			return errors.WithStack(ErrInvalidRequest.WithHintf(`Unable to parse requested audience "%s".`, n).WithCause(err).WithDebug(err.Error()))
+			return errors.WithStack(ErrInvalidRequest.WithHintf("Unable to parse requested audience '%s'.", n).WithCause(err).WithDebug(err.Error()))
```
mitar, line 21: What if audience contains `"` for some reason?

I think we should also sanitize manually final error description before we output it out.

authorize_request_handler.go
```diff
@@ -104,23 +109,23 @@ func (f *Fosite) authorizeRequestParametersFromOpenIDConnectRequest(request *Aut
 		case *jwt.SigningMethodRSA:
 			key, err := f.findClientPublicJWK(oidcClient, t, true)
 			if err != nil {
-				return nil, errors.WithStack(ErrInvalidRequestObject.WithHint("Unable to retrieve RSA signing key from OAuth 2.0 Client.").WithCause(err).WithDebug(err.Error()))
+				return nil, errors.WithStack(ErrInvalidRequestObject.WithHintf("Unable to retrieve RSA signing key from OAuth 2.0 Client because: %s.", err).WithCause(err).WithDebug(err.Error()))
```
mitar, line 112: No, we should never leak `err` out, just when debugging. I removed this recently in one PR. So `because ...` bit I worry about. (Especially if we will implement custom key search strategies with https://github.com/ory/fosite/issues/452)

client_authentication_jwks_strategy.go
```diff
@@ -60,17 +60,17 @@ func (s *DefaultJWKSFetcherStrategy) Resolve(location string, forceRefresh bool)
 	if !ok || forceRefresh {
 		response, err := s.client.Get(location)
 		if err != nil {
-			return nil, errors.WithStack(ErrServerError.WithHintf(`Unable to fetch JSON Web Keys from location "%s".`, location).WithCause(err).WithDebug(err.Error()))
+			return nil, errors.WithStack(ErrServerError.WithHintf("Unable to fetch JSON Web Keys from location '%s' because: %s.", location, err).WithCause(err).WithDebug(err.Error()))
```
mitar, line 63: Same here. `err` could be sensitive.

client_authentication_jwks_strategy.go
```diff
@@ -60,17 +60,17 @@ func (s *DefaultJWKSFetcherStrategy) Resolve(location string, forceRefresh bool)
...
 
 		if response.StatusCode < 200 || response.StatusCode >= 400 {
-			return nil, errors.WithStack(ErrServerError.WithHintf(`Expected successful status code from location "%s", but received code "%d".`, location, response.StatusCode))
+			return nil, errors.WithStack(ErrServerError.WithHintf("Expected successful status code from location '%s' but received code %d.", location, response.StatusCode))
 		}
 
 		var set jose.JSONWebKeySet
 		if err := json.NewDecoder(response.Body).Decode(&set); err != nil {
-			return nil, errors.WithStack(ErrServerError.WithHintf("Unable to decode JSON Web Keys from location \"%s\".", location).WithCause(err).WithDebug(err.Error()))
+			return nil, errors.WithStack(ErrServerError.WithHintf("Unable to decode JSON Web Keys from location '%s' because: %s.", location, err).WithCause(err).WithDebug(err.Error()))
```
mitar, line 73: And here.

errors.go
```diff
@@ -339,12 +340,12 @@ func (e *RFC6749Error) Sanitize() *RFC6749Error {
...
 	if e.Hint != "" {
-		description += "\n\n" + e.Hint
+		description += " " + e.Hint
 	}
 	if e.Debug != "" {
-		description += "\n\n" + e.Debug
+		description += " " + e.Debug
 	}
-	return description
+	return strings.ReplaceAll(description, "\"", "'")
```
mitar, line 348: Also `\` is not allowed, and anything non-ASCII, so I think we have to sanitize this more.

authorize_request_handler.go
aeneasr: Hm I see your point but the problem is that this is really hard to debug. You would have to turn on debug mode for clients to understand why their request failed (e.g. `kid` does not exist, wrong alg, ...). So we need a way to return the appropriate error to the regular OAuth2 consumer without enabling debug.

errors.go
aeneasr: Right, my idea here was to just remove this for any stray characters but I guess we could santize it even more - but would probably mark this as a feature instead of addressing this here.

audience_strategy.go
aeneasr: https://github.com/ory/fosite/issues/527

errors.go
aeneasr: https://github.com/ory/fosite/issues/527

client_authentication_jwks_strategy.go
aeneasr: Hm I don't really think that's the case. `net/http` is basically just saying DIAL errors on hostnames and IPs. Since the location is given by the client anyways I don't think there's much that could leak here?

aeneasr: But yeah scrap my last comment - you're right. I've addressed this!

authorize_request_handler.go
mitar: I think we should then check for known error messages and map it? Like [it is done here](https://github.com/ory/fosite/blob/master/handler/oauth2/strategy_jwt.go#L108-L134) already?

Issue #305, opened 2018-09-26
Add support for JWT Bearer authorization grant


gperinazzo:
Feature request:

[RFC 7523](https://tools.ietf.org/html/rfc7523#section-2.1) defines the usage of JWTs as authorization grants in order to obtain access tokens through the assertion framework (described in [RFC 7521](https://tools.ietf.org/html/rfc7521)). This could be implemented as a grant type handler. The library currently only supports JWT assertions for client authentication.

aeneasr:
Hi there, I'm open to discuss how this can be added to the project. As part of that I'd like to ask some things from you:

1. Can you please describe the use case you are trying to solve and why it is important to have this flow implemented?
2. Can you comment on the applicability of this spec with regard to OpenID Connect and it's certification?

We also welcome contributions to this project. If you plan to contribute a PR for this please lay out a plan in this issue how you would like to approach it. This will save both of us a ton of time!

aeneasr:
Thank you for the explanation and sorry for the long response time. I'm wondering whether this isn't simply OAuth 2.0 Client Credentials where the authentication credentials are a [JWT signed with a private key](https://www.ory.sh/docs/guides/master/hydra/6-how-to/3-advanced#oauth-20-client-authentication-with-rsa-privatepublic-keypairs)? Could you shed some light on the differences?

aeneasr:
The part where we use assertsions as authorization grants is also implemented already by another spec, OpenID Connect request object or request url, see: https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests

gperinazzo:
> Thank you for the explanation and sorry for the long response time. I'm wondering whether this isn't simply OAuth 2.0 Client Credentials where the authentication credentials are a [JWT signed with a private key](https://www.ory.sh/docs/guides/master/hydra/6-how-to/3-advanced#oauth-20-client-authentication-with-rsa-privatepublic-keypairs)? Could you shed some light on the differences?

The difference between the two is that the JWT Bearer grant allows someone to acquire access tokens on behalf of a specific user without user interaction, while the Client Credentials grant allows clients to acquire access tokens that are not related to any user.
The JWT Bearer grant also allows anonymous clients to requests such tokens (with no client authentication required from the token endpoint).
It's similar to using a JWT to authenticate the client, but it authenticates a user instead.

[...]

PR #546, opened 2020-12-14
feat: Authorization Grants using JWT

Xopek:
## Related issue
#305

## Proposed changes
Support for JSON Web Token (JWT) Profile for OAuth 2.0 Client Authentication and Authorization Grants [(RFC7523)](https://tools.ietf.org/html/rfc7523)

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

## Further comments
Implementation details were discussed with @aeneasr beforehand


access_request_handler.go
```diff
@@ -80,14 +79,20 @@ func (f *Fosite) NewAccessRequest(ctx context.Context, r *http.Request, session
...
-	}
-	accessRequest.Client = client
-
 	var found = false
+	var isClientAuth = false
 	for _, loader := range f.TokenEndpointHandlers {
+		if !loader.CanHandleTokenEndpointRequest(accessRequest) {
+			continue
+		}
+		if !loader.CanSkipClientAuth(accessRequest) && !isClientAuth {
```
aeneasr, line 88: I think I understand what this does, but it looks a bit dangerous to me because one could be able to skip client authentication in a future regression. Imagine a misplaced `continue` statement, for example.

Could we maybe solve this another way? One such idea would be to always authenticate the client (as before), but if an error occurs, we don't return it immediately but instead iterate through all `TokenEndpointHandlers` which have `CanSkipClientAuth()`.

```
 	client, clientErr := f.AuthenticateClient(ctx, r, r.PostForm)

 	var found = false
 	for _, loader := range f.TokenEndpointHandlers {
 		if !loader.CanHandleTokenEndpointRequest(accessRequest) {
 			continue
 		}
		if !loader.CanSkipClientAuth(accessRequest) && clientErr != nil {
 			continue
 		}
		if err := loader.HandleTokenEndpointRequest(ctx, accessRequest); err == nil {
 			found = true
		// ...
        }

	if !found && clientErr != nil {
		return nil, errorsx.WithStack(clientErr)
	}
```

An alternative would be to use `CanHandleTokenEndpointRequest` and check if `accessRequest.Client` is set - although I think that is also pretty unsafe.

aeneasr (changes requested):

access_request_handler.go
Xopek: Changed it towards your suggestion, have a look, thanks!

access_request_handler_test.go
```diff
@@ -221,6 +226,172 @@ func TestNewAccessRequest(t *testing.T) {
...
+	for k, c := range []struct {
+		header    http.Header
+		form      url.Values
+		mock      func()
+		method    string
+		expectErr error
+		expect    *AccessRequest
+		handlers  TokenEndpointHandlers
+	}{
+		{
```
aeneasr, line 329: We are only testing success cases here - how about another case:

1. No client auth provided
2. 1. handler allows skipping client auth
2. 2. Handler does not allow skipping client auth
3. Expect error

compose/config.go
```diff
@@ -99,6 +99,18 @@ type Config struct {
 	// UseLegacyErrorFormat controls whether the legacy error format (with `error_debug`, `error_hint`, ...)
 	// should be used or not.
 	UseLegacyErrorFormat bool
+
+	// JWTSkipClientAuth indicates, if client authentication can be skipped, when using jwt as assertion.
+	JWTSkipClientAuth bool
```
aeneasr, line 104: I think "GrantTypeJWTBearerCXanSkipClientAuth" is longer but explains very well what this does :)

compose/config.go
```diff
@@ -99,6 +99,18 @@ type Config struct {
...
+
+	// JWTIDOptional indicates, if jti (JWT ID) claim required or not in JWT.
+	JWTIDOptional bool
```
aeneasr, line 107: Given that we have many forms of JWT, I would prefix these fields with `GrantTypeJWTBearer`

compose/config.go
```diff
@@ -198,3 +210,12 @@ func (c *Config) GetMinParameterEntropy() int {
 		return c.MinParameterEntropy
 	}
 }
+
+// GetJWTMaxDuration returns the maximum time after JWT issued date, during which the JWT is considered valid.
```
aeneasr, line 214: Shouldn't this come from the JWT's `exp` field?

handler/oauth2/flow_authorize_jwt_session.go
```diff
@@ -0,0 +1,32 @@
+/*
```
aeneasr, line 1: What do you think about moving these from `/oauth2/` to `/rfc7523/`?

handler/oauth2/flow_authorize_jwt_token.go
```diff
@@ -0,0 +1,330 @@
...
+	"github.com/ory/fosite"
+	"github.com/ory/x/errorsx"
+	"gopkg.in/square/go-jose.v2"
+	"gopkg.in/square/go-jose.v2/jwt"
+)
+
+const grantTypeJwtBearer = "urn:ietf:params:oauth:grant-type:jwt-bearer"
+
+type AuthorizeJwtGrantHandler struct {
+	AuthorizeJwtGrantStorage AuthorizeJwtGrantStorage
```
aeneasr, line 38: Idiomatic Go upper-cases abbreviations:

```suggestion
	AuthorizeJWTGrantStorage AuthorizeJWTGrantStorage
```

handler/oauth2/flow_authorize_jwt_token.go
```diff
@@ -0,0 +1,330 @@
...
+
+func (c *AuthorizeJwtGrantHandler) PopulateTokenEndpointResponse(ctx context.Context, request fosite.AccessRequester, response fosite.AccessResponder) error {
+	if !c.CanHandleTokenEndpointRequest(request) {
+		return errorsx.WithStack(fosite.ErrUnknownRequest)
+	}
+
+	// if client is authenticated, check grant types
+	if !c.CanSkipClientAuth(request) && !request.GetClient().GetGrantTypes().Has(grantTypeJwtBearer) {
+		return errorsx.WithStack(fosite.ErrUnauthorizedClient.WithHintf("The OAuth 2.0 Client is not allowed to use authorization grant \"%s\".", grantTypeJwtBearer))
+	}
```
aeneasr, line 147: This is duplicated from above - maybe it can be extracted into a common method?

handler/oauth2/flow_authorize_jwt_token.go
```diff
@@ -0,0 +1,330 @@
...
+
+	var keyID string
+	for _, header := range token.Headers {
+		if header.KeyID != "" {
+			keyID = header.KeyID
+			break
+		}
+	}
+
+	keyNotFoundMsg := fmt.Sprintf(
```
aeneasr, line 198: You could prepare the error here:

```
fosite.ErrInvalidGrant.WithHintf("No public JWK was registered .... )
```

handler/oauth2/flow_authorize_jwt_token.go
```diff
@@ -0,0 +1,330 @@
...
+		"No public JWK was registered for issuer \"%s\" and subject \"%s\", and public key is required to check signature of JWT in \"assertion\" request parameter.",
+		unverifiedClaims.Issuer,
+		unverifiedClaims.Subject,
+	)
+	if keyID != "" {
+		key, err := c.AuthorizeJwtGrantStorage.GetPublicKey(ctx, unverifiedClaims.Issuer, unverifiedClaims.Subject, keyID)
+		if err != nil {
+			return nil, errorsx.WithStack(fosite.ErrInvalidGrant.WithHint(keyNotFoundMsg).WithWrap(err).WithDebug(err.Error()))
```
aeneasr, line 206: And then just do: `errorsx.WithStack(errKeyNotFound).WithWrap(err).WithDebug(err.Error())`

handler/oauth2/flow_authorize_jwt_token.go
```diff
@@ -0,0 +1,330 @@
...
+	}
+
+	keys, err := c.AuthorizeJwtGrantStorage.GetPublicKeys(ctx, unverifiedClaims.Issuer, unverifiedClaims.Subject)
+	if err != nil {
+		return nil, errorsx.WithStack(fosite.ErrInvalidGrant.WithHint(keyNotFoundMsg).WithWrap(err).WithDebug(err.Error()))
+	}
+
+	claims := jwt.Claims{}
+	for _, key := range keys.Keys {
+		err := token.Claims(key, &claims)
```
aeneasr, line 218: Given that we are validating the signature here and in `validateToken` as well I am wondering if we could do the (compute intense because probably RSA) computation just once instead?

handler/oauth2/flow_authorize_jwt_token.go
```diff
@@ -0,0 +1,330 @@
...
+		)
+	}
+
+	if !c.JWTIssuedDateOptional && claims.IssuedAt == nil {
+		return errorsx.WithStack(fosite.ErrInvalidGrant.
+			WithHint("The JWT in \"assertion\" request parameter MUST contain an \"iat\" (issued at) claim."),
+		)
+	}
+
+	if claims.IssuedAt != nil && time.Now().Sub(claims.IssuedAt.Time()).Nanoseconds() > c.JWTMaxDuration.Nanoseconds() {
```
aeneasr, line 278: I think you can use `time.Before()` here

handler/oauth2/flow_authorize_jwt_token.go
```diff
@@ -0,0 +1,330 @@
...
+			),
+		)
+	}
+
+	var issuedDate time.Time
+	if claims.IssuedAt != nil {
+		issuedDate = claims.IssuedAt.Time()
+	} else {
+		issuedDate = time.Now()
+	}
```
aeneasr, line 292: I think you can move this block in front of the block above and then only have one check on `iat` instead of two.

handler/oauth2/flow_authorize_jwt_token.go
```diff
@@ -0,0 +1,330 @@
...
+	if claims.Expiry.Time().Sub(issuedDate).Nanoseconds() > c.JWTMaxDuration.Nanoseconds() {
```
aeneasr, line 293: I am wondering if this is useful - given that `issuedDate` is set to `time.Now()` in the call prior, when would this ever validate to false unless Nanoseconds is a negative value?

handler/oauth2/flow_authorize_jwt_token.go
```diff
@@ -0,0 +1,330 @@
...
+		return errorsx.WithStack(fosite.ErrInvalidGrant.
+			WithHintf(
+				"The JWT in \"assertion\" request parameter contains an \"exp\" (expiration time) claim with value \"%s\" that is unreasonably far in the future.",
+				claims.Expiry.Time().Format(time.RFC3339),
+			),
+		)
+	}
+
+	if !c.JWTIDOptional && claims.ID == "" {
```
aeneasr, line 302: Can the JTI really be optional? How can we prevent replay attacks if it was optional?

aeneasr (changes requested):
I hope your are not initimidated by the flood of comments. In general this looks very solid already, but I think we can improve it a bit in terms of readability and I also left some security-related ideas :)

Once the next pass is done, I will have another look at the tests, and if nothing serious comes up I think this is pretty much good to go!

Xopek:
> I hope your are not initimidated by the flood of comments. In general this looks very solid already, but I think we can improve it a bit in terms of readability and I also left some security-related ideas :)
> 
> Once the next pass is done, I will have another look at the tests, and if nothing serious comes up I think this is pretty much good to go!

No problem at all!)

compose/config.go
Xopek: Well it is calculated as expiration_time - issued_date = token duration. And here we set max allowed duration.

handler/oauth2/flow_authorize_jwt_token.go
Xopek: RFC tells that it is optional, that's why i made it so

handler/oauth2/flow_authorize_jwt_token.go
```diff
@@ -0,0 +1,330 @@
...
+	claims := jwt.Claims{}
+	if err := token.Claims(key, &claims); err != nil {
+		return errorsx.WithStack(fosite.ErrInvalidGrant.
+			WithHint("Unable to verify the integrity of the 'assertion' value.").
+			WithWrap(err).WithDebug(err.Error()),
+		)
+	}
+
+	if claims.ID != "" {
+		if err := c.AuthorizeJwtGrantStorage.MarkJWTUsedForTime(ctx, claims.ID, claims.Expiry.Time()); err != nil {
```
drwatsno, line 105: It is better to move this check to the end, otherwise the token is marked as used despite the results of the checks below.

handler/oauth2/flow_authorize_jwt_token.go
aeneasr: You mean the JWT spec, or the JWT-Bearer Auth Grant spec?

compose/config.go
aeneasr: After I read through the code I understand that this is another safety measure in cases where the expiry date is set artificially long?

aeneasr:

Xopek:
> Are you still up for the changes? :) If you need any help, let us know!

Yes! Just need to do some work and then i will return to this PR =)

access_request_handler_test.go
```diff
@@ -221,6 +226,172 @@ func TestNewAccessRequest(t *testing.T) {
...
+				"grant_type": {"foo"},
+			},
+			mock: func() {
+				handler.EXPECT().HandleTokenEndpointRequest(gomock.Any(), gomock.Any()).Return(nil)
+			},
+			handlers: TokenEndpointHandlers{handler},
+			expect: &AccessRequest{
+				GrantTypes: Arguments{"foo"},
+				Request: Request{
+					Client: client,
```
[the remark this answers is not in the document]
Xopek: It is default empty client struct. When Request is created default value is used for client, so it is just an empty struct. So we are checking, that even without client auth (and hence storage wont be called, and client struct will stay empty) everything is working. 

Considering storage - it is not expected to be called, because we are not passing any client credentials. However i agree that it is not very clear why is that. That's why i will add case, where we WILL pass client credentials and in other cases will add explicit instruction, that we are not expecting storage to be called.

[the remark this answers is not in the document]
Xopek: Also i am suggesting to rewrite the whole tests in this file completely, because each test affects other test, because they share mocks and other vars which results in that test cases are not isolated from each other. That's why new tests that were added in this PR were made using `suite` package of testify. Only in this file i write new test in the same as old ones. 

If you are not against this idea maybe i can rewrite them using suite, but i will made it in separate PR.

access_request_handler_test.go
Xopek: We are not, there is one success case and one fail case. However i will add one more without client credentials completely as you suggested)

compose/config.go
Xopek: +

compose/config.go
Xopek: ok

compose/config.go
Xopek: yes, you understand it correctly

handler/oauth2/flow_authorize_jwt_session.go
Xopek: Don't quiet understand why you want to move this particular file to rfc7523, rfc7523 is also about oauth2.0. Can you please elaborate your intention?

handler/oauth2/flow_authorize_jwt_token.go
Xopek: will change

handler/oauth2/flow_authorize_jwt_token.go
Xopek: I checked it now and both of them says it is optional

handler/oauth2/flow_authorize_jwt_token.go
Xopek: yes, why not)

handler/oauth2/flow_authorize_jwt_token.go
Xopek: yeah, thanks!

handler/oauth2/flow_authorize_jwt_token.go
Xopek: Here we take expiration time and substract issuedDate time. So even if issuedDate is set to current time, expiration date can be very far in the future.

handler/oauth2/flow_authorize_jwt_token.go
Xopek: i will delete block above completly, because it doesnt make sense, we already have on check for token lifetime, that uses both expiration and issuedAt dates

handler/oauth2/flow_authorize_jwt_token.go
Xopek: deleted this check completely)

Xopek:
Ok, i addressed all comments and made necessary changes, have a look)

aeneasr:
Hello, I wanted to review this today but am too tired to give it the attention it needs. I will try again on Monday :)

PR #567, opened 2021-03-05
feat: Token Reuse Detection

svrakitin:
## Related issue

https://github.com/ory/hydra/issues/2022

## Proposed changes

This change makes `refresh_token` grant handler to check if we received `fosite.ErrInactiveToken` error from storage, which means there is an attempt to refresh access token using previously-used or revoked refresh token.

This is similar to what is done in Authorization Code Flow. 

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

## Further comments

These are changes just for fosite. I have another PR to make for hydra, so its persistent storage stops deleting refresh tokens from database on revocation. Instead it will mark them as inactive similar to what's done for auth codes.

### DIFF
--- a/handler/oauth2/flow_refresh.go
+++ b/handler/oauth2/flow_refresh.go
@@ -64,7 +64,14 @@ func (c *RefreshTokenGrantHandler) HandleTokenEndpointRequest(ctx context.Contex
 	refresh := request.GetRequestForm().Get("refresh_token")
 	signature := c.RefreshTokenStrategy.RefreshTokenSignature(refresh)
 	originalRequest, err := c.TokenRevocationStorage.GetRefreshTokenSession(ctx, signature, request.GetSession())
-	if errors.Is(err, fosite.ErrNotFound) {
+	if errors.Is(err, fosite.ErrInactiveToken) {
+		// Detected refresh token reuse
+		if rErr := c.handleRefreshTokenReuse(ctx, signature, originalRequest); rErr != nil {
+			return errorsx.WithStack(fosite.ErrServerError.WithWrap(rErr).WithDebug(rErr.Error()))
+		}
+
+		return errorsx.WithStack(fosite.ErrInactiveToken.WithWrap(err).WithDebug(err.Error()))
+	} else if errors.Is(err, fosite.ErrNotFound) {
 		return errorsx.WithStack(fosite.ErrInvalidGrant.WithWrap(err).WithDebugf("The refresh token has not been found: %s", err.Error()))
 	} else if err != nil {
 		return errorsx.WithStack(fosite.ErrServerError.WithWrap(err).WithDebug(err.Error()))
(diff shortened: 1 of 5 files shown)


handler/oauth2/flow_refresh.go
```diff
@@ -64,7 +64,14 @@ func (c *RefreshTokenGrantHandler) HandleTokenEndpointRequest(ctx context.Contex
 	refresh := request.GetRequestForm().Get("refresh_token")
 	signature := c.RefreshTokenStrategy.RefreshTokenSignature(refresh)
 	originalRequest, err := c.TokenRevocationStorage.GetRefreshTokenSession(ctx, signature, request.GetSession())
-	if errors.Is(err, fosite.ErrNotFound) {
+	if errors.Is(err, fosite.ErrInactiveToken) {
```
aeneasr, line 67: If I understand correctly, the idea is that the storage would return a new error code `ErrInactiveToken` when a refresh token was used or revoked - am I correct with that assumption?

aeneasr (commented):
Thank you for your work on this, this looks great already!

I think it would make sense to add an e2e / integration test to https://github.com/ory/fosite/tree/master/integration to check that everything works well together!

Apart from that I think this is pretty much good to go as I think that this should not have any breaking changes, as ErrNotFound still behaves as before!

handler/oauth2/flow_refresh.go
svrakitin: Yes, exactly, same way it does for auth codes.

svrakitin:
@aeneasr Thanks. I will add some integration tests.

handler/oauth2/flow_refresh.go
aeneasr: Nice makes sense!

aeneasr:
Awesome, thank you!

integration/refresh_token_grant_test.go
```diff
@@ -203,6 +203,25 @@ func TestRefreshTokenFlow(t *testing.T) {
 			pass:  true,
 			check: func(t *testing.T, original, refreshed *oauth2.Token, or, rr *introspectionResponse) {},
 		},
+		{
+			description: "should deny access if original token was reused",
```
svrakitin, line 207: Basically goes through the flow once again and fails because of reuse.

integration/refresh_token_grant_test.go
```diff
@@ -203,6 +203,25 @@ func TestRefreshTokenFlow(t *testing.T) {
...
+				tokenSource := oauthClient.TokenSource(oauth2.NoContext, original)
+				_, err := tokenSource.Token()
+				require.Error(t, err)
+				require.Equal(t, http.StatusUnauthorized, err.(*oauth2.RetrieveError).Response.StatusCode)
+
+				refreshed.Expiry = refreshed.Expiry.Add(-time.Hour * 24)
+				tokenSource = oauthClient.TokenSource(oauth2.NoContext, refreshed)
+				_, err = tokenSource.Token()
+				require.Error(t, err)
+				require.Equal(t, http.StatusUnauthorized, err.(*oauth2.RetrieveError).Response.StatusCode)
```
aeneasr, line 222: Would it be possible to also add a test which checks if the original token was invalidated as well? I think `TokenSource` updates the token in the memory structure.

aeneasr (commented):
Thanks again and sorry for the late review, I am on holidays ;)

integration/refresh_token_grant_test.go
svrakitin: Sure, I will look into this.

svrakitin:
@aeneasr Enjoy your holidays. This is not urgent.

integration/refresh_token_grant_test.go
svrakitin: Above it checks that original refresh token was invalidated after the first use. Do you mean I should check access token?

aeneasr: Ah I overlooked that - makes sense!

aeneasr (approved):
Awesome, thank you! 🎉 Your contribution makes Ory better :)

svrakitin:
@aeneasr Thanks for pleasant contributor experience. I will prepare a PR for hydra as soon as there is a new fosite release.

aeneasr:
:) v0.39.0 is out now!

mitar:
@svrakitin Yes, I must say I always admire how well @aeneasr manages to handle maintaining this project. At the same time pretty opinionated, but still welcoming. :-) A rare mix.

aeneasr:
Thank you two, I'm blushing :relaxed:

PR #570, opened 2021-03-14
fix: Do not include nonce in ID tokens when not used

mitar:
## Proposed changes

Not sure why was this there, but it makes no sense and there is no point in making ID tokens larger just to include an empty string.

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

### DIFF
--- a/token/jwt/claims_id_token_test.go
+++ b/token/jwt/claims_id_token_test.go
@@ -30,25 +30,6 @@ import (
 	. "github.com/ory/fosite/token/jwt"
 )
 
-var idTokenClaims = &IDTokenClaims{
-	JTI:                                 "foo-id",
-	Subject:                             "peter",
-	IssuedAt:                            time.Now().UTC().Round(time.Second),
-	Issuer:                              "fosite",
-	Audience:                            []string{"tests"},
-	ExpiresAt:                           time.Now().UTC().Add(time.Hour).Round(time.Second),
-	AuthTime:                            time.Now().UTC(),
-	RequestedAt:                         time.Now().UTC(),
-	AccessTokenHash:                     "foobar",
-	CodeHash:                            "barfoo",
-	AuthenticationContextClassReference: "acr",
-	AuthenticationMethodsReference:      "amr",
-	Extra: map[string]interface{}{
-		"foo": "bar",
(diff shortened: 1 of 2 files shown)


aeneasr (commented):
Thank you! Can you please address the failing tests?

mitar:
Fixed tests.

token/jwt/claims_id_token_test.go
```diff
@@ -66,7 +66,6 @@ func TestIDTokenClaimsToMap(t *testing.T) {
 		"rat":       float64(idTokenClaims.RequestedAt.Unix()),
 		"iss":       idTokenClaims.Issuer,
 		"aud":       idTokenClaims.Audience,
-		"nonce":     idTokenClaims.Nonce,
```
aeneasr, line 69: Could you please add a test which checks if the nonce is set when it's set and vice versa? This basically removes the original test behaviour and just replaces it with the new one. A regression (forgot to set nonce) would not be detected with this.

aeneasr (changes requested):

aeneasr:

aeneasr:
Converting to draft due to lack of response

token/jwt/claims_id_token_test.go
mitar: Done.

mitar:
Addressed comments.

mitar:
I think CI is failing because of invalid configuration, not because of something I would have introduced in this PR.

aeneasr:
Possible, I restarted the CI!

aeneasr:
Right, we had this issue in another PR. I create a PR in Ory Hydra which bumps the fosite dep to your fork. If the CI passes there, then the one here would also pass!

https://github.com/ory/hydra/pull/2531

mitar:
It seems it passed.

Could this be fixed in CI somehow? It is pretty ugly that (contributing to) this repository now depends in this way on Hydra's one?

aeneasr:
> Could this be fixed in CI somehow? It is pretty ugly that (contributing to) this repository now depends in this way on Hydra's one?

done!

aeneasr:
😎

Issue #582, opened 2021-04-14
ResourceOwnerPasswordCredentialsGrantHandler returns 500 error if password in token request is incorrect


vineet-garg:
**Describe the bug**

500 error returned if wrong password specified in a token request. 
The handler is configured to use memory store.

**To Reproduce**

1. configure memory store
2. invoke 
ResourceOwnerPasswordCredentialsGrantHandler HandleTokenEndpointRequest method with wrong password in the request.
3. the error returned is 500 server error

use 

**Expected behavior**

4XX error should be returned as it is an issue with user input. 

**Environment**

- Version: v1.2.3, git sha hash
- Environment: Debian, Docker, ...

**Additional context**

The error handling in both the files : flow_resource_owner.go 
if username == "" || password == "" {
		return errorsx.WithStack(fosite.ErrInvalidRequest.WithHint("Username or password are missing from the POST body."))
	} else if err := c.ResourceOwnerPasswordCredentialsGrantStorage.Authenticate(ctx, username, password); errors.Is(err, fosite.ErrNotFound) {
		return errorsx.WithStack(fosite.ErrInvalidGrant.WithHint("Unable to authenticate the provided username and password credentials.").WithWrap(err).WithDebug(err.Error()))
	} else if err != nil {
		return errorsx.WithStack(fosite.ErrServerError.WithWrap(err).WithDebug(err.Error()))
	}

and memory.go 
func (s *MemoryStore) Authenticate(_ context.Context, name string, secret string) error {
	s.usersMutex.RLock()
	defer s.usersMutex.RUnlock()

	rel, ok := s.Users[name]
	if !ok {
		return fosite.ErrNotFound
	}
	if rel.Password != secret {
		return errors.New("Invalid credentials")
	}
	return nil
}

are not aligned and seems to be contributing to the issue.

aeneasr:
We welcome contributions!

vineet-garg:
I can fix this. The desired error code in this case is 400 invalid_grant ?

aeneasr:
I think that would be acceptable

PR #583, opened 2021-04-16
fix: 582memory store authentication error code

vineet-garg:
## Related issue
#582 
## Proposed changes

Change storage memory error to what ResourceOwnerPasswordCredentialsGrantHandler is expecting

## Checklist

- [ x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md)
      and signed the CLA.
- [ x] I have read the [security policy](../security/policy).
- [ x] I confirm that this pull request does not address a security
      the changes.
- [ x] I have added tests that prove my fix is effective or that my feature
      works.
      appropriate).


storage/memory.go
```diff
@@ -352,7 +351,7 @@ func (s *MemoryStore) Authenticate(_ context.Context, name string, secret string
 		return fosite.ErrNotFound
 	}
 	if rel.Password != secret {
-		return errors.New("Invalid credentials")
+		return fosite.ErrNotFound
```
aeneasr, line 354: This store is really just a reference. I think it’s user friendly to retutn an understandable error message here. Not sure if there is value in making this into a generic „not found“ message

vineet-garg: To provide some context, the code that is calling the storage i.e. (ResourceOwnerPasswordCredentialsGrantHandler) (excerpt pasted below) expects only ErrNotFound from implementations of storage and injects its own top-level message. Earlier the calling code was propagating the error as 500 server error. I am assuming changing the calling code to react to a more appropriate error constant will break the contract with other implementations of storage and that is not an option.

} else if err := c.ResourceOwnerPasswordCredentialsGrantStorage.Authenticate(ctx, username, password); errors.Is(err, fosite.**ErrNotFound**) {
		return errorsx.WithStack(fosite.ErrInvalidGrant.WithHint("Unable to authenticate the provided username and password credentials.").WithWrap(err).WithDebug(err.Error()))

vineet-garg: @aeneasr Please let me know if you have another way to solve this or you prefer to leave it the way it is. Thanks.

aeneasr: Hey - totally my bad, I forgot to reply here! I think we have a "WithDebug" helper on fosite errors which we could use here to provide more context, so:

```suggestion
		return fosite.ErrNotFound.WithDebug("Invalid credentials")
```

vineet-garg:
@aeneasr Is Anyone planning to fix the failures separately? The are unrelated to my PR.

aeneasr:
Yes!

aeneasr (approved):

aeneasr:
TY :)

Issue #514, opened 2020-10-19
Consider dropping go-jwt in favor of go-jose


mitar:
[go-jwt](https://github.com/dgrijalva/jwt-go) seems to not be maintained really. It has open/known [security issue](https://github.com/dgrijalva/jwt-go/issues/428) (which have multiple PRs open to address, earliest in 2018: https://github.com/dgrijalva/jwt-go/pull/286). And it requires numbers to be floats because it cannot validate ints. [PR to fix this has been made](https://github.com/dgrijalva/jwt-go/pull/162) in 2016. This is why fosite has to issue float values as expiration timestamps which are not necessary the most compatible. [See my comment here](https://github.com/dgrijalva/jwt-go/pull/162#issuecomment-712336351).

Fosite is already using [go-jose](https://github.com/square/go-jose) which seems to be more maintained. So not sure why not fully switch to it?

Or we could at least use a fork of go-jwt?

aeneasr:
Yes, let's do that!

mitar:
Another reason to move to go-jose is that go-jose supports clock skew leeway in validation: https://github.com/square/go-jose/blob/v2/jwt/validation.go#L68-L75

go-jwt does not: https://github.com/dgrijalva/jwt-go/blob/master/claims.go#L29

aeneasr:
Absolutely, if anyone is up for the challenge I'm happy to help!

narg95:
Hi, I would propose to first migrate adapting go-Jose to jwt-go types. I took a look and it is feasible without much friction/adaptation. The following jwt-go types needs to be moved to Fosite’s jwt package , which are simple and small: `jwt.Claims` Interface, `jwt.Token` Struct (only Claims and Header fields are used in fosite), claims validation logic with its validation errors. I anticipate also that concrete JWT Strategies (RSA, ECDA) will get short or even just type aliases because go-Jose provides a common interface independent of the signing algorithm, it does type checking on the private key to determine which logic and signing algorithm to use, that means less code to maintain in fosite. The biggest advantage of this approach is having a low impact on fosite external interfaces, only is required changing one import from  jwt-go to fosite’s jwt package, something easy to document in a change log, also unit tests do not need to be changed providing a good level of confidence  in the correctness of the adaptation.

in a second phase we can take advantage of go-Jose extra capabilities.

what do you think? I think, given it’s expected size I would be able to contribute to this change.

aeneasr:
That sounds like a great plan!

PR #593, opened 2021-05-06
feat: transit from jwt-go to go-jose

narg95:
## Related issue

#514 

## Proposed changes

This pull request migrates from `jwt-go` to `go-jose`. The following are the key considerations for this transition:

- `go-jose` [does not support](https://github.com/square/go-jose/issues/185) `alg:none` therefore tests trying to check this functionality where removed
- `JWTStrategy` interface was kept almost identical to provide an easy migration, except for the following changes:
    * `jwt.Token` struct type from `jwt-go` was replaced by a local and simplified type definition `Token` that covers all `fosite` use cases
    *   `jwt.MapClaims` was replaced by a local copy of that data type
    * `jwt.Claims` interface dependency to `jwt-go` was replaced by the local `MapClaims` struct type and not by another interface because it simplified the adaptation and `fosite` always type casted `jwt.Claims` to `*jwt.MapClaims`. Source code performing this type cast was therefore not needed and removed
- `jwt.ValidationErrors` from `jwt-go` was replaced by a local copy in order to keep backwards compatibility with jwt validation errors
 - `jwt.SigningMethod{Family}` from `jwt-go` has not equivalence to `go-jose` signing methods, therefore it was updated by using concrete signing method names e.g. `jwt.SigningMethodRSA` -> `jose.RS256, jose.RS384, jose.RS512`
- `RS256JWTStrategy` and `ES256JWTStrategy` types were kept to stay backwards compatible, but they are just the same implementation calling under the hood the same helper functions with some minor difference in parameters. This is so because `go-jose` makes no distinction on signing methods, this is hidden from callers and it is inferred based on the private/public key provided.

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

## Further comments

As mentioned in the issue, this is a first step in a full migration to `go-jose`. This step aims to remove `jwt-go` but keeping best-effort-backwards-compatible methods and behaviors using underneaths `go-jose`. This way simplifies the transition process. It is known that `go-jose` offer other benefits, but those need to be addressed in further pull requests.

### DIFF
--- a/.github/workflows/oidc-conformity.yml
+++ b/.github/workflows/oidc-conformity.yml
@@ -13,8 +13,8 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 2
-          repository: ory/hydra
-          ref: master
+          repository: narg95/hydra
+          ref: go-jose_migration
       - uses: actions/setup-go@v2
         with:
           go-version: '^1.15.0'
--- a/.github/workflows/oidc-conformity.yml
+++ b/.github/workflows/oidc-conformity.yml
@@ -13,8 +13,8 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 2
-          repository: narg95/hydra
-          ref: go-jose_migration
+          repository: ory/hydra
+          ref: master
       - uses: actions/setup-go@v2
         with:
           go-version: '^1.15.0'
--- a/authorize_request_handler_oidc_request_test.go
+++ b/authorize_request_handler_oidc_request_test.go
@@ -33,14 +33,14 @@ import (
 
 	"github.com/pkg/errors"
 
-	jwt "github.com/dgrijalva/jwt-go"
+	"github.com/ory/fosite/token/jwt"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	jose "gopkg.in/square/go-jose.v2"
 )
 
 func mustGenerateAssertion(t *testing.T, claims jwt.MapClaims, key *rsa.PrivateKey, kid string) string {
-	token := jwt.NewWithClaims(jwt.SigningMethodRS256, claims)
+	token := jwt.NewWithClaims(jose.RS256, claims)
 	if kid != "" {
 		token.Header["kid"] = kid
 	}
@@ -50,15 +50,8 @@ func mustGenerateAssertion(t *testing.T, claims jwt.MapClaims, key *rsa.PrivateK
 }
 
 func mustGenerateHSAssertion(t *testing.T, claims jwt.MapClaims) string {
-	token := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
+	token := jwt.NewWithClaims(jose.HS256, claims)
 	tokenString, err := token.SignedString([]byte("aaaaaaaaaaaaaaabbbbbbbbbbbbbbbbbbbbbbbcccccccccccccccccccccddddddddddddddddddddddd"))
-	require.NoError(t, err)
-	return tokenString
-}
-
-func mustGenerateNoneAssertion(t *testing.T, claims jwt.MapClaims) string {
-	token := jwt.NewWithClaims(jwt.SigningMethodNone, claims)
-	tokenString, err := token.SignedString(jwt.UnsafeAllowNoneSignatureType)
 	require.NoError(t, err)
 	return tokenString
 }
@@ -80,7 +73,6 @@ func TestAuthorizeRequestParametersFromOpenIDConnectRequest(t *testing.T) {
 
 	validRequestObject := mustGenerateAssertion(t, jwt.MapClaims{"scope": "foo", "foo": "bar", "baz": "baz", "response_type": "token", "response_mode": "post_form"}, key, "kid-foo")
 	validRequestObjectWithoutKid := mustGenerateAssertion(t, jwt.MapClaims{"scope": "foo", "foo": "bar", "baz": "baz"}, key, "")
-	validNoneRequestObject := mustGenerateNoneAssertion(t, jwt.MapClaims{"scope": "foo", "foo": "bar", "baz": "baz", "state": "some-state"})
 
 	var reqH http.HandlerFunc = func(rw http.ResponseWriter, r *http.Request) {
 		rw.Write([]byte(validRequestObject))
@@ -190,18 +182,6 @@ func TestAuthorizeRequestParametersFromOpenIDConnectRequest(t *testing.T) {
 			client:     &DefaultOpenIDConnectClient{JSONWebKeysURI: reqJWK.URL, RequestObjectSigningAlgorithm: "RS256", RequestURIs: []string{reqTS.URL}},
 			expectForm: url.Values{"response_type": {"token"}, "response_mode": {"post_form"}, "scope": {"foo openid"}, "request_uri": {reqTS.URL}, "foo": {"bar"}, "baz": {"baz"}},
 		},
-		{
-			d:          "should pass when request object uses algorithm none",
-			form:       url.Values{"scope": {"openid"}, "request": {validNoneRequestObject}},
-			client:     &DefaultOpenIDConnectClient{JSONWebKeysURI: reqJWK.URL, RequestObjectSigningAlgorithm: "none"},
-			expectForm: url.Values{"state": {"some-state"}, "scope": {"foo openid"}, "request": {validNoneRequestObject}, "foo": {"bar"}, "baz": {"baz"}},
-		},
-		{
-			d:          "should pass when request object uses algorithm none and the client did not explicitly allow any algorithm",
-			form:       url.Values{"scope": {"openid"}, "request": {validNoneRequestObject}},
-			client:     &DefaultOpenIDConnectClient{JSONWebKeysURI: reqJWK.URL},
-			expectForm: url.Values{"state": {"some-state"}, "scope": {"foo openid"}, "request": {validNoneRequestObject}, "foo": {"bar"}, "baz": {"baz"}},
-		},
 	} {
 		t.Run(fmt.Sprintf("case=%d/description=%s", k, tc.d), func(t *testing.T) {
 			req := &AuthorizeRequest{
@@ -217,10 +197,15 @@ func TestAuthorizeRequestParametersFromOpenIDConnectRequest(t *testing.T) {
 				if tc.expectErrReason != "" {
 					real := new(RFC6749Error)
 					require.True(t, errors.As(err, &real))
-					assert.EqualValues(t, real.Reason(), tc.expectErrReason)
+					assert.EqualValues(t, tc.expectErrReason, real.Reason())
 				}
 			} else {
-				require.NoError(t, err)
+				if err != nil {
+					real := new(RFC6749Error)
+					errors.As(err, &real)
+					require.NoErrorf(t, err, "Hint: %v\nDebug:%v", real.HintField, real.DebugField)
+				}
+				require.NoErrorf(t, err, "%+v", err)
 				require.Equal(t, len(tc.expectForm), len(req.Form))
 				for k, v := range tc.expectForm {
 					assert.EqualValues(t, v, req.Form[k])
--- a/authorize_request_handler_oidc_request_test.go
+++ b/authorize_request_handler_oidc_request_test.go
@@ -56,6 +56,13 @@ func mustGenerateHSAssertion(t *testing.T, claims jwt.MapClaims) string {
 	return tokenString
 }
 
+func mustGenerateNoneAssertion(t *testing.T, claims jwt.MapClaims) string {
+	token := jwt.NewWithClaims(jwt.SigningMethodNone, claims)
+	tokenString, err := token.SignedString(jwt.UnsafeAllowNoneSignatureType)
+	require.NoError(t, err)
+	return tokenString
+}
+
 func TestAuthorizeRequestParametersFromOpenIDConnectRequest(t *testing.T) {
 	key, err := rsa.GenerateKey(rand.Reader, 1024)
 	if err != nil {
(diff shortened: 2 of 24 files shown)


narg95:
I do not understand why the conformity `OpenID Connect Conformity Tests / oidc-conformity (pull_request) ` is failing. It is also happening in main branch

aeneasr:
The problem here is I think one of forks - on master I have to double check. Basically we are doing a module rewrite with the latest PR commit hash in go.mod of Ory Hydra for fosite. This fails apparently for forks. Also GitHub recently changed the way how SHA is computed for PRs as they use a virtual merge commit to do the diff or something. Long story short - probably not related to your changes but we need to fix it before merge. I thought i fixed it but apparently not so for forks

aeneasr:
I've created https://github.com/ory/hydra/pull/2525 - it shows how to upgrade fosite to point to your branch. To ensure that OIDC works correctly and there are no serious regressions could you please get Ory Hydra tests to pass? If they're green there we can merge this one here right away! Thank you :)

narg95:
@aeneasr sure, I am on it, so far it compiles, linter is happy, I am taking a look to the tests. I had to upgrade my go to 1.6 to use `import "embed"`. `make docker` works. So it looks promising :)

authorize_request_handler_oidc_request_test.go
```diff
@@ -190,18 +182,6 @@ func TestAuthorizeRequestParametersFromOpenIDConnectRequest(t *testing.T) {
 			client:     &DefaultOpenIDConnectClient{JSONWebKeysURI: reqJWK.URL, RequestObjectSigningAlgorithm: "RS256", RequestURIs: []string{reqTS.URL}},
 			expectForm: url.Values{"response_type": {"token"}, "response_mode": {"post_form"}, "scope": {"foo openid"}, "request_uri": {reqTS.URL}, "foo": {"bar"}, "baz": {"baz"}},
 		},
-		{
-			d:          "should pass when request object uses algorithm none",
```
aeneasr, line 194: A particular reason for removing these? These are neccessary for some OIDC flows :/

aeneasr (commented):
I noticed that `none` was removed from the test assertions - it's unfortunately needed to pass OIDC certification iirc.

I started a test run for OIDC certification here: https://github.com/ory/hydra/pull/2526/checks?check_run_id=2554759989

Let's see if it passes :)

authorize_request_handler_oidc_request_test.go
narg95: I removed them because go-jose [does not support](https://github.com/square/go-jose/issues/185) alg:none therefore tests trying to check this functionality where removed. I assumed it was just a sanity check, alg:none is seen as insecure and a door for vulnerabilities. But if required I can provide that functionality. It is just to base64-urlencoded `header.claims.empty-signature`, right?

aeneasr: Yeah, unfortunately it's still supported as an algorithm in OIDC for whatever reason (probably some legacy corp not being able to implement JWT with algs) - let's see the result of https://github.com/ory/hydra/pull/2526/checks?check_run_id=2554759989 and if it passes we can leave this removed

aeneasr:
As suspected, the OIDC tests fail with the none alg missing: https://github.com/ory/hydra/pull/2526/checks?check_run_id=2554759989#step:5:114

aeneasr:
Thank you! I approved the run just now. It's really annoying that GitHub requires us to hit "Approve Run" over and over again :( Especially if the task takes an hour or more to finish...

token/jwt/token.go
```diff
@@ -0,0 +1,229 @@
...
+type Keyfunc func(*Token) (interface{}, error)
+
+func Parse(tokenString string, keyFunc Keyfunc) (*Token, error) {
+	return ParseWithClaims(tokenString, MapClaims{}, keyFunc)
+}
+
+// Parse, validate, and return a token.
+// keyFunc will receive the parsed token and should return the key for validating.
+// If everything is kosher, err will be nil
+func ParseWithClaims(rawToken string, claims MapClaims, keyFunc Keyfunc) (*Token, error) {
```
aeneasr, line 152: I think we need tests for this, as we're using it to verify authentication tokens and it looks like it has several moving parts. I think as part of the tests we should have:

1. Completely invalid token like`asdf`
2. Token that is not a JWT `foo.bar.baz`
3. Token that mismatches the keyfunc:
   1. Expect RSA256 but get RSA512
   2. Expect RSA256 but get None
5.  RSA256 Token with unknown signing key
6. Token with invalid claims:
   1. Expired
   1. Incorrect IAT
   2. ...and other cases that should be covered by this

And finally also some passing tokens like:

1. RSA
2. None
3. ECDSA

aeneasr (commented):
Awesome, thank you for the great work! It looks like the CI in Ory Hydra is passing so this is a ✅ 😃 

I think we need a couple more tests in here but otherwise this looks really solid! I'll go over the jwt token validation strategy once again when tests have been added.

All in all thank you for the great contribution!

aeneasr:
I've updated the CI config so it now runs the conformity tests directly here!

authorize_request_handler_oidc_request_test.go
```diff
@@ -217,10 +217,15 @@ func TestAuthorizeRequestParametersFromOpenIDConnectRequest(t *testing.T) {
...
-					assert.EqualValues(t, real.Reason(), tc.expectErrReason)
+					assert.EqualValues(t, tc.expectErrReason, real.Reason())
 				}
 			} else {
-				require.NoError(t, err)
+				if err != nil {
+					real := new(RFC6749Error)
+					errors.As(err, &real)
+					require.NoErrorf(t, err, "Hint: %v\nDebug:%v", real.HintField, real.DebugField)
+				}
```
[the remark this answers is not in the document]
narg95: kind of, it helps to print more details about an unexpected error to find its cause easier :)
It helped me

token/jwt/token_test.go
```diff
@@ -0,0 +1,47 @@
+package jwt
```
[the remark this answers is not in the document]
narg95: yes, I took most of `parse_test.go` test cases, I prune those we do not need or support e.g. `json.Number` and added specific code to prevent regresses like returning always a toke instance and returning a specific validation error when a Key was not provided, the rest worked perfect. I also appended the test cases that you mentioned.

token/jwt/map_claims.go
```diff
@@ -0,0 +1,154 @@
+package jwt
```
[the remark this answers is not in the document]
narg95: This one did not have explicit tests, it was tested implicitly in `jwt-go` in `parse_test.go`. Those tests I copied.

token/jwt/token.go
narg95: done

narg95:
The OpenID conformity tests are broken because they run based on hydra `main` branch which is incompatible.
I have updated `hydra` `go-jose_migration` branch to point to latest commit here, so we can double check that conformity tests pass.

narg95:
I hope those tests provide the confidence required to merge it 🤞 
This PR is becoming bigger ;)

.github/workflows/oidc-conformity.yml
```diff
@@ -13,8 +13,8 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 2
-          repository: ory/hydra
-          ref: master
+          repository: narg95/hydra
```
narg95, line 16: This is a small hack :) to get the OpenID conformity test running in sync with the migration branch used in hydra PR.
This needs to be reverted before merging

narg95:
hi @aeneasr, perhaps you did not noticed it :) but I already addressed your comments and conformity tests are green, so from my point of view this is ready to be merged. Could you take a look?

Btw. before or after merging the `.github/workflows/oidc-conformity.yml` file must be reverted to point to `ory/hydra` repo and ` master` branch, as I explained above, this was needed to run conformity tests with the correct hydra's go-migration branch.

token/jwt/map_claims.go
```diff
@@ -0,0 +1,154 @@
...
+type MapClaims map[string]interface{}
+
+// Compares the aud claim against cmp.
+// If required is false, this method will return true if the value matches or is unset
+func (m MapClaims) VerifyAudience(cmp string, req bool) bool {
+	aud, ok := m["aud"].([]string)
+	if !ok {
+		strAud, ok := m["aud"].(string)
+		if !ok {
+			return false
```
aeneasr, line 25: If I read this correctly, this would fail when `aud` is not set and `req` is `false` which is contrary to what `If required is false, this method will return true if the value matches or is unset` says.

I think a test for this would be good!

token/jwt/token_test.go
```diff
@@ -0,0 +1,379 @@
...
+		{
+			name:        "valid format invalid content",
+			tokenString: "foo.bar.baz",
+			keyfunc:     defaultKeyFunc,
+			claims:      MapClaims(nil),
+			valid:       false,
+			errors:      ValidationErrorMalformed,
+		},
+		{
+			name:        "wrong key, expected RSA got ECDSA",
```
aeneasr, line 158: This appears to be

```suggestion
			name:        "wrong key, expected ECDSA got RSA",
```

token/jwt/token_test.go
```diff
@@ -0,0 +1,379 @@
...
+			tokenString: "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJmb28iOiJiYXIifQ.FhkiHkoESI_cG3NPigFrxEk9Z60_oXrOT2vGm9Pn6RDgYNovYORQmmA0zs1AoAOf09ly2Nx2YAg6ABqAYga1AcMFkJljwxTT5fYphTuqpWdy4BELeSYJx5Ty2gmr8e7RonuUztrdD5WfPqLKMm1Ozp_T6zALpRmwTIW0QPnaBXaQD90FplAg46Iy1UlDKr-Eupy0i5SLch5Q-p2ZpaL_5fnTIUDlxC3pWhJTyx_71qDI-mAA_5lE_VdroOeflG56sSmDxopPEG3bFlSu1eowyBfxtu0_CuVd-M42RU75Zc4Gsj6uV77MBtbMrf4_7M_NUTSgoIF3fRqxrj0NzihIBg",
+			keyfunc:     publicECDSAKey,
+			claims:      MapClaims{"foo": "bar"},
+			valid:       false,
+			errors:      ValidationErrorSignatureInvalid,
+		},
+		{
+			name:        "nokey, expected RSA got ECDSA",
```
aeneasr, line 166: This appears to be

```suggestion
			name:        "nokey, got RSA but found no key",
```

token/jwt/token_test.go
```diff
@@ -0,0 +1,379 @@
...
+		{
+			name:        "should fail, valid NONE token with invalid verification key",
+			tokenString: "",
+			keyfunc:     defaultKeyFunc,
+			claims:      MapClaims{"foo": "bar"},
+			valid:       false,
+			errors:      ValidationErrorSignatureInvalid,
+			signingKey:  UnsafeAllowNoneSignatureType,
+			method:      SigningMethodNone,
+		},
```
aeneasr, line 218: Could you please add one more test here:

1. `should fail, expected ECDSA but got NONE` which is the invert to https://github.com/ory/fosite/pull/593/files#diff-62425fcddec1b39f6b477e442647c937f444850b79d1acdbfc8b2ad7289d7a31R210

aeneasr (commented):
Thank you! This looks much better already :) Sorry for the slow process in this PR. My workload is so intense at the moment and these type of PRs (that touch cryptography) are always very difficult to review!

I'm leaving for holiday on saturday for ~10 days but I think we can get it merged before :)

token/jwt/map_claims.go
narg95: Good catch, agree, actually the error is in the forked `https://github.com/form3tech-oss/jwt-go/blob/v3.2.1/map_claims.go` and not in `https://github.com/dgrijalva/jwt-go/blob/v3.2.0/map_claims.go`. I will double check and update the PR with the fix and some tests with a new commit so you do not have to review all again.

narg95:
> Thank you! This looks much better already :) Sorry for the slow process in this PR. My workload is so intense at the moment and these type of PRs (that touch cryptography) are always very difficult to review!
> 
> I'm leaving for holiday on saturday for ~10 days but I think we can get it merged before :)

It is understandable, it touches a core component. I think we can manage a merge before Saturday, I think I can get the changes by tomorrow

token/jwt/map_claims.go
narg95: I updated with latest `map_claims.go` from `form3tech-oss/jwt-go` which accepts `aud` as an array. I noted that repo provides `map_claims_test.go`, I copy them and added a test to check for the inconsistency that you found.

token/jwt/token_test.go
narg95: Good catch, fixed, I structured a bit better the tests to make it more explicit what is given and what is expected, I copied tests from `dgrijalva/jwt-go` were a bit confusing.

token/jwt/token_test.go
narg95: fixed

token/jwt/token_test.go
narg95: done

aarmam:
> Thank you! This looks much better already :) Sorry for the slow process in this PR. My workload is so intense at the moment and these type of PRs (that touch cryptography) are always very difficult to review!
> 
> I'm leaving for holiday on saturday for ~10 days but I think we can get it merged before :)

That would be great! I'm looking to add HSM support and this PR would be greatly appreciated!

.github/workflows/oidc-conformity.yml
```diff
@@ -13,8 +13,8 @@ jobs:
...
+          ref: go-jose_migration
```
aeneasr, line 17:
```suggestion
          repository: ory/hydra
          ref: master
```

aeneasr (approved):
Awesome, good job! Will release this as the next minor version :)

mitar:
It is worth noting that now JWT tokens do not unclude `typ` anymore in the header: https://github.com/square/go-jose/issues/126

aeneasr:
https://github.com/ory/fosite/issues/606

token/jwt/map_claims.go
```diff
@@ -0,0 +1,162 @@
...
+// There is no accounting for clock skew.
+// As well, if any of the above claims are not in the token, it will still
+// be considered a valid claim.
+func (m MapClaims) Valid() error {
+	vErr := new(ValidationError)
+	now := TimeFunc().Unix()
+
+	if !m.VerifyExpiresAt(now, false) {
+		vErr.Inner = errors.New("Token is expired")
+		vErr.Errors |= ValidationErrorExpired
```
mitar, line 99: @ narg95 One late question about this change. So I was looking into this codebase while debugging on our side and I noticed that `Errors` is a bitset value, where multiple errors can be set. I like that. But, everywhere else code is using `==` to test for various types of errors like in the switch/case case, where it is `case jwt.ValidationErrorExpired` for example, that switch/case would reach the end if there are two errors set at any point, no?

The `ve.Errors == jwt.ValidationErrorExpired` check which is there is in fact OK, because that case should be handled only if only `ValidationErrorExpired` is set.

PR #595, opened 2021-05-14
refactor: Generate claims in the same way

mitar:
This is a followup to https://github.com/ory/fosite/pull/522.

### DIFF
--- a/token/jwt/claims_jwt_test.go
+++ b/token/jwt/claims_jwt_test.go
@@ -75,7 +75,7 @@ func TestAssert(t *testing.T) {
 		ToMapClaims().Valid())
 	assert.NotNil(t, (&JWTClaims{NotBefore: time.Now().UTC().Add(time.Hour)}).
 		ToMapClaims().Valid())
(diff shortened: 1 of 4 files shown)


mitar:
I think this addresses your concerns from #522. Extra cannot override anything. But we also remove any fields which would have invalid value/is not provided.

token/jwt/claims_jwt_test.go
```diff
@@ -75,7 +75,7 @@ func TestAssert(t *testing.T) {
 		ToMapClaims().Valid())
 	assert.NotNil(t, (&JWTClaims{NotBefore: time.Now().UTC().Add(time.Hour)}).
 		ToMapClaims().Valid())
-	assert.NotNil(t, (&JWTClaims{NotBefore: time.Now().UTC().Add(-time.Hour)}).
+	assert.Nil(t, (&JWTClaims{NotBefore: time.Now().UTC().Add(-time.Hour)}).
```
mitar, line 78: See https://github.com/ory/fosite/pull/522/files#r513190253

aeneasr (approved):

Issue #598, opened 2021-05-22
`auth_time` id_token claim changed to `float64`


narg95:
**Describe the bug**

PR #595 introduced a breaking change on `auth_time` claim data type in `id_token`. Previously the `auth_time` claim was of type `int64` but [this change](TestAuthCodeWithDefaultStrategy/case=ensure_consistent_claims_returned_for_userinfo) transformed it to `float64`. Although values are still equivalent `hydra` tests expects an integer.

Test: `TestAuthCodeWithDefaultStrategy/case=ensure_consistent_claims_returned_for_userinfo` expects the `auth_time` 

**To Reproduce**

1. Checkout hydra's integration with fosite `v0.40` branch from PR [#2526](https://github.com/ory/hydra/pull/2526) 
2. Run 

```go
$ go test -failfast -v -short -tags sqlite -run TestAuthCodeWithDefaultStrategy/case=ensure_consistent_claims_returned_for_userinfo ./oauth2
```

**Expected behavior**

Test PASS

PR #599, opened 2021-05-22
fix: revert float64 auth_time claim

narg95:
## Related issue
#598 

## Proposed changes
Revert `auth_time` claim data type to `int64` in `id_token`. 
It was changed to `float64` in PR #595

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

## Further comments

In another PR timestamp claims for access tokens and id tokens can be transformed to `int64`. 
This is no longer a limitation since `fosite` now uses `go-jose`

### DIFF
--- a/token/jwt/claims_id_token.go
+++ b/token/jwt/claims_id_token.go
@@ -110,7 +110,7 @@ func (c *IDTokenClaims) ToMap() map[string]interface{} {
 	}
 
 	if !c.AuthTime.IsZero() {
-		ret["auth_time"] = float64(c.AuthTime.Unix()) // jwt-go does not support int64 as datatype
+		ret["auth_time"] = c.AuthTime.Unix()
 	} else {
 		delete(ret, "auth_time")
 	}
--- a/token/jwt/claims_id_token_test.go
+++ b/token/jwt/claims_id_token_test.go
@@ -70,7 +70,7 @@ func TestIDTokenClaimsToMap(t *testing.T) {
 		"baz":       idTokenClaims.Extra["baz"],
 		"at_hash":   idTokenClaims.AccessTokenHash,
 		"c_hash":    idTokenClaims.CodeHash,
-		"auth_time": float64(idTokenClaims.AuthTime.Unix()),
+		"auth_time": idTokenClaims.AuthTime.Unix(),
 		"acr":       idTokenClaims.AuthenticationContextClassReference,
 		"amr":       idTokenClaims.AuthenticationMethodsReference,
 	}, idTokenClaims.ToMap())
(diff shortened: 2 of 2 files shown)


mitar:
I do not think this should be reverted, but a test in Hydra should be fixed.

Alternatively, with us moving away from jwt-go, maybe jose has int64 support? Then I would suggest that together with moving to jose, we also change all those float64 to int64.

aeneasr (approved):
Thank you! This should indeed be an int!

If you revert the changes to the CI we can merge right away! :)

narg95:
@aeneasr done
you mean the gihub workflow for conformity test

narg95:
Hi @mitar

although it sounds **going backwards** actually it is going forward to remove `float64` on timestamps, so I think it is a good idea to keep this change. Also adjusting hydra tests would be kind of hacky, I already took a look to it, and we would need to revert that again when we move to `int64`.

Let's get hydra and fosite in harmony and then let's take advantage of `go-jose`

aeneasr (approved):

narg95:
awesome, it was fast :)
could you release `v0.40.1` ?

aeneasr:
done!

mitar:
All of them should be int! Why making just this one int just because it happens to be tested in some way and others not? Then make also `iat`, `rat`, and `exp` an int.

So I think it is better if they are all float or all int. Mixing is not good. Why would we want to have them mixed?

aeneasr:
True, I missed that during review. @narg95 could you please address that? :)

narg95:
Sure, that is the idea.

@mitar I get your point and I agree but also noted that your change broke the implicit contract between hydra and fosite and hydra was broken. I am just trying to fix it with the less effort and bring harmony, currently it is a lot of tedious work keeping Hydra and Fosite repos/branches in sync due to the breaking changes of go-jose and OpenID conformity tests. 
Pilling more changes requires more massaging on hydra tests and under the current setup, as mentioned, it is tedious work.

Let's bring now consistency on timestamp data type, we can now friction-less make improvements, of course once hydra PR [#2526](https://github.com/ory/hydra/pull/2526) gets merged.

aeneasr:
Which change broke something? I’d like to know to do a better job maintaining the library :)

mitar:
The change this PR is reverting. So me unifying all timestamps into floats (while I do think it is better to go to all ints, I was primarily wanting to unify all timestamps and with jwt-go floats were the only option). What broke in Hydra you can see in CI for this commit: https://github.com/ory/hydra/pull/2526/commits/c8dd1549b22055ba1f686fe07100e6b9b6395308

Issue #351, opened 2019-03-09
Redis storage


rps-repo:
Please use issues only to file potential bugs or request features. For everything else please go to
the [ORY Community](https://community.ory.sh/) or join the [ORY Chat](https://www.ory.sh/chat).

Sorry but forum doesn't work and no one responds on discord.  Can someone give an example of how to use redis for storage?  I see the reference to https://github.com/ory/fosite/blob/master/storage/memory.go but I have no idea how to make redis work with that.

aeneasr:
What issues do you have with the forum? There is no redis example, you can check out the SQL implementation though: https://github.com/ory/hydra/blob/master/oauth2/fosite_store_sql.go

Also redis is kind of the wrong database for this IMO.

deepthi-bhattachar:
Hi... I am new to Golang and trying to get a project code up and running(2 years old code). The code uses OAuthStore for compose.Compose as shown below. 

I am trying to use FositeSQLStore instead of MemoryStore(storage.NewExampleStore()), so far no luck. Any help would be great. Please provide any sample code which implements FositeSQLStore.

//main.go
var authProvider = compose.Compose(
   config,
   oauthStore,
   strategy,
   createUserPasswordGrantHandler,
   createRefreshTokenGrantHandler,
)

//oauth2.go
type OAuthStore struct {
	DB            *sql.DB
	Auth          Authenticator
	TokenDuration time.Duration
	Clock         utils.ClockInterface
}

Thanks in advance.

PR #600, opened 2021-05-24
feat: use int64 type for claims with timestamps

nestorvw:
## Proposed changes

Uses `int64` for timestamped claim values. 

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).


narg95:
@aeneasr and @mitar: This PR transform claims with timestamps to `int64` data type
Marshalling to a JWT token works out-of-the-box with `go-jose`, but unmarshalling  does not.
Therefore I opened issue  [#351](https://github.com/square/go-jose/issues/351) in `go-jose` and in the meantime I workarounded  it with a simple `float64->int64` conversion on first level claims.

token/jwt/claims_jwt.go
```diff
@@ -183,38 +183,23 @@ func (c *JWTClaims) FromMap(m map[string]interface{}) {
...
 				if c.ScopeField == JWTScopeFieldString {
 					c.ScopeField = JWTScopeFieldBoth
 				} else if c.ScopeField == JWTScopeFieldUnset {
 					c.ScopeField = JWTScopeFieldList
 				}
 			case []interface{}:
-				c.Scope = make([]string, len(v.([]interface{})))
-				for i, vi := range v.([]interface{}) {
+				c.Scope = make([]string, len(s))
+				for i, vi := range s {
```
mitar, line 202: Nice cleanup. ;-)

token/jwt/claims_jwt.go
```diff
@@ -240,6 +225,17 @@ func (c *JWTClaims) FromMap(m map[string]interface{}) {
 	}
 }
 
+func toTime(v interface{}, def time.Time) (t time.Time) {
+	t = def
+	switch a := v.(type) {
+	case float64:
+		t = time.Unix(int64(a), 0).UTC()
```
mitar, line 232: Do we want to preserve sub-seconds if they are ever passed? Like:

```
	secs := int64(1)
	nsecs := int64((1 - float64(secs)) * 1e9)
	t := time.Unix(secs, nsecs).UTC()
```

token/jwt/map_claims.go
```diff
@@ -76,16 +71,29 @@ func (m MapClaims) VerifyIssuer(cmp string, req bool) bool {
...
 }
 
+func (m MapClaims) toInt64(claim string) (int64, bool) {
+	switch t := m[claim].(type) {
+	case float64:
+		return int64(t), true
+	case int64:
+		return t, true
+	case json.Number:
+		v, err := t.Int64()
```
mitar, line 88: Maybe check if it is `Float64()`, too?

mitar:
I like it!

token/jwt/claims_jwt.go
narg95: Thanks :)

token/jwt/claims_jwt.go
narg95: although it like the idea [RFC 7519](https://datatracker.ietf.org/doc/html/rfc7519) states that core timestamp claims `exp` and `iat` MUST be given in seconds:

> The "exp" (expiration time) claim ... Its value MUST be a number
   containing a NumericDate value.  Use of this claim is OPTIONAL.

> The "iat" (issued at) claim  ... MUST be a number containing a NumericDate value.  Use of this
   claim is OPTIONAL.
   
> NumericDate
      A JSON numeric value representing the number of seconds from
      1970-01-01T00:00:00Z UTC until the specified UTC date/time,
      ignoring leap seconds. 
      
besides that, it requires some state to make `ToMap` and `FromMap` consistent/equivalent. Once the `time.Time` is created from float64 with sub-seconds, the inverse conversion (`ToMap`) must also contains sub-seconds, which means using a float64 instead of a int64 (which is contrary to what we are doing in this PR :) ), but currently there is no way to infer this, so it will required a flag for that and also we will keep using float64 for some cases. 

it looks nice and clever, but at the end makes things more complicated, and not RFC compliant.

token/jwt/map_claims.go
narg95: good catch, done

aeneasr:
Could you please take a look at the failing test cases? Thanks! :)

narg95:
fixed

token/jwt/claims_jwt.go
aeneasr: Agreed @narg95 !

aeneasr (approved):

aeneasr (approved):
Great job as always @narg95 - thank you!

narg95:
@aeneasr my pleasure :)

Issue #590, opened 2021-04-28
Support for clients' secret rotation


simonbrynych:
**Is your feature request related to a problem? Please describe.**

Currently it's not possible to simultaneously rotate secrets of individual clients.

**Describe the solution you'd like**

That would be great to set two passwords while both of them are valid in that time. That allows user to smoothly rotate secret without any outage.

```
&fosite.DefaultClient{
  ID:           ID,
  Secret:       [originalSecret, newSecret],
  RedirectURIs:  redirectURIs,
  ResponseTypes: responses,
  GrantTypes:    grants,
  Scopes:        scopes,
}
```

aeneasr:
The OAuth2 "compliant" (or rather OpenID Connect) way to do this is to create a copy of the client with a new password and update the client ID and the client secret in the OAuth2 consumers. This is because OpenID Connect Dynamic Client Registration currently does not have a spec for secret rotation and the API design more or less prohibits secret rotation.

Alternative would be using JWT-based auth for clients (`private_key_jwt`) which supports rotation out of the box.

aeneasr:
Nevertheless - I completely understand the feature and think it's a good idea to support that!

simonbrynych:
We want to automate entire process so having two active secrets is preferred way for us. What you think about a way how to get this feature done? Can we kindly ask you to add this functionality?

aeneasr:
That's not how open source works ;) But you can come up with an idea how of to implement it, discuss it in this issue, and then execute on the plan in a PR

simonbrynych:
Ok, we will prepare an idea PR and we are gonna share that with you. ;]

rplnt:
@aeneasr Hi, we have looked into this and have the following proposal:

1. Create new interface that would expand on `fosite.Client`, adding just one extra method returning secondary hashed secret. E.g.:
```go
type ClientWithSecretRotation interface {
	Client
	// GetHashedSecretPrevious returns the previous hashed secret as it is stored in the store.
	GetHashedSecretPrevious() []byte
}
```
2. In `fosite.AuthenticateClient` (as well as in `fosite.NewIntrospectionRequest`) check first secret, but on failure try to convert to the new interface to see if secondary secret is available and check that as well. E.g.:
```go
func (f *Fosite) checkClientSecret(ctx context.Context, client Client, clientSecret []byte) error {
	var err error
	err = f.Hasher.Compare(ctx, client.GetHashedSecret(), []byte(clientSecret))
	if err == nil {
[...]

PR #608, opened 2021-06-09
feat: Add rotation support for client secrets

rplnt:
## Related issue
Issue: #590
Discussed with: @aeneasr

## Proposed changes
The aim of this PR is to introduce support for rotation of client's secrets.
New interface is created that extends `fosite.Client` so as not to break compatibility of custom clients. The `fosite.DefaultClient` implements this interface by introducing `RotatedSecrets` field. When not set (nil), the functionality stays the same as before.

## Checklist
      and signed the CLA.
      the changes.
      works.
      appropriate).

## Further comments
Detail of how the change above is utilized in the fosite code:
1. check regular (main, current) secret by using `GetHashedSecret` method
2. if 1. fails, try to see if provided client implements `fosite.ClientWithSecretRotation`
3. if 2. succeeds, call `GetRotatedHashes` and loop over provided hashes to compare the secret 

Some questions I have:
- I'm not sure if the addition of this feature is documented enough, or if there is some other place I could add it to?
- By having the `checkClientSecret` method be private, I'm unable to test it directly, so only a few basic scenarios were added testing it via the methods that use it. Should I opt for public method just in order to test it better? Any other preference?

### DIFF
--- a/client_authentication_test.go
+++ b/client_authentication_test.go
@@ -183,6 +183,12 @@ func TestAuthenticateClient(t *testing.T) {
 		},
 		{
 			d:      "should pass because client is confidential and id and secret match in post body",
+			client: &DefaultOpenIDConnectClient{DefaultClient: &DefaultClient{ID: "foo", Secret: []byte("invalid_hash"), RotatedSecrets: [][]byte{barSecret}}, TokenEndpointAuthMethod: "client_secret_post"},
+			form:   url.Values{"client_id": []string{"foo"}, "client_secret": []string{"bar"}},
+			r:      new(http.Request),
+		},
+		{
+			d:      "should pass because client is confidential and id and rotated secret match in post body",
 			client: &DefaultOpenIDConnectClient{DefaultClient: &DefaultClient{ID: "foo", Secret: barSecret}, TokenEndpointAuthMethod: "client_secret_post"},
 			form:   url.Values{"client_id": []string{"foo"}, "client_secret": []string{"bar"}},
 			r:      new(http.Request),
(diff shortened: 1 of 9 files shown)


client_authentication_test.go
```diff
@@ -512,6 +537,28 @@ func TestAuthenticateClient(t *testing.T) {
 	}
 }
 
+// func TestCheckClientSecret(t *testing.T) {
```
aeneasr, line 540: Do you want to enable this test or remove it? :)

aeneasr (changes requested):
Awesome, thank you for your contribution! This looks very good. Just one or two minor things :)

client_authentication_test.go
rplnt: Tried to clean commit history and somewhat ended up with this there :)

Removed it now, but it relates to my last question.

aeneasr:
Thank you! I think this looks good. Documentation could be added in the README, otherwise we don't really have docs. Regarding the tests, I think the current set up is sufficient!

Sorry for the long review time, I missed the force push in my notifications :/

rplnt:
@aeneasr It's all right, just let me know if there's something else needed.

Haven't found a great section to put it into, but it's in the example that is in the readme at least.

aeneasr:
Maybe you could do a follow up PR in https://github.com/ory/fosite-example/ ? With rotation enabled there. I think that's probably the best option.

Merging this one :)

seamory:
I'm confuse what does the RotatedSecrets do? Anyone who can help me explain it? Thanks.

rplnt:
@seamory You can set `RotatedSecrets` in DefaultClient as a slice of secrets and when the main secret check fails the rotated secrets are checked instead. The purpose is to be able to rotate secrets without any disruption (or synchronisation between services) as multiple secrets can be valid at the same time.

Have a look here https://github.com/ory/fosite-example/pull/33 (it's using changes in `NewExampleStore` from this PR)

seamory:
> @seamory You can set `RotatedSecrets` in DefaultClient as a slice of secrets and when the main secret check fails the rotated secrets are checked instead. The purpose is to be able to rotate secrets without any disruption (or synchronisation between services) as multiple secrets can be valid at the same time.
> 
> Have a look here [ory/fosite-example#33](https://github.com/ory/fosite-example/pull/33) (it's using changes in `NewExampleStore` from this PR)

Thanks, i'm clear now.

Issue #615, opened 2021-08-05
Globalization of error messages


vivshankar:
**Is your feature request related to a problem? Please describe.**

I am looking for a way to provide translated error descriptions based on the request locale. This is a common enough requirement when a service is provided globally. 

**Describe the solution you'd like**

1. Supply a directory of locale bundles in the Config. Nil falls back to the current behavior.
2. Wrap the hard-coded error message strings with `GetTranslatedMessage({message_code}, {default message})`. The current messages would be the second parameter. This ensures backward compatibility if the locale bundles are not supplied as part of Config.

**Describe alternatives you've considered**

Compare based on the verbose error messages and translate prior to sending response. This isn't a scalable approach, given the project will continue to evolve and new messages would be added.

**Additional context**

aeneasr:
Hello there and thank you for the idea. Unfortunately the OAuth2 spec does not have translation thought in and requires certain error messages and codes (in plaintext english). So you would fail some certifications (e.g. OpenID Connect) if those errors where in another language. Therefore I am not sure if we should build this into fosite :/

vivshankar:
@aeneasr Thanks for your response.

However, I would like to point out a couple of things.

First, the OAuth 2.0 specification released in 2012 says -

```
error_description
         OPTIONAL.  Human-readable ASCII [USASCII] text providing
         additional information, used to assist the client developer in
         understanding the error that occurred.
         Values for the "error_description" parameter MUST NOT include
         characters outside the set %x20-21 / %x23-5B / %x5D-7E.
```

It does not say that the error_description should be in English only. It also does not mandate what that error description text should be.

Second, the proposal here is to allow an optional language bundle directory to be provided by a consumer of fosite. If the language bundle isn't provided or locale-specific files are unavailable, you would fall back to the current message. So there would be no conformance violation.

Third, given that OAuth and Open ID Connect is basically used everywhere in the world now, it doesn't seem right to not provide an option to translate the error description, as desired.

I am happy to make a contribution if that would help.

aeneasr:
I was specifically referring to the section right above the cited one: https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1

These error messages are checked during certification (we have gone through certification several times) and even small typos can lead to failing certification!

If we were to add translation I would highly recommend to go with a well developed library (maybe you know one?). Alternatively writing your own error writer could also be an option!

vivshankar:
[...]

PR #617, opened 2021-08-30
[#615] Globalization

vivshankar:
The changes are proposed in #615 .

The changes involved:

- `MessageCatalog` interface exposes two funcs that can be used to choose an implementation of choice. The `DefaultMessageCatalog` provides a sample implementation.
- By default, `MessageCatalog` is nil and the implementation falls back to current behavior (tested by running the unit tests)
- Moved all free-text hints into error_hints.go and used `WithHintID` in place of `WithHintf` and `WithHint`. This should be the recommended approach for adding hints going forward.

There are no Breaking Changes.

## Related issue(s)

Linked to PR.

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).


vivshankar:
@aeneasr The lint errors are false positives: `G101: Potential hardcoded credentials`

vivshankar:
@aeneasr When you get a chance, could you please validate this? I am looking at closing this out and following this up with fixing any new `WithHint` and am concerned about the delta number growing too tedious to modify.

errors.go
```diff
@@ -499,3 +545,30 @@ func (e *RFC6749Error) ToValues() url.Values {
 
 	return values
 }
+
+func (e *RFC6749Error) getDefaultHint(ID i18n.ErrorHintType, args ...interface{}) string {
+	defHint, ok := i18n.HintsMap[ID]
+	if !ok {
+		// This should never happen if the hints are properly populated in the
+		// map.
+		panic(fmt.Sprintf("Hint must be populated in the map for: %s", ID))
```
aeneasr, line 554: I would much rather prefer not to panic here but instead return a generic error message

handler/pkce/handler.go
```diff
@@ -116,12 +117,12 @@ func (c *Handler) validate(challenge, method string, client fosite.Client) error
 	case "":
 		if !c.EnablePlainChallengeMethod {
 			return errorsx.WithStack(fosite.ErrInvalidRequest.
-				WithHint("Clients must use code_challenge_method=S256, plain is not allowed.").
+				WithHintID(i18n.ErrHintInvalidPKCECodeChallengeMethod, "S256").
```
aeneasr, line 120: This is no longer `WithHint("Clients must use code_challenge_method=S256, plain is not allowed.").`

handler/pkce/handler.go
```diff
@@ -116,12 +117,12 @@ func (c *Handler) validate(challenge, method string, client fosite.Client) error
...
 				WithDebug("The server is configured in a way that enforces PKCE S256 as challenge method for clients."))
 		}
 	default:
 		return errorsx.WithStack(fosite.ErrInvalidRequest.
-			WithHint("The code_challenge_method is not supported, use S256 instead."))
+			WithHintID(i18n.ErrHintInvalidPKCECodeChallengeMethod, "S256"))
```
aeneasr, line 125: This is no longer `WithHint("The code_challenge_method is not supported, use S256 instead."))`

handler/rfc7523/handler.go
```diff
@@ -239,43 +240,43 @@ func (c *Handler) findPublicKeyForToken(ctx context.Context, token *jwt.JSONWebT
...
+			WithHintID(i18n.ErrHintInvalidClientAssertionAudienceDuplicate),
 		)
 	}
 
 	if !claims.Audience.Contains(c.TokenURL) {
 		return errorsx.WithStack(fosite.ErrInvalidGrant.
-			WithHintf(
-				"The JWT in \"assertion\" request parameter MUST contain an \"aud\" (audience) claim containing a value \"%s\" that identifies the authorization server as an intended audience.",
+			WithHintID(
+				i18n.ErrHintInvalidClientAssertionAudience,
```
aeneasr, line 250: This message changed

aeneasr (changes requested):
Unfortunately I found several messages where the contents have changed. It's pretty tough to review because I have to look up every reference manually. Is there a way we can improve the process?

My requirement for this PR is that all messages stay the same unless it's a cosmetic typo or grammatical fix

errors.go
vivshankar: Addessed

handler/pkce/handler.go
vivshankar: Addressed

handler/pkce/handler.go
vivshankar: Addressed

handler/rfc7523/handler.go
vivshankar: Addressed

vivshankar:
@aeneasr I agree it is tedious to review this PR. Unfortunately, I can't think of a better way to review given this is the first big shift to globalization, other than check every message. If you prefer, I can break this down into multiple pull requests but the effort overall is going to remain the same. Also, as much as I can, I have kept to the same messages but there are cases where the message is subtly different (for example, the hint that indicates the algorithm expected for the signed JWT) and those I have combined to keep the number of messages sane.

aeneasr:
Ok, looks like I just need to find the time and focus to go through every one individually then 😅 I'll try to do another pass next week. I've reviewed about 40% so far but need to redo that because I was sort of sloppy during review.

PR #620, opened 2021-09-14
Fix: Only check verifier isn't empty when determining whether to do PKCE auth and ignore challenge

edorsey:
I was attempting to use oryd/hydra with a [Matrix Synapse](https://github.com/matrix-org/synapse) as well as [Grafana](https://github.com/grafana/grafana).

These projects implement server-side auth, but do not follow [this recommendation](https://github.com/ory/hydra/issues/1512#issuecomment-520413192):

>Note: although PKCE so far was recommended as a mechanism to protect
native apps, this advice applies to all kinds of OAuth clients,
including web applications.

However, they do include a "challenge" parameter, which appears to be included to try to maximize compatibility with generic OAuth.

This would cause the code below this logic to attempt PKCE verification, even though PKCE is not enforced or expected to be by the client.

Figured out that Synapse uses [`authlib`](https://github.com/lepture/authlib) underneath, which means it probably represents quite a few apps.

This change makes this logic only consider `verifier`, and ignores `challenge` in order to address this.

## Related issue(s)

I think this discussion may have actually been related to this -> https://github.com/ory/hydra/discussions/2671

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

### DIFF
--- a/handler/pkce/handler.go
+++ b/handler/pkce/handler.go
@@ -158,8 +158,8 @@ func (c *Handler) HandleTokenEndpointRequest(ctx context.Context, request fosite
 	if err := c.validate(challenge, method, client); err != nil {
 		return err
 	}
-
-	if !c.Force && challenge == "" && verifier == "" {
+	
+	if !c.Force && verifier == "" {
 		return nil
 	}
 


aeneasr:
This doesn’t look correct to me. Could you maybe first open an issue with all the information you have gathered so far? I do not believe that it is possible to perform PKCE without those two values involved.

edorsey:
> This doesn’t look correct to me. Could you maybe first open an issue with all the information you have gathered so far? I do not believe that it is possible to perform PKCE without those two values involved.

Yes, I believe you're correct, both values are needed for PKCE. This does not contradict that.

These are private clients that do not implement PKCE, but instead use `client_secret_basic`. It shouldn't be performing PKCE for these clients and as they do not support it, but because they are providing a parameter named `challenge` it is attempting PKCE and failing.

When it fails, it tells you "The PKCE code verifier must be at least 43 characters.". That's because the `verifier` is 0 characters long, but challenge is not an empty string.

I'll try to get around to re-producing the problem again in order to better document it, if you're still not convinced.

aeneasr:
I don't think that is the case. PKCE requires the `code_challenge` and `code_verifier` parameters to be available (see [RFC](https://datatracker.ietf.org/doc/html/rfc7636#section-4.2)). `challenge` itself is not a valid parameter and will be ignored by Ory Hydra / Ory Fosite! Is it possible that you have PKCE enforced and your clients are simply not doing PKCE requests?

edorsey:
> I don't think that is the case. PKCE requires the `code_challenge` and `code_verifier` parameters to be available (see [RFC](https://datatracker.ietf.org/doc/html/rfc7636#section-4.2)). `challenge` itself is not a valid parameter and will be ignored by Ory Hydra / Ory Fosite! Is it possible that you have PKCE enforced and your clients are simply not doing PKCE requests?

Ah, yes, getting the variable name confused with query parameter name.

I very well might have something misconfigured, review the logs during the problem with fresh eyes is probably a good idea.

I intend to get the logs together this Sunday. Feel free to close this until I’ve provided the logs though.

aeneasr:
Ok, closing this in that case because I think the change does not make sense as is, which is also why the CI is failing :)

edorsey:
> I don't think that is the case. PKCE requires the `code_challenge` and `code_verifier` parameters to be available (see [RFC](https://datatracker.ietf.org/doc/html/rfc7636#section-4.2)). `challenge` itself is not a valid parameter and will be ignored by Ory Hydra / Ory Fosite! Is it possible that you have PKCE enforced and your clients are simply not doing PKCE requests?

Finally circling back. I was adding the errant `code_challenge` on my end. Sorry for the busy work. I appreciate the prompt response and your patience with my hasty contribution. ✌️

Issue #255, opened 2018-03-31
No ability to keep refresh tokens after use


taylornz:
Hello,

I am using Hydra with ifttt.com who have a particular requirement on refresh tokens

From the IFTTT platform site: "Please note that refresh tokens cannot have a time-based expiry. The only time it is technically permissible for a refresh token to expire is after an access token has been refreshed. At that time, it is acceptable to return a new refresh token; however, **we require that the previous refresh token not immediately expire.**

This is something that is part of the IFTTT oauth connection test, which fails against a standard installation of hydra.

I can build a custom copy of hydra with a patched version of fosite/handler/oauth2/flow_refresh.go where the call to TokenRevocationStorage.RevokeRefreshToken is removed, however keeping a custom compile of hydra is less than ideal.

Is there a sensible way to make something like this configurable that would be acceptable for an upstream merge into fosite? perhaps a special scope name on requester.GetGrantedScopes() eg "delay_refresh_token_delete" or "keep_refresh_token" ?

https://tools.ietf.org/html/rfc6819#section-5.2.2.3 does mention regarding Refresh Token Rotation : Note: This measure may cause problems in clustered environments, since usage of the currently valid refresh token must be ensured.  In such an environment, other measures might be more appropriate. This must be why IFTTT require delayed  deletion ( or no deletion ) of the refresh token.

I would very much appreciate any guidance that could be provided with a view to providing a pull request for evaluation.

Thank you.

arekkas:
Thank you for the clear write up. The idea if revoking RT immediately is to prevent replay attacks and potential issuance of more than one new refresh tokens. That’s why this is disabled at the moment.

Does the test require a second acces token, or just no error message?

taylornz:
Thank you very much for taking the time to reply.

I understand that the way Hydra/Fosite is currently set up is ideal from a security perspective and that my issue is a relatively unique restraint of the IFTTT.com platform.

The IFTTT connection test performs the following:

1: Retrieve authorization code ( Currently from Hydra )
-Code was provided
-State matches original original state from request

2: Request access token ( Currently from Hydra )
-Status Code is 200
-Body is valid JSON
-Response is an object
-Access token is provided
-Access token is a string
-Access token is at least one character long
-Refresh token is provided
-Refresh token is a string
-Refresh token is at least one character long

3: Retrieve user information ( A custom endpoint which proxies to Hydra introspection )
Status Code is 200
Body is valid JSON
Response has one top-level "data" object
"name" is a string
[...]

PR #632, opened 2021-10-21
feat: allow for permanent refresh_token(s)

bill-robbins-ss:
This is a branch of `v0.40.2` that introduces the ability to disable one-time-use refresh tokens (RT). From my reading of issues here in the fosite repository and the Hydra repository, a number of users desire a feature like this. My hope is that this change can be released with a new version `v0.40.3` and a new version of Hydra `v1.10.7`. With these new releases I can update my deployment of Hydra to pick up the new feature. 

One time use refresh-tokens can introduce issues where a user's "session" stored in a 3rd party database can be completely disabled, forcing the end-user to go through an OAuth2 flow with Hydra again. The related issues I have included below outline some of the ways this can happen and include some discussion. 

For brevity here is a use case involving threads that can permanently disable a session:

1. Execute an OAuth2 flow with Hydra and receive an `access_token` (AT-0) and `refresh_token` (RT-0). 
2. Thread-A: (AT-0 has expired) use RT-0 to obtain new tokens, AT-1, RT-1
3. Thread-B: (AT-0 has expired) _also_ uses RT-0 to obtain a new RT, but this fails because Thread-1 has already used the token
4. Thread-A: uses AT-1 to call an API endpoint (ie; `/userinfo`)
5. Thread-A: (AT-1 has expired) use RT-1 to obtain new tokens. This **fails** because in step 3, Thread-B attempted to use RT-0 after it was used. There is now no way to get back into a functional state and the end-user must go through the flow again.

This can be worked-around by implementing a distributed-lock pattern in the client to Hydra, but I believe this places an undue burden on the client developers. This burden is exacerbated by the nature of implementing OAuth2 flows for an application: each client will likely be an independent 3rd party entity to the entity hosting Hydra. Further, a distributed-locking pattern can not account for the other failure cases, such as network errors.

Note: There is a companion [PR in Hydra](https://github.com/ory/hydra/pull/2815) that adds a configuration option to enable/disable this behavior.

## Related issue(s)

#255 
[Hydra #1831](https://github.com/ory/hydra/issues/1831)

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

## Further comments

I implemented this change with the goal of changing as little as possible, hence the use of `Context` to pass information from Hydra to this library. I have seen no other mechanism for accessing configuration provided in Hydra to fosite, nor is the Strategy Pattern possible here, as `HandleTokenEndpointRequest` does not use a strategy for revoking a token. Instead `HandleTokenEndpointRequest` accesses `TokenRevocationStorage` directly, meaning that this behavior is applied to every Strategy involved. A deeper change could be made to introduce a higher-level strategy for handling token refresh, or introducing a TokenRevocation strategy but this is in conflict with my goal of making a small, precise, change.

### DIFF
--- a/context.go
+++ b/context.go
@@ -35,5 +35,5 @@ const (
 	AccessResponseContextKey    = ContextKey("accessResponse")
 	AuthorizeRequestContextKey  = ContextKey("authorizeRequest")
 	AuthorizeResponseContextKey = ContextKey("authorizeResponse")
-	EnableOneTimeUseRefreshTokenContextKey  = ContextKey("enableOneTimeUseRefreshToken")
+	RefreshTokenRotationEnabledContextKey  = ContextKey("refreshTokenRotationEnabled")
 )
--- a/context.go
+++ b/context.go
@@ -35,4 +35,5 @@ const (
 	AccessResponseContextKey    = ContextKey("accessResponse")
 	AuthorizeRequestContextKey  = ContextKey("authorizeRequest")
 	AuthorizeResponseContextKey = ContextKey("authorizeResponse")
+	EnableOneTimeUseRefreshTokenContextKey  = ContextKey("enableOneTimeUseRefreshToken")
 )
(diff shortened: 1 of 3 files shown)


bill-robbins-ss:
closing in favor of refresh token grace period

PR #638, opened 2021-11-24
fix: always rollback

### LINKED ISSUE #637: Rollback is not called if there is a panic
### Preflight checklist

- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).

### Describe the bug

Current use of `BeginTx`, `CommitTx` and `RollbackTx` in Fosite is problematic because it does not call `RollbackTx` when panic happens. We could assume that Fosite itself does not panic, but that is not necessary true for code it calls into. For example, I had a bug inside my storage implementation which triggered panic. As the panic propagated and was handled higher up in my request handler and logged, transaction itself was not rolled back, leaving the connection over it in an inconsistent state. With connection pooling that brought some stability issues.

### Reproducing the bug

Sadly I do not have a simple reproduction here. One which I was able to do is with Sqlite and setting `SetMaxOpenConns` to 1. Then if a transaction is not rolled back, no other operation can be done on the database and my code hangs. This is how I am now testing that all transactions are properly rolled back.

### Relevant log output

_No response_

### Relevant configuration

_No response_

### Version

observed on current master branch

_No response_

_No response_

### Additional Context

_No response_

mitar:

Fixes #637.

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

## Further comments

There is not much change necessary to get this fixed. See the diff.

There is one logical change I have done: before rollback was not called if commit failed, now I made it that it is called if commit fails. It is generally safe to call rollback outside of a transaction in PostgreSQL (you only get warning), but if the connection is in error state, then rollback is required. So just to be safe, I think it is better to rollback on any error, including when commit fails (because we might not know what is the reason for failure of that call and in what state is the connection).


handler/oauth2/flow_authorize_code_token.go
```diff
@@ -170,22 +170,20 @@ func (c *AuthorizeExplicitGrantHandler) PopulateTokenEndpointResponse(ctx contex
...
-	if err := c.CoreStorage.InvalidateAuthorizeCodeSession(ctx, signature); err != nil {
-		if rollBackTxnErr := storage.MaybeRollbackTx(ctx, c.CoreStorage); rollBackTxnErr != nil {
-			return errorsx.WithStack(fosite.ErrServerError.WithWrap(err).WithDebugf("error: %s; rollback error: %s", err, rollBackTxnErr))
+	defer func() {
+		if err != nil {
+			if rollBackTxnErr := storage.MaybeRollbackTx(ctx, c.CoreStorage); rollBackTxnErr != nil {
+				err = errorsx.WithStack(fosite.ErrServerError.WithWrap(err).WithDebugf("error: %s; rollback error: %s", err, rollBackTxnErr))
+			}
 		}
+	}()
```
aeneasr, line 179: This wouldn't roll back on a panic, it rewires the logic from explicit rollback to implicit via defer. A panic would need a recover on the top level: https://go.dev/blog/defer-panic-and-recover

mitar: Of course it would. From the link you posted itself (emphasis mine):

> Panic is a built-in function that stops the ordinary flow of control and begins panicking. When the function F calls panic, execution of F stops, **any deferred functions in F are executed normally**, and then F returns to its caller. To the caller, F then behaves like a call to panic. The process continues up the stack until all functions in the current goroutine have returned, at which point the program crashes.

aeneasr: In German we say: "Wer lesen kann ist klar im Vorteil" (loosely translated to "Who can read has a clear advantage"). You're absolutely right, this does indeed catch any panic happening!

handler/oauth2/flow_authorize_code_token.go
```diff
@@ -198,7 +196,7 @@ func (c *AuthorizeExplicitGrantHandler) PopulateTokenEndpointResponse(ctx contex
 		responder.SetExtra("refresh_token", refresh)
 	}
 
-	if err := storage.MaybeCommitTx(ctx, c.CoreStorage); err != nil {
+	if err = storage.MaybeCommitTx(ctx, c.CoreStorage); err != nil {
```
aeneasr, line 199: One problem I have observed with this control flow is that RDBMS roll back transactions automatically when commits fail. When rolling back a failed transaction, we will always end up with an error because there is nothing to roll back any more. This in turn causes the root of the issue (e.g. duplicate key) to be quite difficult to find/debug, because it is hidden in the rollback error and is probably missing content.

We had this issue in Hydra, the fix is to just return commit errors normally.

handler/oauth2/flow_authorize_code_token.go
mitar: :-)

handler/oauth2/flow_authorize_code_token.go
mitar: > One problem I have observed with this control flow is that RDBMS roll back transactions automatically when commits fail.

Yes, I note this (and rationale for this change) in the description of the PR:

> There is one logical change I have done: before rollback was not called if commit failed, now I made it that it is called if commit fails. It is generally safe to call rollback outside of a transaction in PostgreSQL (you only get warning), but if the connection is in error state, then rollback is required. So just to be safe, I think it is better to rollback on any error, including when commit fails (because we might not know what is the reason for failure of that call and in what state is the connection).

> When rolling back a failed transaction, we will always end up with an error because there is nothing to roll back any more.

Hm, at least in [PostgreSQL you do not get an error](https://www.postgresql.org/docs/devel/sql-rollback.html):

> Issuing `ROLLBACK` outside of a transaction block emits a warning and otherwise has no effect.

With which database have you noticed the error?

> This in turn causes the root of the issue (e.g. duplicate key) to be quite difficult to find/debug, because it is hidden in the rollback error and is probably missing context.

I think this can be fixed a bit. So regular error looks like:

`errorsx.WithStack(fosite.ErrServerError.WithWrap(storageErr).WithDebug(storageErr.Error()))`

Current rollback error hides the regular error completely:

`err = errorsx.WithStack(fosite.ErrServerError.WithWrap(rbErr).WithDebug(rbErr.Error()))`

But in `PopulateTokenEndpointResponse` we have in fact a precedence which I think addresses the issue:

`errorsx.WithStack(fosite.ErrServerError.WithWrap(err).WithDebugf("error: %s; rollback error: %s", err, rollBackTxnErr))`

I think I could adopt this for rollback error everywhere. So:

`errorsx.WithStack(fosite.ErrServerError.WithWrap(storageErr).WithDebugf("error: %s; rollback error: %s", storageErr, rbErr))`

aeneasr: > With which database have you noticed the error?

I don't quite remember, I think it was either MySQL or CockroachDB :/

> I think I could adopt this for rollback error everywhere. So:

SGTM! :)

aeneasr:
Are you still up for the changes? :monocle_face: If you need any help, let us know!

I'll convert this to a draft in the meanwhile. When it's ready for review again, please remove the draft indicator! 👀

mitar:
Yes yes. I was just a bit busy with other things.

aeneasr:
All good. I think I am known by now as one of the slowest maintainers in open source 😓

handler/oauth2/flow_authorize_code_token.go
mitar: Done.

mitar:
> All good. I think I am known by now as one of the slowest maintainers in open source

What? No way. You are one of best maintainers I have seen around. Regularly checking in, always giving great feedback, pushing PRs through and not leaving them stick around.

aeneasr:
> What? No way. You are one of best maintainers I have seen around. Regularly checking in, always giving great feedback, pushing PRs through and not leaving them stick around.

You made my day, thank you :)

mitar:
BTW, those tests are failing also in `master` branch.

mitar:
ping @aeneasr

aeneasr:
Ah yes it fails because the refresh PR has not yet been merged. I'd like for the OIDC conformity siute to pass first so that we can confirm that the rollback has no unintended effects :)

mitar:
Which other PR?

aeneasr:
i merged a workaround, let's see if we're green!

PR #643, opened 2022-01-15
feat: Add support for error details

beornf:
Add support for adding details to RFC6749 errors.

## Checklist

      and signed the CLA.
- [ ] I have referenced an issue containing the design document if my change
      introduces a new feature.
      the changes.
      works.
      appropriate).

## Further comments

Adding details to custom errors makes it possible to return necessary context data along with the error (e.g. user ID, MFA flows).

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -266,6 +266,7 @@ type (
 		HintField        string
 		CodeField        int
 		DebugField       string
+		DetailsField     map[string]interface{}
 		cause            error
 		useLegacyFormat  bool
 		exposeDebug      bool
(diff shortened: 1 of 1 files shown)


aeneasr (changes requested):
Thank you for the PR! What would be the reason for adding this field? It does not seem to be used in the library anywhere and as we always say: When in doubt, leave it out :)

beornf:
I'm using an error handler on RFC6749Error and also processing my own RFC6749Error with custom data. I saw the `DetailsCarrier` interface was commented but can try putting data in the error some other way.

aeneasr:
Got it! Thank you for the explanation! I think you should be able to achieve the same by embedding the error in a custom struct in your code. The codebase here is already pretty complex and I’d like to keep it as focused as possible :)

beornf:
Sure thanks for the tip I'll close this PR then.

PR #653, opened 2022-02-24
feat: Enable urn:ietf:params:oauth:grant-type:jwt-bearer for unauthenticated Clients

akrause-cel:
This PR allows the `urn:ietf:params:oauth:grant-type:jwt-bearer` grant type to work with public clients (`token_endpoint_auth_method` none). The issuer is used to derive the client ID that is used to verify whether client is indeed public or not.

This change is required to reach compatibility with Salesforce Named Credentials with Authentication Method `WT Token Exchange`. For further Details see:

* [Named Credential](https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_callouts_named_credentials.htm)
* [Sample](https://ambassadorpatryk.com/2020/08/jwt-validation-policy-and-salesforce-named-credentials/)
* [APEX Class](https://developer.salesforce.com/docs/atlas.en-us.apexref.meta/apexref/apex_class_Auth_JWTBearerTokenExchange.htm) 

It is also backed by [rfc7523](https://datatracker.ietf.org/doc/html/rfc7523#section-2.1), that says that Authentication of the client is optional.

This is not a breaking change. Also security is not impacted as clients will need to be made public explicitly.

This is not a major design change. All it does is check whether an assertion passed by an unauthenticated client (no client credentials in token request) in the context of an `urn:ietf:params:oauth:grant-type:jwt-bearer` grant resembles a registered and public OAuth2 client by extracting the issuer field.

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

## Further comments

In order to be useful for me and my project this change needs to be built into hydra. However hydra is not affected by any of the changes and continues to work as expected.

Below is a curl representation of a Salesforce Request:

```
curl -X 'POST' 'https://webhook.site/' -H 'connection: close' -H 'content-length: 839' -H 'content-type: application/x-www-form-urlencoded' -H 'accept: text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2' -H 'host: webhook.site' -H 'pragma: no-cache' -H 'cache-control: no-cache' -H 'user-agent: SFDC-Callout/54.0' -d $'grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Ajwt-bearer&assertion=eyJraWQiOiJKV1RfQ0VSVCIsInR5cCI6IkpXVCIsImFsZyI6IlJTMjU2In0.eyJpc3MiOiI1ZWQ4MTJkYy05MzllLTExZWMtYjkwOS0wMjQyYWMxMjAwMDQiLCJzdWIiOiI1ZWQ4MTJkYy05MzllLTExZWMtYjkwOS0wMjQyYWMxMjAwMDQiLCJhdWQiOlsiaHR0cHM6Ly9oeWRyYS5jZWxvbmlzbGFicy5pby9vYXV0aDIvdG9rZW4iLCJ0ZWFtLWxhYnMiLCJodHRwOi8vbG9jYWxob3N0OjQ0NDQvb2F1dGgyL3Rva2VuIl0sIm5iZiI6MTY0NTY4NTA4NSwiaWF0IjoxNjQ1Njg1MDg1LCJleHAiOjE2NDU3NzE0ODV9.W_WaaBqn4XHgpB0JoVoN9lwSiZtic-hpRP24JF4OgRiIXUA4wqM7iDyq__OaFPRoYRZu1hWZzTk-i5bOB05Se5opmB_JOnaz2c7G3953sod9GOmiQdLdwHr2ZQ9G-ry6bWeieRMvWLtQ3TjrUjyycyZ-EKdz0aosJRhIqtR7VA7c0Q8wiLLifOOX5Wg0t4fRGLyN16dG2ME1KTyx3BPB65tJirXVqE0a4XGx6wDxL_00Z3YEgxWpylCULE62eRFt35l3VizwG3yp_cQZpBM2MJL5VpJMXT_-zKT_Lvpj81VM1l2s6Jm0vyrC2TbmBfS1Rbkuv-fgSxrimtjPUUEDaw&scope=pycelonis-pack.pycelonis-model'
```

### DIFF
--- a/client_authentication.go
+++ b/client_authentication.go
@@ -227,38 +227,10 @@ func (f *Fosite) DefaultClientAuthenticationStrategy(ctx context.Context, r *htt
 	if err != nil {
 		// check if grant type is urn:ietf:params:oauth:grant-type:jwt-bearer
 		// if yes, try to derive client ID from JWT Bearer token ("iss"). This is needed as certain providers
-		// (e.g. Salesforce) do not provide the means supply client credentials with the bearer token)
-		if grantType := form.Get("grant_type"); grantType == grantTypeJWTBearer {
-			assertion := form.Get("assertion")
-
-			if assertion == "" {
-				return nil, err
-			}
-			// do not validate the key (hence noop token func), all this is just to get the issuer as client ID
-			token, err := josejwt.ParseSigned(assertion)
-			// if there is no assertion is not a jwt (signature ignored), then authentication of the client has obviously failed
-			if err != nil {
-				return nil, err
-			}
-
-			// parse claims without looking at token
-			claims := jwt.MapClaims{}
-			if err := token.UnsafeClaimsWithoutVerification(&claims); err != nil {
-				// if claims can't be parsed, well then something is wrong
-				return nil, err
-			}
-
-			tokenIssuer, ok := claims["iss"]
-			if !ok {
-				// token does not contain an issuer, hence authentication of client has obviously failed
-				return nil, err
-			}
-			clientID, ok = tokenIssuer.(string)
-			if !ok {
-				// token does not contain a valid issuer, hence authentication of client has obviously failed
-				return nil, err
-			}
-		} else {
+		// (e.g. Salesforce) do not provide the means to supply client credentials with the bearer token.
+		// this is also covered by https://datatracker.ietf.org/doc/html/rfc7523#section-2.1
+		// that says that Authentication of the client is optional
+		if clientID, _, err = jwtBearerNoClientAuthRequest(r, form, err); err != nil {
 			return nil, err
 		}
 	}
@@ -288,6 +260,43 @@ func (f *Fosite) DefaultClientAuthenticationStrategy(ctx context.Context, r *htt
 	}
 
 	return client, nil
+}
+
+// Tries to read client ID from (yet) unverified token, if request came unauthenticated
+func jwtBearerNoClientAuthRequest(r *http.Request, form url.Values, originalErr error) (clientID, clientSecret string, err error) {
+	if grantType := form.Get("grant_type"); grantType != grantTypeJWTBearer {
+		return "", "", originalErr
+	}
+	assertion := form.Get("assertion")
+
+	if assertion == "" {
+		return "", "", errorsx.WithStack(ErrInvalidRequest.WithHint("Required parameter assertion is missing."))
+	}
+	// do not validate the key (hence noop token func), all this is just to get the issuer as client ID
+	token, err := josejwt.ParseSigned(assertion)
+	// if there is no assertion is not a jwt (signature ignored), then authentication of the client has obviously failed
+	if err != nil {
+		return "", "", errorsx.WithStack(ErrInvalidRequest.WithHint("JWT in assertion can't be parsed."))
+	}
+
+	// parse claims without looking at token
+	claims := jwt.MapClaims{}
+	if err := token.UnsafeClaimsWithoutVerification(&claims); err != nil {
+		// if claims can't be parsed, well then something is wrong
+		return "", "", errorsx.WithStack(ErrInvalidRequest.WithHint("Claims in JWT in assertion can't be parsed"))
+	}
+
+	tokenIssuer, ok := claims["iss"]
+	if !ok {
+		// token does not contain an issuer, hence authentication of client has obviously failed
+		return "", "", errorsx.WithStack(ErrInvalidRequest.WithHint("Unauthenticated request with JWT assertion does not contain issuer [\"iss\"] that is used to derive client ID."))
+	}
+	clientID, ok = tokenIssuer.(string)
+	if !ok {
+		// token does not contain a valid issuer, hence authentication of client has obviously failed
+		return "", "", errorsx.WithStack(ErrInvalidRequest.WithHint("Unauthenticated request with JWT assertion does not contain valid issuer [\"iss\"]."))
+	}
+	return
 }
 
 func (f *Fosite) checkClientSecret(ctx context.Context, client Client, clientSecret []byte) error {
--- a/client_authentication.go
+++ b/client_authentication.go
@@ -270,33 +270,30 @@ func jwtBearerNoClientAuthRequest(r *http.Request, form url.Values, originalErr 
 	assertion := form.Get("assertion")
 
 	if assertion == "" {
-		return "", "", errorsx.WithStack(ErrInvalidRequest.WithHint("Required parameter assertion is missing."))
+		return "", "", originalErr
 	}
 	// do not validate the key (hence noop token func), all this is just to get the issuer as client ID
 	token, err := josejwt.ParseSigned(assertion)
 	// if there is no assertion is not a jwt (signature ignored), then authentication of the client has obviously failed
 	if err != nil {
-		return "", "", errorsx.WithStack(ErrInvalidRequest.WithHint("JWT in assertion can't be parsed."))
+		return "", "", errorsx.WithStack(ErrUnauthorizedClient.WithHint("JWT in assertion is invalid."))
 	}
 
 	// parse claims without looking at token
-	claims := jwt.MapClaims{}
+	claims := struct {
+		Issuer string `json:"iss"`
+	}{}
 	if err := token.UnsafeClaimsWithoutVerification(&claims); err != nil {
 		// if claims can't be parsed, well then something is wrong
-		return "", "", errorsx.WithStack(ErrInvalidRequest.WithHint("Claims in JWT in assertion can't be parsed"))
+		return "", "", errorsx.WithStack(ErrUnauthorizedClient.WithHint("Claims of JWT assertion in assertion can not be parsed."))
 	}
 
-	tokenIssuer, ok := claims["iss"]
-	if !ok {
+	if claims.Issuer == "" {
 		// token does not contain an issuer, hence authentication of client has obviously failed
-		return "", "", errorsx.WithStack(ErrInvalidRequest.WithHint("Unauthenticated request with JWT assertion does not contain issuer [\"iss\"] that is used to derive client ID."))
+		return "", "", errorsx.WithStack(ErrUnauthorizedClient.WithHint("Unauthenticated request with JWT assertion does not contain issuer [\"iss\"] that is used to derive client ID."))
 	}
-	clientID, ok = tokenIssuer.(string)
-	if !ok {
-		// token does not contain a valid issuer, hence authentication of client has obviously failed
-		return "", "", errorsx.WithStack(ErrInvalidRequest.WithHint("Unauthenticated request with JWT assertion does not contain valid issuer [\"iss\"]."))
-	}
-	return
+
+	return claims.Issuer, "", nil
 }
 
 func (f *Fosite) checkClientSecret(ctx context.Context, client Client, clientSecret []byte) error {
--- a/client_authentication.go
+++ b/client_authentication.go
@@ -36,12 +36,16 @@ import (
 	"github.com/ory/fosite/token/jwt"
 	"github.com/pkg/errors"
 	jose "gopkg.in/square/go-jose.v2"
+	josejwt "gopkg.in/square/go-jose.v2/jwt"
 )
 
 // ClientAuthenticationStrategy provides a method signature for authenticating a client request
 type ClientAuthenticationStrategy func(context.Context, *http.Request, url.Values) (Client, error)
 
-const clientAssertionJWTBearerType = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
+const (
+	clientAssertionJWTBearerType = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
+	grantTypeJWTBearer           = "urn:ietf:params:oauth:grant-type:jwt-bearer"
+)
 
 func (f *Fosite) findClientPublicJWK(oidcClient OpenIDConnectClient, t *jwt.Token, expectsRSAKey bool) (interface{}, error) {
 	if set := oidcClient.GetJSONWebKeys(); set != nil {
@@ -221,7 +225,42 @@ func (f *Fosite) DefaultClientAuthenticationStrategy(ctx context.Context, r *htt
 
 	clientID, clientSecret, err := clientCredentialsFromRequest(r, form)
 	if err != nil {
-		return nil, err
+		// check if grant type is urn:ietf:params:oauth:grant-type:jwt-bearer
+		// if yes, try to derive client ID from JWT Bearer token ("iss"). This is needed as certain providers
+		// (e.g. Salesforce) do not provide the means supply client credentials with the bearer token)
+		if grantType := form.Get("grant_type"); grantType == grantTypeJWTBearer {
+			assertion := form.Get("assertion")
+
+			if assertion == "" {
+				return nil, err
+			}
+			// do not validate the key (hence noop token func), all this is just to get the issuer as client ID
+			token, err := josejwt.ParseSigned(assertion)
+			// if there is no assertion is not a jwt (signature ignored), then authentication of the client has obviously failed
+			if err != nil {
+				return nil, err
+			}
+
+			// parse claims without looking at token
+			claims := jwt.MapClaims{}
+			if err := token.UnsafeClaimsWithoutVerification(&claims); err != nil {
+				// if claims can't be parsed, well then something is wrong
+				return nil, err
+			}
+
+			tokenIssuer, ok := claims["iss"]
+			if !ok {
+				// token does not contain an issuer, hence authentication of client has obviously failed
+				return nil, err
+			}
+			clientID, ok = tokenIssuer.(string)
+			if !ok {
+				// token does not contain a valid issuer, hence authentication of client has obviously failed
+				return nil, err
+			}
+		} else {
+			return nil, err
+		}
 	}
 
 	client, err := f.Store.GetClient(ctx, clientID)
--- a/client_authentication_test.go
+++ b/client_authentication_test.go
@@ -87,6 +87,7 @@ func clientBasicAuthHeader(clientID, clientSecret string) http.Header {
 
 func TestAuthenticateClient(t *testing.T) {
 	const at = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"
+	const gtJWT = "urn:ietf:params:oauth:grant-type:jwt-bearer"
 
 	hasher := &BCrypt{WorkFactor: 6}
 	f := &Fosite{
@@ -506,6 +507,31 @@ func TestAuthenticateClient(t *testing.T) {
 				"jti": "12345",
 				"aud": "not-token-url",
 			}, rsaKey, "kid-foo")}, "client_assertion_type": []string{at}},
+			r:         new(http.Request),
+			expectErr: ErrInvalidClient,
+		},
+		{
+			d:      "should pass because assertion contains client id as issuer and client is public",
+			client: &DefaultOpenIDConnectClient{DefaultClient: &DefaultClient{ID: "bar", Public: true}, JSONWebKeys: rsaJwks, TokenEndpointAuthMethod: "none"},
+			form: url.Values{"assertion": {mustGenerateRSAAssertion(t, jwt.MapClaims{
+				"sub": "bar",
+				"exp": time.Now().Add(time.Hour).Unix(),
+				"iss": "bar",
+				"jti": "12345",
+				"aud": "token-url",
+			}, rsaKey, "kid-foo")}, "grant_type": []string{gtJWT}},
+			r: new(http.Request),
+		},
+		{
+			d:      "should fail despite valid assertion because client is private",
+			client: &DefaultOpenIDConnectClient{DefaultClient: &DefaultClient{ID: "bar", Secret: barSecret, Public: false}, JSONWebKeys: rsaJwks, TokenEndpointAuthMethod: "none"},
+			form: url.Values{"assertion": {mustGenerateRSAAssertion(t, jwt.MapClaims{
+				"sub": "bar",
+				"exp": time.Now().Add(time.Hour).Unix(),
+				"iss": "bar",
+				"jti": "12345",
+				"aud": "token-url",
+			}, rsaKey, "kid-foo")}, "grant_type": []string{gtJWT}},
 			r:         new(http.Request),
 			expectErr: ErrInvalidClient,
 		},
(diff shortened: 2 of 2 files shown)


client_authentication.go
```diff
@@ -251,6 +262,43 @@ func (f *Fosite) DefaultClientAuthenticationStrategy(ctx context.Context, r *htt
...
 
+// Tries to read client ID from (yet) unverified token, if request came unauthenticated
+func jwtBearerNoClientAuthRequest(r *http.Request, form url.Values, originalErr error) (clientID, clientSecret string, err error) {
+	if grantType := form.Get("grant_type"); grantType != grantTypeJWTBearer {
+		return "", "", originalErr
+	}
+	assertion := form.Get("assertion")
+
+	if assertion == "" {
+		return "", "", errorsx.WithStack(ErrInvalidRequest.WithHint("Required parameter assertion is missing."))
```
aeneasr, line 273: Please return the original error here

client_authentication.go
```diff
@@ -251,6 +262,43 @@ func (f *Fosite) DefaultClientAuthenticationStrategy(ctx context.Context, r *htt
...
+	}
+	// do not validate the key (hence noop token func), all this is just to get the issuer as client ID
+	token, err := josejwt.ParseSigned(assertion)
+	// if there is no assertion is not a jwt (signature ignored), then authentication of the client has obviously failed
+	if err != nil {
+		return "", "", errorsx.WithStack(ErrInvalidRequest.WithHint("JWT in assertion can't be parsed."))
```
aeneasr, line 279:
```suggestion
		return "", "", errorsx.WithStack(ErrUnauthorized.WithHint("JWT in assertion is invalid."))
```

client_authentication.go
```diff
@@ -251,6 +262,43 @@ func (f *Fosite) DefaultClientAuthenticationStrategy(ctx context.Context, r *htt
...
+	}
+
+	// parse claims without looking at token
+	claims := jwt.MapClaims{}
+	if err := token.UnsafeClaimsWithoutVerification(&claims); err != nil {
+		// if claims can't be parsed, well then something is wrong
+		return "", "", errorsx.WithStack(ErrInvalidRequest.WithHint("Claims in JWT in assertion can't be parsed"))
```
aeneasr, line 286:
```suggestion
		return "", "", errorsx.WithStack(ErrUnauthorized.WithHint("Claims of JWT assertion in assertion can not be parsed."))
```

client_authentication.go
```diff
@@ -251,6 +262,43 @@ func (f *Fosite) DefaultClientAuthenticationStrategy(ctx context.Context, r *htt
...
+	}
+	// do not validate the key (hence noop token func), all this is just to get the issuer as client ID
+	token, err := josejwt.ParseSigned(assertion)
+	// if there is no assertion is not a jwt (signature ignored), then authentication of the client has obviously failed
+	if err != nil {
+		return "", "", errorsx.WithStack(ErrInvalidRequest.WithHint("JWT in assertion can't be parsed."))
+	}
+
+	// parse claims without looking at token
+	claims := jwt.MapClaims{}
```
aeneasr, line 283: If you use a struct here you can save all the type assertion code later:

```suggestion
	claims := struct{Issuer string `json:"issuer"`}{}
```

client_authentication_test.go
```diff
@@ -509,6 +510,31 @@ func TestAuthenticateClient(t *testing.T) {
 			r:         new(http.Request),
 			expectErr: ErrInvalidClient,
 		},
+		{
+			d:      "should pass because assertion contains client id as issuer and client is public",
```
aeneasr, line 514: Can you please also add some failing tests?

aeneasr (changes requested):
Awesome, thank you for your contribution! This looks pretty good and I have some ideas how to improve it further :)

client_authentication_test.go
akrause-cel: I have added the following tests:

- Pass: should pass because assertion contains client id as issuer and client is public (Testing the added functionality)
- Fail: should fail despite valid assertion because client is private (testing whether behaviour is consistent with config)
- Fail: should fail despite valid assertion because client unknown (testing whether behaviour is consistent with config)

akrause-cel:
Thanks for your feedback and review. I hope after incorporating the changes it can be merged

client_authentication.go
```diff
@@ -251,6 +262,40 @@ func (f *Fosite) DefaultClientAuthenticationStrategy(ctx context.Context, r *htt
...
+func jwtBearerNoClientAuthRequest(r *http.Request, form url.Values, originalErr error) (clientID, clientSecret string, err error) {
+	if grantType := form.Get("grant_type"); grantType != grantTypeJWTBearer {
+		return "", "", originalErr
+	}
+	assertion := form.Get("assertion")
+
+	if assertion == "" {
+		return "", "", originalErr
+	}
+	// do not validate the key (hence noop token func), all this is just to get the issuer as client ID
```
piotrmsc, line 275: Do we check if the assertion is a single JWT?

akrause-cel: Assertion is retrieved through form.Get("assertion") which gets the first value associated with the given key.

piotrmsc: I'm referring to RFC mentioning   `The value of the "assertion" parameter MUST contain a single JWT.`

akrause-cel: What is the concrete risk you see here? If form contains multiple assertions the first one is taken. If assertion parameter has multiple JWTs contained, the subsequent Parsing will fail. Hence I think we have sufficient assurance. However, if you insist I can check whether the form's assertion array has length 1 and if > 1 raise an error.

piotrmsc: there's no concrete risk, I'm getting familiar with code base here in general and at the same time I went through the RFC you have linked ;-)  I've looked into the entire file and see it's always done using form.Get() without checking if multiple assertions exist.

client_authentication.go
```diff
@@ -251,6 +262,40 @@ func (f *Fosite) DefaultClientAuthenticationStrategy(ctx context.Context, r *htt
...
+	token, err := josejwt.ParseSigned(assertion)
+	// if there is no assertion is not a jwt (signature ignored), then authentication of the client has obviously failed
+	if err != nil {
+		return "", "", errorsx.WithStack(ErrUnauthorizedClient.WithHint("JWT in assertion is invalid."))
+	}
+
+	// parse claims without looking at token
+	claims := struct {
+		Issuer string `json:"iss"`
```
aeneasr, line 284: Could you please back the assertion that the issuer is used for identifying the client in the RFC?

aeneasr: Yeah, shouldn't this be `sub`?

```
   2.   The JWT MUST contain a "sub" (subject) claim identifying the
        principal that is the subject of the JWT.  Two cases need to be
        differentiated:

        A.  For the authorization grant, the subject typically
            identifies an authorized accessor for which the access token
            is being requested (i.e., the resource owner or an
            authorized delegate), but in some cases, may be a
            pseudonymous identifier or other value denoting an anonymous
            user.

        B.  For client authentication, the subject MUST be the
            "client_id" of the OAuth client.
```

akrause-cel: @aeneasr : You are referring to Section 2.2. Which is all bout using a JWT for Client Authentication within an arbitrary OAuth flow (in the example `authorization_code `). 

This is is referring to Section 2.1 which defines a new grant type (`urn:ietf:params:oauth:grant-type:jwt-bearer`). 
This PR applies when client is unauthenticated. It helps to find out that this is permitted (client is configured to be public). From my PoV this is a grey area of the RFC because the RFC does not specify how to figure out whether a client must be authenticated. 

Also I believe it gives users all freedom to configure fosite (through hydra) for maximum compatibility.

aeneasr: I'm not 100% sure if I follow. Could you please point me to the RFC which explains that `issuer` should be used to identify the OAuth2 Client for the use case you describe?

akrause-cel: As said this is not described in [RFC](https://datatracker.ietf.org/doc/html/rfc7523#section-3.2.1).

> Authentication of the client is optional, as described in
>    [Section 3.2.1](https://datatracker.ietf.org/doc/html/rfc7523#section-3.2.1) of OAuth 2.0 [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)] and consequently, the
>    "client_id" is only needed when a form of client authentication that
>    relies on the parameter is used.

The open question is: How should the server know whether authentication for this particular client is optional? Here my proposal is to use the issuer.

aeneasr: And just to avoid misunderstanding, this looks very useful and sane to me. My reservation is just that we're using a field I think isn't supposed to be used in that way! But it makes sense to have client auth resolve the client ID somehow without relying on regular client auth.

---

https://datatracker.ietf.org/doc/html/rfc7523#section-3 does not seem to indicate that we can use the `iss` value as a substitute for the OAuth2 Client ID. It just says that the `iss` key is required and needs to be used to look up the public key to verify the JSON Web Token signature. Inferring the OAuth2 Client ID from the issuer field might be dangerous as we're confusing concepts in a highly sensitive security context.

For client authentication, this appears to be quite straight forward as it states in RFC7521 (the one which RFC 7523 extends)

> - client_id
      OPTIONAL.  The client identifier as described in [Section 2.2](https://datatracker.ietf.org/doc/html/rfc7521#section-2.2) of
      OAuth 2.0 [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)].  The "client_id" is unnecessary for client
      assertion authentication because the client is identified by the
      subject of the assertion.  If present, the value of the
      "client_id" parameter MUST identify the same client as is
      identified by the client assertion.

https://datatracker.ietf.org/doc/html/rfc7521#section-4.2

For authorization grants, there appear to be no such rules. Here, [RFC7521#4.1](https://datatracker.ietf.org/doc/html/rfc7521#section-4.1) states:

> Authentication of the client is optional, as described in
   [Section 3.2.1](https://datatracker.ietf.org/doc/html/rfc7521#section-3.2.1) of OAuth 2.0 [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)], and consequently, the
   "client_id" is only needed when a form of client authentication that
   relies on the parameter is used.

In my interpretation, this means that `client_id` is not needed **if and only if** we need to authenticate / identify the client. It does not state that we can use the `iss` as a fallback to identify the client.

@Xopek you originally wrote this, what's your take on this?

akrause-cel: Taking a step back and looking at hydra we are already using the issuer to identify the trust relationship (https://www.ory.sh/docs/hydra/guides/oauth2-grant-type-jwt-bearer#oauth2-jwt-bearer-grant-type-validation). 

Using the issuer to also identify the client gives users a chance to control where going without auth is appropriate and where not.

However, dropping the requirement for client authentication for the `urn:ietf:params:oauth:grant-type:jwt-bearer` grant entirely is fine for me. I am happy to make the following changes:

1. Authenticate client when credentials are present (Form, basic or any other form supported) ==> No change needed
2. Allow unauthenticated requests for `urn:ietf:params:oauth:grant-type:jwt-bearer` grant.

james-d-elliott: From my reading of the spec, the client_id/client_secret is only excluded from the form for public clients, and potentially the client_id should be in the sub claim:

[Section 3](https://datatracker.ietf.org/doc/html/rfc7523#section-3) (see 2B):

```
        B.  For client authentication, the subject MUST be the
            "client_id" of the OAuth client.
```

I've also done a quick look at what other OP's are doing and many set both `iss` and `sub` to the client ID, and ensure the `aud` includes the token endpoint URL or token endpoint's issuer URL. 

See [Curity docs](https://curity.io/resources/learn/jwt-assertion/#jwt-format):

```
RFC 7521 doesn't dictate the value of the issuer (iss) claim, but it does require one. OpenID Connect does, however. It mandates that the issuer and subject (sub) are the client ID. For this reason, Curity requires both claims to be present with a value of the client ID. Both standards require an audience claims (aud); OpenID Connect says it should be the OAuth server's token endpoint, whereas RFC 7523 allows it to be any identifier that represents the OAuth server.
```

I'll do some more digging as to why Curity claims this and if I can make some sense of it. Just thought I'd share for the sake of potentially helping this along if there is something I missed.

james-d-elliott: This [OpenID Connect Health Relationship Trust spec](https://openid.net/specs/openid-heart-oauth2-2015-12-07.html#RequestsToTokenEndpoint) has this to say:

> Full clients and direct access clients as defined above MUST authenticate to the authorization server's token endpoint using a JWT assertion as defined by the [JWT Profile for OAuth 2.0 Client Authentication and Authorization Grants](https://openid.net/specs/openid-heart-oauth2-2015-12-07.html#RFC7523) [RFC7523] and the private_key_jwt method defined in [OpenID Connect Core](https://openid.net/specs/openid-heart-oauth2-2015-12-07.html#OpenID.Core) [OpenID.Core]. The assertion MUST use the claims as follows:
> 
> iss
>     the client ID of the client creating the token
> sub
>     the client ID of the client creating the token
> aud
>     the URL of the authorization server's token endpoint
> iat
>     the time that the token was created by the client
> exp
>     the expiration time, after which the token MUST be considered invalid
> jti
>     a unique identifier generated by the client for this authentication. This identifier MUST contain at least 128 bits of entropy and MUST NOT be re-used by any subsequent authentication token.

aeneasr: Also please check: https://github.com/ory/fosite/blob/56c818b5544a2d7ce1e411279135ddfb3f01c04b/handler/rfc7523/handler.go#L46-L47

akrause-cel: >  Try creating a new public client without a secret, use the client assertion, and also set client_id in the request body with the ID of the client that is public.

Unfortunately my client (Salesforce.com) does not allow me to set the client_id. Although not nice this behavior seems to be backed by [rfc7523](https://datatracker.ietf.org/doc/html/rfc7523#section-2.1):

> Authentication of the client is optional, as described in
>    [Section 3.2.1](https://datatracker.ietf.org/doc/html/rfc7523#section-3.2.1) of OAuth 2.0 [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)] and consequently, the
>    "client_id" is only needed when a form of client authentication that
>    relies on the parameter is used.

The whole aim of the PR is not to authenticate but to identify.

aeneasr: Ok, I see, that is indeed unfortunate. For salesforce, does it use the client ID for the issuer? We will be phasing out the ability to define custom client IDs with Ory Hydra 2.0, so this might be another blocker.

BTW I pinged @Xopek and he'll take a look tomorrow as well - he was the one who wrote this feature initially

akrause-cel: SFDC is really easy on claim contents. However they do not allow adding claims. Below a screenshot of the config options:

Screenshot 2022-03-22 at 19 11 04

In my solution I use a uuid as client id to be on the safe side.

Xopek: Hi all! Skiping client authentification is only possible with  urn:ietf:params:oauth:grant-type:jwt-bearer grant (as indicated in [rfc7523](https://datatracker.ietf.org/doc/html/rfc7523#section-1) that's why i added this ability, but only for this concrete grant.

About issuer and client id: in [RFC 7521](https://datatracker.ietf.org/doc/html/rfc7521)  we can see this:
`there are two common patterns described below: In the first pattern, depicted in Figure 1, the client obtains an
   assertion from a third-party entity capable of issuing, renewing,
   transforming, and validating security tokens.  Typically, such an
   entity is known as a "security token service" (STS) or just "token
   service"...In the second pattern, depicted in Figure 2, the client creates
   assertions locally`

and this:

`Issuer
      A unique identifier for the entity that issued the assertion.
      Generally, this is the entity that holds the key material used to
      sign or integrity-protect the assertion.  Examples of issuers are
      OAuth clients (when assertions are self-issued) and third-party
      security token services.  If the assertion is self-issued, the
      Issuer value is the client identifier.  If the assertion was
      issued by a security token service (STS), the Issuer should
      identify the STS in a manner recognized by the authorization
      server.`

So we can see that issuer (and iss claim) can be client itself or some security service. If client issued assertion ,then indeed issuer must be equal to client id, but only this case. In other cases client id != issuer.

I think we must be explicit in our intentions. Either we should provide client id somehow explicitly and then we can check it or we must skip client authentication entirely. I don't like the idea to try fetch client id from issuer. Because we don't know if issuer of assertion is client itself or not.  

However we are not restricted to use custom claims in assertion, we can use one to provide client id. Actually one already exists: https://www.rfc-editor.org/rfc/rfc8693.html#name-client_id-client-identifier

aeneasr: Thank you @Xopek , I agree with this statement:

> I think we must be explicit in our intentions. Either we should provide client id somehow explicitly and then we can check it or we must skip client authentication entirely. I don't like the idea to try fetch client id from issuer. Because we don't know if issuer of assertion is client itself or not.

@akrause-cel have you tried setting

https://github.com/ory/fosite/blob/56c818b5544a2d7ce1e411279135ddfb3f01c04b/handler/rfc7523/handler.go#L46-L47

I'm not 100% if it allows skipping client auth completely, but if it does, it might just very well solve your use case!

akrause-cel: I can check that in hydra. However doing so would introduce a regression (since currently hydra has it enabled and does not provide a way to disable it). Is this the intended way forward?

aeneasr: We can for sure make this configurable in Hydra!

aeneasr: Just follow the traces of https://github.com/ory/hydra/blob/e3236dd9f6febbb2c147417f6c21a7285418e055/driver/registry_base.go#L303 :)

You'll also need to add it to the config schema: https://github.com/ory/hydra/blob/master/spec/config.json

akrause-cel: One concern that is that the holder of the private key for the trusted JWT can basically ask for any authorization/scope and will receive a grant. There is no way to control this. Do you agree to merge this into ory?

In my case this is not much of an issue, but I imagine that for others this might turn out to be quite significant. Especially since the config is global.

aeneasr: > One concern that is that the holder of the private key for the trusted JWT can basically ask for any authorization/scope and will receive a grant. There is no way to control this. Do you agree to merge this into ory?

The change we'll introduce is not much different from now. Right now you can:

1. Create a trust relationship (scope, subject, issuer)
2. Use that trust relationship to issue a grant for (scope, subject, issuer) with any authenticated client

With this feature you can 

1. Create a trust relationship (scope, subject, issuer)
2. Use that trust relationship to issue a grant for (scope, subject, issuer) without an authenticated client

apart from that, nothing changes. According to https://datatracker.ietf.org/doc/html/rfc7523#section-1

>    This document defines how a JWT Bearer Token can be used to request
   an access token when a client wishes to utilize an existing trust
   relationship, expressed through the semantics of the JWT, **without a
   direct user-approval step** at the authorization server.  It also
   defines how a JWT can be used as a client authentication mechanism.
   The use of a security token for client authentication is orthogonal
   to and separable from using a security token as an authorization
   grant.  They can be used either in combination or separately.  Client
   authentication using a JWT is nothing more than an alternative way
   for a client to authenticate to the token endpoint and must be used
   in conjunction with some grant type to form a complete and meaningful
   protocol request.  JWT authorization grants may be used **with or
   without client authentication or identification**.  Whether or not
   client authentication is needed in conjunction with a JWT
   authorization grant, as well as the supported types of client
   authentication, are **policy decisions at the discretion of the
   authorization server.**

This section gives us authority to choose whether auth is required or not. The policy decision is up to the policy maker (i.e. sysadmin), thus it appears to be acceptable / intended for such a feature to be configurable. I for sure don't agree that's a good decision, but a decision should be possible regardless.

It's unfortunate that Okta again, as in many other cases, is not enforcing strict security policies but force downstreams to opt for insecure behavior. /tyin-rant

akrause-cel: > The change we'll introduce is not much different from now. Right now you can:
> 
> 1. Create a trust relationship (scope, subject, issuer)
> 2. Use that trust relationship to issue a grant for (scope, subject, issuer) with any authenticated client
> 
> With this feature you can
> 
> 1. Create a trust relationship (scope, subject, issuer)
> 2. Use that trust relationship to issue a grant for (scope, subject, issuer) without an authenticated client

There is one subtle difference with the implementation of this PR. Here the Admin gets to decide which scopes to grant as there is still a client config involved. With the global flag it is only all or nothing. If hydra acts as a provider for many different resources, this might become an issue.

akrause-cel: > But maybe I'm missing something important here :)

Actually no. I was missing it :-)

Will create a PR against hydra as soon as I have time and I find a workaround for https://github.com/ory/hydra/issues/3044

aeneasr: Cool! I'll then close this PR here, I don't think we'll be able to merge it as is given the discussion we had previously. Is that ok?

aeneasr:
Closing as per https://github.com/ory/fosite/pull/653#discussion_r834301412

@akrause-cel if you do not manage to get this working as expected, please feel free to reach out in this thread again :)

akrause-cel:
@aeneasr : looks like it isn't that simple. There is a foreign key constraint in table `hydra_oauth2_access`. Apparently the Insert 
``` 
"INSERT INTO \"hydra_oauth2_access\" (\"active\", \"challenge_id\", \"client_id\", \"form_data\", \"granted_audience\", \"granted_scope\", \"request_id\", \"requested_at\", \"requested_audience\", \"scope\", \"session_data\", \"signature\", \"subject\") VALUES (:active, :challenge_id, :client_id, :form_data, :granted_audience, :granted_scope, :request_id, :requested_at, :requested_audience, :scope, :session_data, :signature, :subject)"
```
References a client id. 

Hence issuing access tokens for unknown clients seems not the way to go.

Please advise.

aeneasr:
Phew, that’s a tough one. Without a client ID most token lifecycle features (revokation, refresh for example) won’t work or will behave unexpectedly. The best way to address those is to write a test suite covering the cases where the Client ID is NULL, after the DB model has adjusted accordingly by removing the NOT NULL constraint

akrause-cel:
That is most likely beyond my time available. Looking that with this PR and a custom build I have a workable solution....

akrause-cel:
This was the intent for making this PR. But I simply do not have the capacity to perform all these changes.

akrause-cel:
Given all the constraints I believe this PR is the best solution

aeneasr:
I understand, we’ll have to park this then and track it as a separate issue. As some of the other commenters have echoed, using the issuer for identification is problematic and can’t thus be merged upstream, sorry :(

PR #655, opened 2022-03-03
fix: empty client secret via basic auth header means "none" authn

enj:
The existing client authentication code treats an empty client_secret
query parameter to be equivalent to "none" authentication instead of
"client_secret_post."

This change updates the basic auth check to be consistent with this.
That is, an empty secret via the basic auth header is considered to
mean "none" instead of "client_secret_basic."

The "golang.org/x/oauth2" library probes for both methods of
authentication, starting with the basic auth header approach first.

As required, both client ID and secret are encoded in one header:

https://github.com/golang/oauth2/blob/ee480838109b20d468babcb00b7027c82f962065/internal/token.go#L174-L176

Since this code path is taken regardless of if the client is public
or confidential, without this change, it results in an error in the
fosite code.  Since the request fails, the library code simply tries
with the other approach, which works because the client secret is
empty.  Needless to say, it is inefficient to force public clients
to make multiple requests.

Signed-off-by: Monis Khan <mok@vmware.com>

## Checklist

      and signed the CLA.
- [ ] I have referenced an issue containing the design document if my change
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/client_authentication.go
+++ b/client_authentication.go
@@ -233,7 +233,7 @@ func (f *Fosite) DefaultClientAuthenticationStrategy(ctx context.Context, r *htt
 		// If this isn't an OpenID Connect client then we actually don't care about any of this, just continue!
 	} else if ok && form.Get("client_id") != "" && form.Get("client_secret") != "" && oidcClient.GetTokenEndpointAuthMethod() != "client_secret_post" {
 		return nil, errorsx.WithStack(ErrInvalidClient.WithHintf("The OAuth 2.0 Client supports client authentication method '%s', but method 'client_secret_post' was requested. You must configure the OAuth 2.0 client's 'token_endpoint_auth_method' value to accept 'client_secret_post'.", oidcClient.GetTokenEndpointAuthMethod()))
-	} else if _, _, basicOk := r.BasicAuth(); basicOk && ok && oidcClient.GetTokenEndpointAuthMethod() != "client_secret_basic" {
+	} else if _, secret, basicOk := r.BasicAuth(); basicOk && ok && secret != "" && oidcClient.GetTokenEndpointAuthMethod() != "client_secret_basic" {
 		return nil, errorsx.WithStack(ErrInvalidClient.WithHintf("The OAuth 2.0 Client supports client authentication method '%s', but method 'client_secret_basic' was requested. You must configure the OAuth 2.0 client's 'token_endpoint_auth_method' value to accept 'client_secret_basic'.", oidcClient.GetTokenEndpointAuthMethod()))
 	} else if ok && oidcClient.GetTokenEndpointAuthMethod() != "none" && client.IsPublic() {
 		return nil, errorsx.WithStack(ErrInvalidClient.WithHintf("The OAuth 2.0 Client supports client authentication method '%s', but method 'none' was requested. You must configure the OAuth 2.0 client's 'token_endpoint_auth_method' value to accept 'none'.", oidcClient.GetTokenEndpointAuthMethod()))
--- a/client_authentication_test.go
+++ b/client_authentication_test.go
@@ -163,6 +163,25 @@ func TestAuthenticateClient(t *testing.T) {
 			r:      new(http.Request),
 		},
 		{
+			d:      "should pass because client is public and client secret is empty in query param",
+			client: &DefaultOpenIDConnectClient{DefaultClient: &DefaultClient{ID: "foo", Public: true}, TokenEndpointAuthMethod: "none"},
+			form:   url.Values{"client_id": []string{"foo"}, "client_secret": []string{""}},
+			r:      new(http.Request),
+		},
+		{
+			d:      "should pass because client is public and client secret is empty in basic auth header",
+			client: &DefaultOpenIDConnectClient{DefaultClient: &DefaultClient{ID: "foo", Public: true}, TokenEndpointAuthMethod: "none"},
+			form:   url.Values{},
+			r:      &http.Request{Header: clientBasicAuthHeader("foo", "")},
+		},
+		{
+			d:         "should fail because client requires basic auth and client secret is empty in basic auth header",
+			client:    &DefaultOpenIDConnectClient{DefaultClient: &DefaultClient{ID: "foo", Public: true}, TokenEndpointAuthMethod: "client_secret_basic"},
+			form:      url.Values{},
+			r:         &http.Request{Header: clientBasicAuthHeader("foo", "")},
+			expectErr: ErrInvalidClient,
+		},
+		{
 			d:      "should pass with client credentials containing special characters",
 			client: &DefaultOpenIDConnectClient{DefaultClient: &DefaultClient{ID: "!foo%20bar", Secret: complexSecret}, TokenEndpointAuthMethod: "client_secret_post"},
 			form:   url.Values{"client_id": []string{"!foo%20bar"}, "client_secret": []string{complexSecretRaw}},


client_authentication_test.go
```diff
@@ -162,6 +162,18 @@ func TestAuthenticateClient(t *testing.T) {
 			form:   url.Values{"client_id": []string{"foo"}},
 			r:      new(http.Request),
 		},
+		{
+			d:      "should pass because client is public and client secret is empty in query param",
+			client: &DefaultOpenIDConnectClient{DefaultClient: &DefaultClient{ID: "foo", Public: true}, TokenEndpointAuthMethod: "none"},
+			form:   url.Values{"client_id": []string{"foo"}, "client_secret": []string{""}},
+			r:      new(http.Request),
+		},
```
enj, line 170: Note that this test already passes on `master` without any changes.

enj:
@aeneasr ping 😄

aeneasr (approved):
This looks good to me from a code perspective! Can you confirm that this is compliant with the OAuth2 and OpenID specs and link the appropriate sections here either explicitly or implicitly stating that this is ok? Thank you! :)

enj:
@aeneasr from https://datatracker.ietf.org/doc/html/rfc6749#section-2.3:

```
2.3.  Client Authentication

   If the client type is confidential, the client and authorization
   server establish a client authentication method suitable for the
   security requirements of the authorization server.  The authorization
   server MAY accept any form of client authentication meeting its
   security requirements.

   Confidential clients are typically issued (or establish) a set of
   client credentials used for authenticating with the authorization
   server (e.g., password, public/private key pair).

   The authorization server MAY establish a client authentication method
   with public clients.  However, the authorization server MUST NOT rely
   on public client authentication for the purpose of identifying the
   client.

   The client MUST NOT use more than one authentication method in each
   request.

2.3.1.  Client Password

   Clients in possession of a client password MAY use the HTTP Basic
   authentication scheme as defined in [RFC2617] to authenticate with
   the authorization server.  The client identifier is encoded using the
   "application/x-www-form-urlencoded" encoding algorithm per
   Appendix B, and the encoded value is used as the username; the client
   password is encoded using the same algorithm and used as the
   password.  The authorization server MUST support the HTTP Basic
   authentication scheme for authenticating clients that were issued a
   client password.

   For example (with extra line breaks for display purposes only):

     Authorization: Basic czZCaGRSa3F0Mzo3RmpmcDBaQnIxS3REUmJuZlZkbUl3

   Alternatively, the authorization server MAY support including the
   client credentials in the request-body using the following
   parameters:

   client_id
         REQUIRED.  The client identifier issued to the client during
         the registration process described by Section 2.2.

   client_secret
         REQUIRED.  The client secret.  The client MAY omit the
         parameter if the client secret is an empty string.




Hardt                        Standards Track                   [Page 16]


RFC 6749                        OAuth 2.0                   October 2012


   Including the client credentials in the request-body using the two
   parameters is NOT RECOMMENDED and SHOULD be limited to clients unable
   to directly utilize the HTTP Basic authentication scheme (or other
   password-based HTTP authentication schemes).  The parameters can only
   be transmitted in the request-body and MUST NOT be included in the
   request URI.

   ...
```

This bit seems to imply that an empty secret is a valid password:

```
   client_secret
         REQUIRED.  The client secret.  The client MAY omit the
         parameter if the client secret is an empty string.
```

To me it sounds like the same would be true when encoding the empty string in the basic auth header?

aeneasr:
Thank you, I'm not quite sure if I share your interpretation here. Sorry for being very exact on this, but I don't want to introduce incorrect things to the library:

>    Clients in possession of a client password MAY use the HTTP Basic
   authentication scheme as defined in [RFC2617] to authenticate with
   the authorization server

To me this sounds that clients without a password do not use this method. See also requirement:

> The authorization server MUST support the HTTP Basic
   authentication scheme for authenticating clients that were issued a
   client password.

The client secret section you mentioned refers to the other method (not HTTP basic auth):

>
>   Alternatively, the authorization server MAY support including the
   client credentials in the **request-body** using the following
   parameters:
>
>   client_id
         REQUIRED.  The client identifier issued to the client during
         the registration process described by Section 2.2.
>
>   client_secret
         REQUIRED.  The client secret.  The client MAY omit the
         parameter if the client secret is an empty string.


This is probably because HTTP Basic Auth does not support "no password":

* https://stackoverflow.com/questions/49580896/http-authentication-without-username

However, looking deeper into this I'm not sure whether this is correct:

* https://datatracker.ietf.org/doc/html/rfc2617#section-2

```
The "basic" authentication scheme is based on the model that the
   client must authenticate itself with a user-ID and a password for
   each realm.  The realm value should be considered an opaque string
   which can only be compared for equality with other realms on that
   server. The server will service the request only if it can validate
   the user-ID and password for the protection space of the Request-URI.
   There are no optional authentication parameters.

   For Basic, the framework above is utilized as follows:

      challenge   = "Basic" realm
      credentials = "Basic" basic-credentials

   Upon receipt of an unauthorized request for a URI within the
   protection space, the origin server MAY respond with a challenge like
   the following:

      WWW-Authenticate: Basic realm="WallyWorld"

   where "WallyWorld" is the string assigned by the server to identify
   the protection space of the Request-URI. A proxy may respond with the
   same challenge using the Proxy-Authenticate header field.

   To receive authorization, the client sends the userid and password,
   separated by a single colon (":") character, within a base64 [[7](https://datatracker.ietf.org/doc/html/rfc2617#ref-7)]
   encoded string in the credentials.

      basic-credentials = base6[4](https://datatracker.ietf.org/doc/html/rfc2617#ref-4)-user-pass
      base64-user-pass  = <base64 [4] encoding of user-pass,
      user-pass   = userid ":" password
      userid      = *<TEXT excluding ":">
      password    = *TEXT

   Userids might be case sensitive.

   If the user agent wishes to send the userid "Aladdin" and password
   "open sesame", it would use the following header field:
```

where

```
      user-pass   = userid ":" password
      userid      = *<TEXT excluding ":">
      password    = *TEXT
```

Means that userid/password have 0...n TEXT elements, so they can be empty (see https://datatracker.ietf.org/doc/html/rfc2616#section-2.1 for explanation).

So yeah, I'm kinda torn between this is OK and this is not OK 😅 

WDYT?

enj:
> Thank you, I'm not quite sure if I share your interpretation here. Sorry for being very exact on this, but I don't want to introduce incorrect things to the library:

No need to be sorry - foundational security code must be handled with care :)

> > Clients in possession of a client password MAY use the HTTP Basic
> > authentication scheme as defined in [RFC2617] to authenticate with
> > the authorization server
> 
> To me this sounds that clients without a password do not use this method. See also requirement:

I generally think of all clients having a password, with public clients having the empty string as their password.  This may not be the correct interpretation though.

> > The authorization server MUST support the HTTP Basic
> > authentication scheme for authenticating clients that were issued a
> > client password.
> 
> ...
>
> Means that userid/password have 0...n TEXT elements, so they can be empty (see https://datatracker.ietf.org/doc/html/rfc2616#section-2.1 for explanation).
> 
> So yeah, I'm kinda torn between this is OK and this is not OK 😅
> 
> WDYT?

I do not know 😂 

There does not seem to be enough specific guidance here.

The existing library code enforces that an empty client secret can only be used with a public client and I think this change is in-line with that.  The alternative approach could be to tighten the existing `form.Get("client_secret") != ""` check to instead check if the `client_secret` param was passed at all.

cfryanr:
I'm always impressed by the attention paid to the specs by the fosite project. In case it helps, I'll share another opinion on the specs.

I agree with your reading of the basic auth spec that it allows for empty usernames and empty passwords to be encoded into a basic auth authorization header. So if that it allowed, then the question becomes how should fosite handle client authentication with an empty password?

The OAuth2 spec says that:

> Confidential clients or other clients issued client credentials MUST authenticate with the authorization server as described in [Section 2.3](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3) when making requests to the token endpoint.

Source: https://datatracker.ietf.org/doc/html/rfc6749#section-3.2.1

That requirement does not say what public clients should do for authentication. However, that same section also says:

> A client MAY use the "client_id" request parameter to identify itself when sending requests to the token endpoint.  In the "authorization_code" "grant_type" request to the token endpoint, an unauthenticated client MUST send its "client_id" to prevent itself from inadvertently accepting a code intended for a client with a different "client_id".

That seems to say that a public client must use the "client_id" request parameter when exchanging an authcode. It doesn't seem to prevent a public client from also performing client authentication if it so chooses, although it seems like it would be considered unnecessary.

The spec also says:

> The authorization server MAY establish a client authentication method with public clients.  However, the authorization server MUST NOT rely on public client authentication for the purpose of identifying the client.

Source: https://datatracker.ietf.org/doc/html/rfc6749#section-2.3

So the authorization server is explicitly allowed (but not required) to perform client authentication for public clients.

The next section of the spec says:

> Clients in possession of a client password MAY use the HTTP Basic authentication scheme as defined in [[RFC2617](https://datatracker.ietf.org/doc/html/rfc2617)] to authenticate with the authorization server.  The client identifier is encoded using the "application/x-www-form-urlencoded" encoding algorithm per [Appendix B](https://datatracker.ietf.org/doc/html/rfc6749#appendix-B), and the encoded value is used as the username; the client password is encoded using the same algorithm and used as the password.  The authorization server MUST support the HTTP Basic authentication scheme for authenticating clients that were issued a client password.

Source: https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1

This says what a client in possession of a password may do. It does not say anything about what a public client without a password may or may not do. One way to interpret it is that public clients may also use basic auth, since the spec has no opinion on what public clients may/should/must do in regard to basic auth.

The above quoted paragraph also says that the server must support basic auth for clients that were issued a client password. But again, it does not say anything about what the server may/should/must do for public clients with no password.

This seems to give fosite leeway to decide what it wants to do when a client with no password performs seemingly unnecessary client authentication.

I think you may be able to argue that `golang.org/x/oauth2` is performing unnecessary client authentication for public clients. Maybe it should have a configuration option to skip client authentication (while still sending the client_id param). It doesn't seem to offer such an option. It may not be the only client-side library which makes this choice either, since the spec seems to have no opinion on public clients performing client authentication (except that it is not required).

One might argue that for fosite to support as many client-side libraries as possible, maybe it should allow this kind of seemingly unnecessary client authentication for pubic clients.

I'm not sure if fosite allows a confidential client to have a password set to be the empty string? If so, that case should be handled with caution. Of course, it should also carefully handle the inverse case where a confidential client does have a non-empty password but the client authentication request includes an empty password (this is clearly a failed authentication).

client_authentication.go
```diff
@@ -233,7 +233,7 @@ func (f *Fosite) DefaultClientAuthenticationStrategy(ctx context.Context, r *htt
 		// If this isn't an OpenID Connect client then we actually don't care about any of this, just continue!
 	} else if ok && form.Get("client_id") != "" && form.Get("client_secret") != "" && oidcClient.GetTokenEndpointAuthMethod() != "client_secret_post" {
 		return nil, errorsx.WithStack(ErrInvalidClient.WithHintf("The OAuth 2.0 Client supports client authentication method '%s', but method 'client_secret_post' was requested. You must configure the OAuth 2.0 client's 'token_endpoint_auth_method' value to accept 'client_secret_post'.", oidcClient.GetTokenEndpointAuthMethod()))
-	} else if _, _, basicOk := r.BasicAuth(); basicOk && ok && oidcClient.GetTokenEndpointAuthMethod() != "client_secret_basic" {
+	} else if id, secret, basicOk := r.BasicAuth(); basicOk && ok && id != "" && secret != "" && oidcClient.GetTokenEndpointAuthMethod() != "client_secret_basic" {
```
aeneasr, line 236: I think we should always assume that `id` is there?

aeneasr (changes requested):
Thank you all for the comments, I agree with your sentiments and I think we have leeway to make Fosite here a bit more flexible! So just a few comments left to address

client_authentication.go
enj: Sure, I just followed the `form.Get("client_id") != ""` logic above.  `r.BasicAuth()` will return `"", "", true` if the basic auth header is set to `:` (i.e. empty username and password).

enj: Fixed.

aeneasr (approved):

enj:
@aeneasr no rush, but any timeline on the next fosite release?  Just wondering when I can close the loop downstream.  Thanks!

Issue #630, opened 2021-10-19
Access token and code hash are not generated with the right key length


vivshankar:
**Describe the bug**

The access token hash should be computed based on the alg header key length and not hard coded to 256. This is from the spec.

at_hash
    OPTIONAL. Access Token hash value. Its value is the base64url encoding of the left-most half of the hash of the octets of the ASCII representation of the access_token value, where the hash algorithm used is the hash algorithm used in the alg Header Parameter of the ID Token's JOSE Header. For instance, if the alg is RS256, hash the access_token value with SHA-256, then take the left-most 128 bits and base64url encode them. The at_hash value is a case sensitive string. 

Code hash is similar.

c_hash
    Code hash value. Its value is the base64url encoding of the left-most half of the hash of the octets of the ASCII representation of the code value, where the hash algorithm used is the hash algorithm used in the alg Header Parameter of the ID Token's JOSE Header. For instance, if the alg is HS512, hash the code value with SHA-512, then take the left-most 256 bits and base64url encode them. The c_hash value is a case sensitive string. 
    If the ID Token is issued from the Authorization Endpoint with a code, which is the case for the response_type values code id_token and code id_token token, this is REQUIRED; otherwise, its inclusion is OPTIONAL. 

**To Reproduce**

1. Configure the ID token alg header to use, say, RS512
2. Run the flow to produce the id_token
3. Look at the c_hash and at_hash

**Expected behavior**

Implemented per specification.

**Environment**

All

**Additional context**

None

vivshankar:
https://openid.net/specs/openid-connect-core-1_0.html#HybridIDToken

I am on the phone but you should see this in the other sections of the specification too. It's part of the core

vivshankar:
@mitar Sorry I guess I am not getting your point. The fact that it is hard coded is the problem. It should not be hard coded. It should be based on the alg in the ID token header.

mitar:
I think this is not a very secure design. It is error prone and hard to debug/follow the logic. Generally for every field there should be only one place where it is being set.

But let's see what @aeneasr thinks about this issue.

vivshankar:
I have offered 3 paths that can be followed:

1. Use the snippet of code I posted a few comments back that computes the hash based on the alg in the session headers.
2. Provide a hash strategy that can be overriden to be symmetric with the JWTStrategy interface.
3. Custom handler that is allowed to compute this hash outside of Fosite core handlers

IMO the first is the cleanest and most correct based on the specification.

[...]

PR #659, opened 2022-03-26
fix: [#630] Use the correct algorithm for at_hash and c_hash

vivshankar:

The defect is described in #630 along with the discussion.

1. `at_hash`: https://openid.net/specs/openid-connect-core-1_0.html#CodeIDToken
2. `c_hash`: https://openid.net/specs/openid-connect-core-1_0.html#HybridIDToken

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

## Further comments

There are two key changes:

1. Standardize computation of the hash across the different flows
2. Introduce `ComputeHash` that extracts the algorithm, if available, from the openid.Session header and computes the hash. It falls back to 256 key length to retain backward compatibility.

### DIFF
--- a/handler/openid/helper.go
+++ b/handler/openid/helper.go
@@ -27,7 +27,6 @@ import (
 	"crypto/sha256"
 	"crypto/sha512"
 	"encoding/base64"
-	"hash"
 	"strconv"
 
 	"github.com/ory/fosite"
@@ -93,21 +92,15 @@ func (i *IDTokenHandleHelper) IssueExplicitIDToken(ctx context.Context, ar fosit
 // ComputeHash computes the hash using the alg defined in the id_token header
 func (i *IDTokenHandleHelper) ComputeHash(ctx context.Context, sess Session, token string) (string, error) {
 	var err error
-	hashSize := 256
+	hash := sha256.New()
 	if alg, ok := sess.IDTokenHeaders().Get("alg").(string); ok {
-		if hashSize, err = strconv.Atoi(alg[2:]); err != nil {
-			hashSize = 256
+		if hashSize, err := strconv.Atoi(alg[2:]); err == nil {
+			if hashSize == 384 {
+				hash = sha512.New384()
+			} else if hashSize == 512 {
+				hash = sha512.New()
+			}
 		}
-	}
-
-	var hash hash.Hash
-	if hashSize == 384 {
-		hash = sha512.New384()
-	} else if hashSize == 512 {
-		hash = sha512.New()
-	} else {
-		// fallback to 256
-		hash = sha256.New()
 	}
 
 	buffer := bytes.NewBufferString(token)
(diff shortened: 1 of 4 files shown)


handler/openid/helper.go
```diff
@@ -76,3 +89,34 @@ func (i *IDTokenHandleHelper) IssueExplicitIDToken(ctx context.Context, ar fosit
...
+	hashSize := 256
+	if alg, ok := sess.IDTokenHeaders().Get("alg").(string); ok {
+		hashSize, err = strconv.Atoi(alg[2:])
+		if err != nil {
+			hashSize = 256
+		}
+	}
+
+	var hash hash.Hash
+	if hashSize == 384 {
```
aeneasr, line 105: I think it would make sense to set hash directly, without the intermediate hashSize variable :)

aeneasr (changes requested):
Awesome, thank you for your contribution! This looks pretty good and I have some ideas how to improve it further :)

handler/openid/helper.go
vivshankar: Hmm, could you elaborate on this please? The `alg` may or may not be present in the header, so the first step here is to compute the hash size.

aeneasr: I guess something along this pseudo-code:

```
hash := sha256.New()

if alg, ok := sess.IDTokenHeaders().Get("alg").(string); ok {
 		hashSize, _ := strconv.Atoi(alg[2:]) // 0 per default
 		switch hashSize {
 			case 384
				hash = sha512.New()
 			// ...
 		}
 	}

```

vivshankar: @aeneasr Made the change.

handler/openid/helper.go
```diff
@@ -76,3 +88,27 @@ func (i *IDTokenHandleHelper) IssueExplicitIDToken(ctx context.Context, ar fosit
 	resp.SetExtra("id_token", token)
 	return nil
 }
+
+// ComputeHash computes the hash using the alg defined in the id_token header
+func (i *IDTokenHandleHelper) ComputeHash(ctx context.Context, sess Session, token string) (string, error) {
+	var err error
+	hash := sha256.New()
+	if alg, ok := sess.IDTokenHeaders().Get("alg").(string); ok {
+		if hashSize, err := strconv.Atoi(alg[2:]); err == nil {
```
aeneasr, line 97: If `alg` is a string with `len() < 2` this will panic - can you please add a check to ensure this doesn't happen and also a test? The test can either be using Go fuzzing ( https://go.dev/doc/fuzz/ ) or be a hardcoded test.

Since this alg parameter can be user sent, it would be problematic that this can panic easily :)

Thanks!

aeneasr (changes requested):

handler/openid/helper.go
vivshankar: The IDTokenHelper is used in the handlers with the express intention of using it with JWTs. The consumer should accordingly use either a valid algorithm (RS256 etc.) or no alg in the ID token headers. I am happy to add the validation but I don't see this being a problem. For example - if someone uses `RSA256`or `R`, that would clearly be a bad implementation and easy for a developer to identify and fix.

vivshankar:
@aeneasr Pushed in the change based on your review comments. Hope this works.

aeneasr (approved):

Issue #628, opened 2021-10-16
Pushed authorization request


vivshankar:
Pushed authorization request is a mechanism to secure the authorization request. The spec is [here](https://www.rfc-editor.org/rfc/rfc9126.html).

It is proposed that this be added to Fosite.

**Additional context**

This has been discussed with @aeneasr over Slack.

PR #660, opened 2022-03-26
feat: [#628] PAR implementation

vivshankar:
This PR implements [RFC9126 - Pushed Authorization Request](https://www.rfc-editor.org/rfc/rfc9126.html).

It introduces the following:

1. New PushedAuthorizeEndpointHandlers
2. Request and response writer handlers
3. New PARStorage interface
4. New functions for PAR lifecycle added to OAuth2Provider interface and corresponding implementation in Fosite
5. New default handler for PAR called `PushedAuthorizeHandler` that complies with the `PushedAuthorizeEndpointHandler` interface

#628 covers the details of the feature

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

## Further comments

N/A


aeneasr:
Just fyi, currently on vacation and not able to review this most likely in the next 2-3 weeks!

handler/par/flow_pushed_authorize.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/ory/fosite"
+	"github.com/ory/fosite/token/hmac"
+	"github.com/ory/x/errorsx"
+)
+
+const (
+	defaultPARKeyLength = 32
+)
+
+var b64 = base64.URLEncoding.WithPadding(base64.NoPadding)
```
aeneasr, line 19: Is this up to us, or is it specified in a spec?

handler/par/flow_pushed_authorize.go
```diff
@@ -0,0 +1,88 @@
...
+	IsRedirectURISecure func(*url.URL) bool
+}
+
+// HandlePushedAuthorizeEndpointRequest handles a pushed authorize endpoint request. To extend the handler's capabilities, the http request
+// is passed along, if further information retrieval is required. If the handler feels that he is not responsible for
+// the pushed authorize request, he must return nil and NOT modify session nor responder neither requester.
+func (c *PushedAuthorizeHandler) HandlePushedAuthorizeEndpointRequest(ctx context.Context, ar fosite.AuthorizeRequester, resp fosite.PushedAuthorizeResponder) error {
+	storage, ok := c.Storage.(fosite.PARStorage)
+	if !ok {
+		return errorsx.WithStack(fosite.ErrServerError.WithHint("Invalid storage type"))
```
aeneasr, line 38: Could we return a better error here - such as `OAuth2 storage provider does not support Pushed Authorization Requests`?

handler/par/flow_pushed_authorize.go
```diff
@@ -0,0 +1,88 @@
...
+	if err = storage.CreatePARSession(ctx, requestURI, ar); err != nil {
+		return errorsx.WithStack(fosite.ErrServerError.WithHint("Unable to store the PAR session").WithWrap(err).WithDebug(err.Error()))
+	}
+
+	resp.SetRequestURI(requestURI)
+	resp.SetExpiresIn(int(expiresIn.Seconds()))
+	return nil
+}
+
+func (c *PushedAuthorizeHandler) secureChecker() func(*url.URL) bool {
```
aeneasr, line 83: I would feel more comfortable with a signature of `secureChecker(*url.URL) bool` to avoid mistakes where the inner function is not properly being invoked because the caller messed up.

authorize_request_handler.go
```diff
@@ -158,6 +158,12 @@ func (f *Fosite) authorizeRequestParametersFromOpenIDConnectRequest(request *Aut
 	}
 
 	claims := token.Claims
+	// Reject the request if the "request_uri" authorization request
+	// parameter is provided.
+	if requestURI, _ := claims["request_uri"].(string); isPARRequest && requestURI != "" {
```
aeneasr, line 163: I believe this can panic if the key does not exist

authorize_request_handler.go
```diff
@@ -158,6 +158,12 @@ func (f *Fosite) authorizeRequestParametersFromOpenIDConnectRequest(request *Aut
...
+		return errorsx.WithStack(ErrInvalidRequestObject.WithHint("The request must not contain 'request_uri'."))
```
aeneasr, line 164: Clarify the error message a bit:

```
Pushed Authorization Requests can not contain the 'request_uri' parameter.
```

authorize_request_handler.go
```diff
@@ -275,7 +281,51 @@ func (f *Fosite) validateResponseMode(r *http.Request, request *AuthorizeRequest
...
+	requestURI := r.Form.Get("request_uri")
+	if requestURI == "" || !strings.HasPrefix(requestURI, f.PushedAuthorizationRequestURIPrefix) {
+		// nothing to do here
+		return false, nil
+	}
+
+	clientID := r.Form.Get("client_id")
+
+	storage, ok := f.Store.(PARStorage)
+	if !ok {
```
aeneasr, line 294: We have a similar error check in the code above. Can we introduce a singular error that we reuse when the storage does not implement PAR?

authorize_request_handler.go
```diff
@@ -275,7 +281,51 @@ func (f *Fosite) validateResponseMode(r *http.Request, request *AuthorizeRequest
...
+	if parRequest, err = storage.GetPARSession(ctx, requestURI); err != nil {
+		return false, errorsx.WithStack(ErrInvalidRequestURI.WithHint("Invalid PAR session").WithWrap(err).WithDebug(err.Error()))
+	}
+
+	// hydrate the request object
+	request.Merge(parRequest)
+	request.RedirectURI = parRequest.GetRedirectURI()
+	request.ResponseTypes = parRequest.GetResponseTypes()
+	request.State = parRequest.GetState()
+	request.ResponseMode = parRequest.GetResponseMode()
```
aeneasr, line 310: Could you point to where these values are also set so that we ensure we did not forget setting some important ones? Could you please also point to the RFC section (and maybe add them as a comment) that state which parameters can be set as part of PAR?

authorize_request_handler.go
```diff
@@ -275,7 +281,51 @@ func (f *Fosite) validateResponseMode(r *http.Request, request *AuthorizeRequest
...
+
+	if err := storage.DeletePARSession(ctx, requestURI); err != nil {
+		return false, errorsx.WithStack(ErrServerError.WithWrap(err).WithDebug(err.Error()))
+	}
+
+	// validate the clients match
+	if clientID != request.GetClient().GetID() {
+		return false, errorsx.WithStack(ErrInvalidRequest.WithHint("The 'client_id' must match the pushed authorization request."))
```
aeneasr, line 318:
```
The 'client_id' must match the one sent in the pushed authorization request.
```

authorize_request_handler.go
```diff
@@ -290,6 +340,18 @@ func (f *Fosite) NewAuthorizeRequest(ctx context.Context, r *http.Request) (Auth
...
 
+	// Check if this is a continuation from a pushed authorization request
+	if !isPARRequest {
+		if isPAR, err := f.authorizeRequestFromPAR(ctx, r, request); err != nil {
+			return request, err
+		} else if isPAR {
+			// No need to continue
+			return request, nil
+		} else if f.EnforcePushedAuthorization {
+			return request, errorsx.WithStack(ErrInvalidRequest.WithHint("Pushed authorization request is enforced."))
```
aeneasr, line 351: Pushed Authorization Requests are enforced but no such request was sent.

fosite.go
```diff
@@ -85,19 +86,34 @@ func (t *RevocationHandlers) Append(h RevocationHandler) {
 	*t = append(*t, h)
 }
 
+// PushedAuthorizeEndpointHandlers is a list of PushedAuthorizeEndpointHandler
+type PushedAuthorizeEndpointHandlers []PushedAuthorizeEndpointHandler
+
+// Append adds an AuthorizeEndpointHandler to this list. Ignores duplicates based on reflect.TypeOf.
+func (a *PushedAuthorizeEndpointHandlers) Append(h PushedAuthorizeEndpointHandler) {
```
aeneasr, line 93: Please add a test for this in fosite_test

handler/par/flow_pushed_authorize.go
```diff
@@ -0,0 +1,88 @@
+package par
```
aeneasr, line 1: Does it make sense to introduce a handler for this? To me, it appears as if there would be only one implementation of PAR? Or are there already multiple ones?

aeneasr (changes requested):
Thank you very much, I have done a preliminary review of the PR. I still need to read the spec very carefully to see what we can and can't set as part of PAR.

What would be immensely helpful is if you could add code comments to the sections that specify things like "this value MUST be set" or similar. That way, we always know what values are set and why according to the spec.

I'm also not 100% sure how far we need to go with validation here? Do we validate e.g. redirect URLs as part of PAR, or do we validate them when we actually execute the PAR request? Or what about response types? I saw that the PAR handler checks e.g. scope and audience, but it's missing (unless I overlooked it) further validation steps.

Finally, it's not quite clear to me how an exemplary HTTP handler would look like - maybe you could add such a handler to ory/fosite-example?

Thank you so much!

handler/par/flow_pushed_authorize.go
james-d-elliott: I took a look at the spec and nothing specifically references the format of the reference value. Here is the specific section that discusses it:

[2.2](https://datatracker.ietf.org/doc/html/rfc9126#section-2.2).  Successful Response:

>    If the verification is successful, the server MUST generate a request
>    URI and provide it in the response with a "201" HTTP status code.
>    The following parameters are included as top-level members in the
>    message body of the HTTP response using the "application/json" media
>    type as defined by [[RFC8259](https://datatracker.ietf.org/doc/html/rfc8259)].
> 
>    request_uri
>       The request URI corresponding to the authorization request posted.
>       This URI is a single-use reference to the respective request data
>       in the subsequent authorization request.  The way the
>       authorization process obtains the authorization request data is at
>       the discretion of the authorization server and is out of scope of
>       this specification.  There is no need to make the authorization
>       request data available to other parties via this URI.
> 
>    The format of the "request_uri" value is at the discretion of the
>    authorization server, but it MUST contain some part generated using a
>    cryptographically strong pseudorandom algorithm such that it is
>    computationally infeasible to predict or guess a valid value (see
>    [Section 10.10 of [RFC6749]](https://datatracker.ietf.org/doc/html/rfc6749#section-10.10) for specifics).  The authorization server
>    MAY construct the "request_uri" value using the form
>    "urn:ietf:params:oauth:request_uri:<reference-value>" with
>    "<reference-value>" as the random part of the URI that references the
>    respective authorization request data.

[10.10](https://datatracker.ietf.org/doc/html/rfc6749#section-10.10).  Credentials-Guessing Attacks:
>    The authorization server MUST prevent attackers from guessing access
>    tokens, authorization codes, refresh tokens, resource owner
>    passwords, and client credentials.
> 
>    The probability of an attacker guessing generated tokens (and other
>    credentials not intended for handling by end-users) MUST be less than
>    or equal to 2^(-128) and SHOULD be less than or equal to 2^(-160).
> 
>    The authorization server MUST utilize other means to protect
>    credentials intended for end-user usage.

vivshankar:
@aeneasr Yes. Just back from travel. I will look at this late next week.

handler/par/flow_pushed_authorize.go
vivshankar: It's just a state type value. I don't really see a need to make this configurable as long as it meets the spec requirements, which it does.

handler/par/flow_pushed_authorize.go
vivshankar: Yes, we can do that. The expectation would be to never hit this state anyway. This would be a major misconfiguration, where the service is writing a HTTP handler calling into the PAR handlers (NewPushAuthorizationRequest etc.) without actually implementing the storage interface. This was really more of a safety net.

handler/par/flow_pushed_authorize.go
vivshankar: Ok let me take a look. I think I copied this from somewhere else in Fosite.

authorize_request_handler.go
vivshankar: This won't panic. The use of the `, _` ensures that. To confirm, I tried this with Go playground.

authorize_request_handler.go
vivshankar: Ok, I can make this change.

authorize_request_handler.go
vivshankar: [Link to RFC section](https://www.rfc-editor.org/rfc/rfc9126.html#name-request)

```
A client sends the parameters that comprise an authorization request directly to the PAR endpoint. A typical parameter set might include: client_id, response_type, redirect_uri, scope, state, code_challenge, and code_challenge_method as shown in the example below. However, the pushed authorization request can be composed of any of the parameters applicable for use at the authorization endpoint, including those defined in [[RFC6749](https://www.rfc-editor.org/rfc/rfc9126.html#RFC6749)] as well as all applicable extensions. The request_uri authorization request parameter is one exception, and it MUST NOT be provided.
```

These are set in the call to NewAuthorizeRequest from within the NewPushedAuthorizationRequest.

handler/par/flow_pushed_authorize.go
vivshankar: The intent is to allow for additional handlers to extend the spec. For example - there can be FAPI-specific constraints that an implementer might want to implement. So while the basic implementation is just one handler, it makes sense for a SDK like Fosite to offer extension points. This is something I appreciate with the implementation for the TokenEndpointHandler and for the AuthorizeEndpointHandler, so I would like to preserve that extensibility.

aeneasr:
Thank you for the update @vivshankar - glad to have you back! I will look at the comments as soon as possible

handler/par/flow_pushed_authorize.go
aeneasr: Ok, that makes sense to me!

authorize_request_handler.go
aeneasr: You're right!

handler/par/flow_pushed_authorize.go
aeneasr: Makes sense! Good error messages can help tremendously when debugging, so even if this a coding error I think it just gives that extra 1% better developer experience :)

handler/par/flow_pushed_authorize.go
aeneasr: I agree, what I wanted to know also was whether the vocabulary of B64 without padding (I think it's `[a-zA-Z0-9+/]+`) matches the one defined in the RFC

vivshankar:
@aeneasr I have addressed the review comments (guess it didn't need to wait till next week :-) ) 

Here is an example of how a HTTP handler could be written. This is similar or same as the authorize HTTP handler func.

```
authorizeRequest, err := oauth2Provider.NewPushedAuthorizeRequest(ctx, r)
if err != nil {
  oauth2Provider.WritePushedAuthorizeError(rw, authorizeRequest, err)
  return
}

// Next we create a response
resp, err := oauth2Provider.NewPushedAuthorizeResponse(ctx, authorizeRequest, session)
if err != nil {
  oauth2Provider.WritePushedAuthorizeError(rw, authorizeRequest, err)
  return
}

oauth2Provider.WritePushedAuthorizeResponse(rw, authorizeRequest, resp)
```

fosite.go
vivshankar: Done

authorize_request_handler.go
vivshankar: Done

authorize_request_handler.go
vivshankar: Done

authorize_request_handler.go
vivshankar: Done

authorize_request_handler.go
vivshankar: Done

handler/par/flow_pushed_authorize.go
vivshankar: Done

handler/par/flow_pushed_authorize.go
vivshankar: Done

authorize_request_handler.go
aeneasr: @vivshankar did you have time to address this comment?

aeneasr (changes requested):
Thank you! Sorry for the long review times :(

Can you please add an integration tests for this feature? Since we're adding a completely new flow and endpoint it is required to ensure the code works as intended.

You can find an example here: https://github.com/ory/fosite/blob/master/integration/authorize_code_grant_test.go

vivshankar:
@aeneasr Not entirely clear what's failing with the linter. I have completed the changes as requested. I will add to fosite-example in a couple of days.

vivshankar:
@aeneasr Will you be able to review the latest changes and we can close this item? I am happy to contribute the new config provider for PAR as well.

authorize_request_handler.go
vivshankar: I have addressed this comment by adding a specific error message for this. Not quite sure what you mean by "similar error check".

vivshankar:
@aeneasr I have now made the changes based on the `Configurator`. Can you please review?

pushed_authorize_request_handler.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/ory/fosite/i18n"
+	"github.com/ory/x/errorsx"
+)
+
+// NewPushedAuthorizeRequest validates the request and produces an AuthorizeRequester object that can be stored
+func (f *Fosite) NewPushedAuthorizeRequest(ctx context.Context, r *http.Request) (AuthorizeRequester, error) {
+	request := NewAuthorizeRequest()
+	request.Request.Lang = i18n.GetLangFromRequest(f.Config.GetMessageCatalog(ctx), r)
+
+	if r.Method != "" && r.Method != "POST" {
```
aeneasr, line 17: When would r.Method be empty? Can we remove this check?

pushed_authorize_request_handler.go
```diff
@@ -0,0 +1,63 @@
...
+		r.Form.Set("client_id", client.GetID())
+	}
+
+	// Validate as if this is a new authorize request
+	fr, err := f.newAuthorizeRequest(ctx, r, true)
+	if err != nil {
+		return fr, err
+	}
+
+	if fr.GetRequestedScopes().Has("openid") && r.Form.Get("redirect_uri") == "" {
```
aeneasr, line 58: Do you have a reference as to why this is required? It looks like this should be left up to the OpenID Connect handler

authorize_request_handler_oidc_request_test.go
```diff
@@ -213,7 +213,7 @@ func TestAuthorizeRequestParametersFromOpenIDConnectRequest(t *testing.T) {
 				},
 			}
 
-			err := f.authorizeRequestParametersFromOpenIDConnectRequest(context.Background(), req)
+			err := f.authorizeRequestParametersFromOpenIDConnectRequest(context.Background(), req, false)
```
aeneasr, line 216: Would we need a test suite which has this set to true?

authorize_request_handler.go
aeneasr: We check on multiple occasions that the config or store implement PAR. We could use a single error `var PARNotImplemented = ...` and reuse that error instead of writing new errors for every check

pushed_authorize_request_handler.go
```diff
@@ -0,0 +1,63 @@
...
+
+	if r.Method != "" && r.Method != "POST" {
+		return request, errorsx.WithStack(ErrInvalidRequest.WithHintf("HTTP method is '%s', expected 'POST'.", r.Method))
+	}
+
+	if err := r.ParseMultipartForm(1 << 20); err != nil && err != http.ErrNotMultipart {
+		return request, errorsx.WithStack(ErrInvalidRequest.WithHint("Unable to parse HTTP body, make sure to send a properly formatted form request body.").WithWrap(err).WithDebug(err.Error()))
+	}
+	request.Form = r.Form
+	request.State = request.Form.Get("state")
```
aeneasr, line 25: The lines from the start of the function until here are more or less a duplicate of newAuthorizeRequest. Do we need to copy all of the code, or can we re-use newAuthorizeRequest instead?

integration/pushed_authorize_code_grant_test.go
```diff
@@ -0,0 +1,202 @@
...
+			resp, err := http.DefaultClient.Do(req)
+
+			require.NoError(t, err)
+
+			body, err := checkStatusAndGetBody(t, resp, c.parStatusCode)
+			require.NoError(t, err, "Unable to get body after PAR. Err=%v", err)
+
+			if resp.StatusCode != http.StatusCreated {
+				return
+			}
```
aeneasr, line 145: Shouldn't this be a failure instead? Right now it will just silently pass the test

integration/pushed_authorize_code_grant_test.go
```diff
@@ -0,0 +1,202 @@
...
+
+			m := map[string]interface{}{}
+			err = json.Unmarshal(body, &m)
+
+			assert.NoError(t, err, "Error occurred when unamrshaling the body: %v", err)
+			assert.Condition(t, func() bool { return m["request_uri"].(string) != "" }, "request_uri is empty")
+			assert.Condition(t, func() bool {
+				return strings.HasPrefix(m["request_uri"].(string), "urn:ietf:params:oauth:request_uri:")
+			}, "PAR Prefix is incorrect: %v", m["request_uri"].(string))
+			assert.EqualValues(t, 300, int(m["expires_in"].(float64)), "Invalid expires_in value=%v", m["expires_in"])
```
aeneasr, line 155: An easier solution to write and read this is to use gjson:

```go
assert.Empty(t, gjson.Get(body, "request_uri").String()), "request_uri is empty"")
```

integration/pushed_authorize_code_grant_test.go
```diff
@@ -0,0 +1,202 @@
...
+			req, err = http.NewRequest("POST", ts.URL+"/auth", strings.NewReader(data.Encode()))
+			require.NoError(t, err)
+
+			req.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+
+			resp, err = http.DefaultClient.Do(req)
+			require.NoError(t, err)
+			require.Equal(t, c.authStatusCode, resp.StatusCode)
+			if resp.StatusCode != http.StatusOK {
+				return
```
aeneasr, line 170: Shouldn't this fail the test or do some other assertion besides the status code check?

aeneasr (changes requested):
Thank you! I have some comments left but overall this looks quite good!

pushed_authorize_request_handler.go
vivshankar: This comes into play if you use one of the test frameworks. I guess I could remove it but I know I added it in my own implementation for some reason, but can't remember at the moment.

pushed_authorize_request_handler.go
vivshankar: This is currently missing from the OpenID Connect handler I believe. In any case, this comes from the [Open ID Connect Core specification](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest), specifically -

```
redirect_uri
REQUIRED. Redirection URI to which the response will be sent. This URI MUST exactly match one of the Redirection URI values for the Client pre-registered at the OpenID Provider, with the matching performed as described in Section 6.2.1 of [[RFC3986]](https://openid.net/specs/openid-connect-core-1_0.html#RFC3986) (Simple String Comparison). When using this flow, the Redirection URI SHOULD use the https scheme; however, it MAY use the http scheme, provided that the Client Type is confidential, as defined in Section 2.1 of OAuth 2.0, and provided the OP allows the use of http Redirection URIs in this case. The Redirection URI MAY use an alternate scheme, such as one that is intended to identify a callback into a native application.
```

I feel it should be checked like all of the other standard checks we perform.

authorize_request_handler_oidc_request_test.go
vivshankar: The `true` case is covered in the PAR request. The final parameter is `isPARRequest`.

pushed_authorize_request_handler.go
vivshankar: We need the `request` to perform other checks before we call into newAuthorizeRequest. The order of validation follows the spec.

integration/pushed_authorize_code_grant_test.go
vivshankar: The failure occurs before this call when we assert the HTTP status code.

integration/pushed_authorize_code_grant_test.go
vivshankar: The failure occurs before this call when we assert the HTTP status code.

vivshankar: It does attempt to exchange the code for the token and that would fail if the code isn't successfully returned, but I see your point. Adding a check to verify the code is non-empty.

mchristophe:
Hello everyone! Thanks for this PR and bringing this topic. This is exactly what I was needing. I am using Ory Hydra (which is great btw, thanks for the hard work!) and I was missing the PAR notion. Hope we will all be able to test it soon on our implementation of Hydra.

mchristophe:
I have another question regarding this implementation.
There are currently 2 topics regarding the parameters sent to the `/authorize` endpoint and some concerns regarding how it works today, especially within OpenBanking world where clients are asked to sign confidential transactions. There are ongoing discussions about how we should rethink the scope notion to be more granular, making them more dynamic.
While waiting for a proper answer from IETF, 2 topics regarding the subject:
1. The OAuth 2.0 Authorization Framework: JWT-Secured Authorization Request (JAR)
RFC 9101](https://datatracker.ietf.org/doc/rfc9101/ ): instead of sending the parameters, goal would be to provide a request object (JWT signed) sent to the AS, or sent as reference (to a place reachable by the AS)
2. [OAuth 2.0 Rich Authorization Requests](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-rar): **DRAFT** new parameter authorization_details that is used to carry fine-grained authorization data in OAuth messages

The latest one would be a perfect fit for the PAR integration, while the 1st one would be a quick solution to manage custom parameters as secured as possible within financial sector.

My question is: could we join forces to proactively work on integrating those features from IETF ? I know the second one is in draft status, but we could already maybe analyse how it would fit into fosite ?

Thanks for your feedback :)

vivshankar:
@mchristophe I can take a crack because we seem to be on the same journey.

- JAR: Fosite already supports signed request objects. So you are covered there. That format is also supported with the PAR implementation, so a client can send the signed request object as part of the PAR payload. I have my own extensions here that enforces the request object to contain all parameters other than client credentials, which is a requirement for FAPI 1.0 Advanced Final (if that is a concern for you).
- Rich authorization request is something I have been paying close attention to and is a major advancement to capture fine-grained consents. In my personal opinion, the `authorization_details` make perfect sense for expressive consent statements, while scopes continue to be good enough for course grained consent statements. It is however still a draft, last I checked, and ory (understandably) waits for it to become a RFC. This is the same reason DPoP hasn't been implemented even though DPoP is the only available solution to enforce sender-constrained access tokens for SPAs.

mchristophe:
@vivshankar : Question to finally clear my mind on that. When something is possible on Fosite (the JAR in this matter), does it mean that it is available on Hydra as well ? I know Hydra is an implementation of Fosite, so I guess the answer is yes. But the question popped up in my head while reading the discussion here.
Regarding your second point, I asked the authors of RAR draft why they were not using scopes instead of types within the json object definition (authorization_details). I have the feeling that doing so would assure backward compatibility w/ old oAuth implementation w/o too much effort. In the meantime, I guess I could already use that kind of new parameter within the request object and do my own check

vivshankar:
@mchristophe I expect so but since I don't work in ory or on Hydra, I will defer to @aeneasr to answer that.

aeneasr:
Thank you @vivshankar - great answer!

> @vivshankar : Question to finally clear my mind on that. When something is possible on Fosite (the JAR in this matter), does it mean that it is available on Hydra as well ? I know Hydra is an implementation of Fosite, so I guess the answer is yes. But the question popped up in my head while reading the discussion here.

Most of the times fixes land automatically in Ory Hydra. Here this is however not the case, as it requires implementation of additional endpoints and business logic. That in turn requires e2e tests, API specs, docs, ...

From maintainer side, this topic is far back in the list of priorities for Ory Hydra to support as we're focusing on v2 and releasing ory hydra to ory cloud

pushed_authorize_request_handler.go
aeneasr: Thanks, i removed it and resolved the failing test cases :)

https://github.com/ory/fosite/pull/660/commits/49729feaddcd3faaf43b1d06b601ac93b47599df

vivshankar: @aeneasr Thanks.

pushed_authorize_request_handler.go
aeneasr: I see, nice find - thank you. I do actually think that we might not have such a check in place yet!

https://github.com/ory/fosite/issues/685

pushed_authorize_request_handler.go
aeneasr: ok, make ssnese - thanks!

integration/pushed_authorize_code_grant_test.go
aeneasr: I see, we're explicitly testing for the response code before - make sense

integration/pushed_authorize_code_grant_test.go
aeneasr: It's still a bit verbose with the typed handling, but in the grand scheme of things it's fine :)

integration/pushed_authorize_code_grant_test.go
aeneasr: This is the last point - I couldn't find whether you made changes to address this comment. Is it still open or resolved?

aeneasr: I see now, you added the check below

aeneasr (approved):
Thank you for your perseverance! This was not an easy change to get merged and with other priorities coming in between I was not able to be as responsive as I wish I could have been. That's the price once you go professional, you don't have as much time any more to be on GitHub :/

A job well done - looking forward to see this in prod!

integration/pushed_authorize_code_grant_test.go
vivshankar: Right after this if check, there's a check on `code` being non-empty. The validation of the code itself would happen when the request is made to the token endpoint.

mchristophe:
Hello @aeneasr , thanks for your job! Indeed a very nice feature nicely handled. I did not understand your answer to my last question: does this mean PAR will be part of the next version of Hydra ? Sorry for asking this again, but I want to clear that out :). Thanks

maricavor:
Hello, we are using Hydra version 2.2.0, but no PAR flow there, which version will support PAR flow?

FMBmyver:
Hi! Great initiative here! is it possible to know if PAR is included in hydra now? :)

PR #662, opened 2022-04-06
feat: Rotation Function to HMAC Strategy

erictg:
The usage of key rotation for the HMAC Strategy was unclear so I added a function to do key rotation.

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/token/hmac/hmacsha.go
+++ b/token/hmac/hmacsha.go
@@ -121,7 +121,7 @@ func (c *HMACStrategy) Validate(token string) (err error) {
 	return err
 }
 
-// Rotate moves the current secret to the list of rotated secrets and sets the
+// Rotate moves the current secret to the list of rotated secrets and sets the newSecret to the GlobalSecret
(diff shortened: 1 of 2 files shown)


aeneasr (approved):
Awesome, thank you!

aeneasr (commented):
Unfortunately, on second thought, I don't think it makes sense to merge it. The code itself is perfect, and the idea solid. But unfortunately we are planning to deprecate all static config options to finally introduce hot reloading into Ory Fosite and Ory Hydra. To do hot reloading, we can no longer configure things as static fields in services or strategies, but will instead rely on function calls.

For that reason, we would probably need to remove this piece of code once that feature lands, so it's better to not introduce it and prevent any breaking changes.

Your contribution is still valuable, it just does not align with the things we have planned for the next couple of weeks and months. Sorry :(

PR #698, opened 2022-08-31
Fix: Use proper scopes in refresh token handler

### LINKED ISSUE #696: Unable to request narrower scopes in OAuth2 Refresh Token Exchange
### Preflight checklist

### Describe the bug

Hello!

I recently discovered that the [Refresh Token Grant Handler](https://github.com/ory/fosite/blob/master/handler/oauth2/flow_refresh.go#L54) does not currently honor the (optional) user requested `scope` parameter to request a subset of the originally granted scopes. 

For reference, here is the [code in question](https://github.com/ory/fosite/blob/master/handler/oauth2/flow_refresh.go#L102-L107):
```
		if !c.Config.GetScopeStrategy(ctx)(request.GetClient().GetScopes(), scope) {
			return errorsx.WithStack(fosite.ErrInvalidScope.WithHintf("The OAuth 2.0 Client is not allowed to request scope '%s'.", scope))
		}
		request.GrantScope(scope)
	}
```

The issues that I see with the current implementation are:
1. The current implementation does not read the `scope` form param.
2. The implementation currently compares the originally granted scopes against the current **client scopes**, but this endpoint should be completely agnostic of the client scopes. I think it should only depend on the originally granted scopes (and optionally the `scope` form param). The result is that the refresh token exchange will fail if the client scopes are themselves narrowed after a refresh token is created. 

This is the implementation that I would propose, given my understanding of https://www.rfc-editor.org/rfc/rfc6749#section-6 
```
	if request.GetRequestForm().Has("scope") {
		request.SetRequestedScopes(strings.Split(request.GetRequestForm().Get("scope"), " "))
	} else {
		request.SetRequestedScopes(originalRequest.GetGrantedScopes())
	}

	for _, scope := range request.GetRequestedScopes() {
		if !c.Config.GetScopeStrategy(ctx)(originalRequest.GetGrantedScopes(), scope) {
			return errorsx.WithStack(fosite.ErrInvalidScope.WithHintf("The OAuth 2.0 Client is not allowed to request scope '%s'.", scope))
		}
		request.GrantScope(scope)
	}
```

Some pertinent notes from my reading of https://www.rfc-editor.org/rfc/rfc6749#section-6:
```
scope
         [Section 3.3](https://www.rfc-editor.org/rfc/rfc6749#section-3.3).  The requested scope MUST NOT include any scope
         not originally granted by the resource owner, and if omitted is
         treated as equal to the scope originally granted by the
         resource owner.
```
* This optional `scope` parameter should allow the client to request a **narrower set** of scopes than were originally granted by the resource owner. 
* The refresh token grant handler should be completely agnostic of the **client scopes**. It should only depend on the originally granted scopes, and optionally the `scope` form param. 
* I would expect this narrower set of scopes to apply only to this specific access token, and not to future refresh token requests. (This is referenced later in the RFC with `"If a new refresh token is issued, the refresh token scope MUST be identical to that of the refresh token included by the client in the request."`)

### Reproducing the bug

The current unit tests reproduce this issue. For example:

```
=== RUN   TestRefreshFlow_HandleTokenEndpointRequest/strategy=hmac/case=should_pass
    flow_refresh_test.go:194: 
        	Error Trace:	flow_refresh_test.go:194
        	            				flow_refresh_test.go:356
        	Error:      	Not equal: 
        	            	expected: fosite.Arguments{"foo", "bar", "offline"}
        	            	actual  : fosite.Arguments{"foo", "offline"}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,4 +1,3 @@
        	            	-(fosite.Arguments) (len=3) {
        	            	+(fosite.Arguments) (len=2) {
        	            	  (string) (len=3) "foo",
        	            	- (string) (len=3) "bar",
        	            	  (string) (len=7) "offline"
        	Test:       	TestRefreshFlow_HandleTokenEndpointRequest/strategy=hmac/case=should_pass
        --- FAIL: TestRefreshFlow_HandleTokenEndpointRequest/strategy=hmac/case=should_pass (0.00s)


Expected :fosite.Arguments{"foo", "bar", "offline"}
Actual   :fosite.Arguments{"foo", "offline"}
```

### Relevant log output

_No response_

### Relevant configuration

_No response_

### Version

master

_No response_

_No response_

### Additional Context

_No response_

silverspace:
Fixes #696 

The token refresh handler currently has two issues related to scopes:
1. It does not read or honor the optional `scope` form parameter
2. It defaults to the current `client` scopes, rather than the originally `granted` scopes. One issue related to this is that if the client scopes are narrowed, then the existing refresh tokens may fail to refresh, even if the user originally consented to those scopes. 

This PR fixes the OAuth2 token refresh handler to:
- Read and use the optional 'scope' form parameter, if present.
- Otherwise default to requesting the originally granted scopes.
    
The token refresh handler endpoint should be completely agnostic of:
- The originally **requested** scopes
- The **client scopes** (both current and past client scopes)

#696

## Checklist

      I confirm that I got green light (please contact [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push the changes.

### DIFF
--- a/handler/oauth2/flow_refresh.go
+++ b/handler/oauth2/flow_refresh.go
@@ -96,11 +96,20 @@ func (c *RefreshTokenGrantHandler) HandleTokenEndpointRequest(ctx context.Contex
 	}
 
 	request.SetSession(originalRequest.GetSession().Clone())
-	request.SetRequestedScopes(originalRequest.GetRequestedScopes())
 	request.SetRequestedAudience(originalRequest.GetRequestedAudience())
 
-	for _, scope := range originalRequest.GetGrantedScopes() {
-		if !c.Config.GetScopeStrategy(ctx)(request.GetClient().GetScopes(), scope) {
+	if _, ok := request.GetRequestForm()["scope"]; ok {
+		requestedScopes := fosite.RemoveEmpty(strings.Split(request.GetRequestForm().Get("scope"), " "))
+		if len(requestedScopes) == 0 {
+			return errorsx.WithStack(fosite.ErrInvalidScope.WithHintf("The requested scope parameter is empty"))
+		}
+		request.SetRequestedScopes(requestedScopes)
+	} else {
+		request.SetRequestedScopes(originalRequest.GetGrantedScopes())
+	}
+
+	for _, scope := range request.GetRequestedScopes() {
+		if !c.Config.GetScopeStrategy(ctx)(originalRequest.GetGrantedScopes(), scope) {
 			return errorsx.WithStack(fosite.ErrInvalidScope.WithHintf("The OAuth 2.0 Client is not allowed to request scope '%s'.", scope))
 		}
 		request.GrantScope(scope)
--- a/handler/oauth2/flow_refresh_test.go
+++ b/handler/oauth2/flow_refresh_test.go
@@ -139,7 +139,36 @@ func TestRefreshFlow_HandleTokenEndpointRequest(t *testing.T) {
 					expectErr: fosite.ErrInvalidGrant,
 				},
 				{
-					description: "should fail because offline scope has been granted but client no longer allowed to request it",
+					description: "should fail because requested hierarchic scope was not granted in original request",
+					setup: func(config *fosite.Config) {
+						areq.GrantTypes = fosite.Arguments{"refresh_token"}
+						areq.Client = &fosite.DefaultClient{
+							ID:         "foo",
+							GrantTypes: fosite.Arguments{"refresh_token"},
+							// This can be anything, the current client scopes should be completely ignored.
+							Scopes: []string{"foo.read", "foo.write", "offline"},
+						}
+
+						token, sig, err := strategy.GenerateRefreshToken(nil, nil)
+						require.NoError(t, err)
+
+						areq.Form.Add("refresh_token", token)
+						// Request bar.read, which was not originally granted.
+						areq.Form.Add("scope", "foo.read foo.create offline")
+						err = store.CreateRefreshTokenSession(nil, sig, &fosite.Request{
+							Client:         areq.Client,
+							GrantedScope:   fosite.Arguments{"foo.read", "foo.write", "offline"},
+							RequestedScope: fosite.Arguments{"foo.read", "foo.write", "offline"},
+							Session:        sess,
+							Form:           url.Values{"scope": []string{"foo.read foo.write offline"}},
+							RequestedAt:    time.Now().UTC().Add(-time.Hour).Round(time.Hour),
+						})
+						require.NoError(t, err)
+					},
+					expectErr: fosite.ErrInvalidScope,
+				},
+				{
+					description: "should fail because scope parameters are comma separated instead of space separated",
 					setup: func(config *fosite.Config) {
 						areq.GrantTypes = fosite.Arguments{"refresh_token"}
 						areq.Client = &fosite.DefaultClient{
@@ -151,10 +180,64 @@ func TestRefreshFlow_HandleTokenEndpointRequest(t *testing.T) {
 						require.NoError(t, err)
 
 						areq.Form.Add("refresh_token", token)
+						areq.Form.Add("scope", "foo.read,foo.write,offline")
 						err = store.CreateRefreshTokenSession(nil, sig, &fosite.Request{
 							Client:         areq.Client,
-							GrantedScope:   fosite.Arguments{"foo", "offline"},
-							RequestedScope: fosite.Arguments{"foo", "offline"},
+							GrantedScope:   fosite.Arguments{"foo.read", "foo.write", "offline"},
+							RequestedScope: fosite.Arguments{"foo.read", "foo.write", "offline"},
+							Session:        sess,
+							Form:           url.Values{"scope": []string{"foo.read foo.write offline"}},
+							RequestedAt:    time.Now().UTC().Add(-time.Hour).Round(time.Hour),
+						})
+						require.NoError(t, err)
+					},
+					expectErr: fosite.ErrInvalidScope,
+				},
+				{
+					description: "should fail because scope parameter is defined but empty",
+					setup: func(config *fosite.Config) {
+						areq.GrantTypes = fosite.Arguments{"refresh_token"}
+						areq.Client = &fosite.DefaultClient{
+							ID:         "foo",
+							GrantTypes: fosite.Arguments{"refresh_token"},
+						}
+
+						token, sig, err := strategy.GenerateRefreshToken(nil, nil)
+						require.NoError(t, err)
+
+						areq.Form.Add("refresh_token", token)
+						areq.Form.Add("scope", "    ")
+						err = store.CreateRefreshTokenSession(nil, sig, &fosite.Request{
+							Client:         areq.Client,
+							GrantedScope:   fosite.Arguments{"foo.read", "offline"},
+							RequestedScope: fosite.Arguments{"foo.read", "offline"},
+							Session:        sess,
+							Form:           url.Values{"scope": []string{"foo.read offline"}},
+							RequestedAt:    time.Now().UTC().Add(-time.Hour).Round(time.Hour),
+						})
+						require.NoError(t, err)
+					},
+					expectErr: fosite.ErrInvalidScope,
+				},
+				{
+					description: "should fail with hierarchic scope requested that was not originally granted",
+					setup: func(config *fosite.Config) {
+						areq.GrantTypes = fosite.Arguments{"refresh_token"}
+						areq.Client = &fosite.DefaultClient{
+							ID:         "foo",
+							GrantTypes: fosite.Arguments{"refresh_token"},
+							Scopes:     []string{"<UNUSED>"},
+						}
+
+						token, sig, err := strategy.GenerateRefreshToken(nil, nil)
+						require.NoError(t, err)
+
+						areq.Form.Add("refresh_token", token)
+						areq.Form.Add("scope", "foo offline")
+						err = store.CreateRefreshTokenSession(nil, sig, &fosite.Request{
+							Client:         areq.Client,
+							GrantedScope:   fosite.Arguments{"foo.read", "foo.write", "offline"},
+							RequestedScope: fosite.Arguments{"foo.read", "foo.write", "offline"},
 							Session:        sess,
 							Form:           url.Values{"foo": []string{"bar"}},
 							RequestedAt:    time.Now().UTC().Add(-time.Hour).Round(time.Hour),
@@ -170,7 +253,7 @@ func TestRefreshFlow_HandleTokenEndpointRequest(t *testing.T) {
 						areq.Client = &fosite.DefaultClient{
 							ID:         "foo",
 							GrantTypes: fosite.Arguments{"refresh_token"},
-							Scopes:     []string{"foo", "bar", "offline"},
+							Scopes:     []string{"<UNUSED>"},
 						}
 
 						token, sig, err := strategy.GenerateRefreshToken(nil, nil)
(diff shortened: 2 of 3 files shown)


handler/oauth2/flow_refresh.go
```diff
@@ -96,11 +96,20 @@ func (c *RefreshTokenGrantHandler) HandleTokenEndpointRequest(ctx context.Contex
 	}
 
 	request.SetSession(originalRequest.GetSession().Clone())
-	request.SetRequestedScopes(originalRequest.GetRequestedScopes())
 	request.SetRequestedAudience(originalRequest.GetRequestedAudience())
 
-	for _, scope := range originalRequest.GetGrantedScopes() {
-		if !c.Config.GetScopeStrategy(ctx)(request.GetClient().GetScopes(), scope) {
```
aeneasr, line 103: For context: This design decision was originally made to cover the use case where a client could gain access to scopes that are not in its allowed list. This can happen when a client is updated. I'd say this behavior is correct. The client should not receive scopes which are not in its allowed list.

aeneasr (changes requested):
Thank you very much for the detailed issue and PR! I think this is working though as intended for a reason, please check my comment. Thanks!

handler/oauth2/flow_refresh.go
james-d-elliott: This check was merely moved here from what I can tell: https://github.com/ory/fosite/pull/698/files#diff-e733adcae5e1fe83d4192fc112318cfd1d77b5ad76b6d71d383ae705ea29e3f2R111-R112

From looking purely at the code change what happens is if the request to the token endpoint is a refresh  and the form contains a `scopes` key it uses that as the source of the *intended* scopes, if it doesn't it uses the original request as a source of the intended scopes, then performs the original check to ensure the client is allowed to issue those scopes.

I can see another issue with this PR though. I'll make a review.

handler/oauth2/flow_refresh.go
```diff
@@ -96,11 +96,20 @@ func (c *RefreshTokenGrantHandler) HandleTokenEndpointRequest(ctx context.Contex
...
+	if _, ok := request.GetRequestForm()["scope"]; ok {
+		requestedScopes := fosite.RemoveEmpty(strings.Split(request.GetRequestForm().Get("scope"), " "))
```
james-d-elliott, line 102: We need to check this against originalRequest.GetRequestedScopes() to ensure none of the scopes in the form `scope` value were not originally granted:

```
         The scope of the access request as described by Section 3.3.  
         The requested scope MUST NOT include any scope
         not originally granted by the resource owner, and if omitted is
         treated as equal to the scope originally granted by the
         resource owner.
```

james-d-elliott (changes requested):

handler/oauth2/flow_refresh.go
james-d-elliott: It is strange that the conformity test failed though...

handler/oauth2/flow_refresh.go
silverspace: I don't think the originally **requested** scopes should be used anywhere in the refresh token handler. Note that your RFC quote above only mentions the originally **granted** scopes, not the originally **requested** scopes. 

I've ensured that all scopes in the form `scope` were originally granted by comparing them against the originally **granted** scopes list:
```
	for _, scope := range request.GetRequestedScopes() {
		if !c.Config.GetScopeStrategy(ctx)(originalRequest.GetGrantedScopes(), scope) {
```

handler/oauth2/flow_refresh.go
silverspace: > For context: This design decision was originally made to cover the use case where a client could gain access to scopes that are not in its allowed list. This can happen when a client is updated. I'd say this behavior is correct. The client should not receive scopes which are not in its allowed list.

Do you mind elaborating on this use case, so I can tell whether my implementation solves for that attack vector? Note that my implementation of the refresh token handler completely ignores the client scopes, so those can be updated or  removed and won't change the refresh token handler response. 

My read of the RFC is that increasing (or decreasing) client scopes should only apply to **net new** consents, and should not affect any previous consented scopes.

handler/oauth2/flow_refresh.go
james-d-elliott: Oh I misread the change, the other review is more pertinent my bad.

handler/oauth2/flow_refresh.go
james-d-elliott: I see the issue, we should probably still check the intended scopes against the client to ensure it absolutely can issue them (even though it theoretically did before).

silverspace: I think this would be a bug against the RFC spec, for the following scenario: 
- Client is initially published with required scopes `foo.read foo.write`. 
- User consents to the client app, and creates a refresh token with scope `foo.read foo.write`. 
- Client decides that they don't actually need a bunch of **write** permissions, so the client app is updated to just be `foo.read`. 
- When the existing refresh token is exchanged for an access token **without the `scope` form parameter explicitly requesting `foo.read` only**, the token refresh fails. This failure happens because the requested scope is `foo.read foo.write`, but the client scopes are now `foo.read` only. 

Please see the bolded text below from [RFC6749](https://www.rfc-editor.org/rfc/rfc6749#section-6):

> The authorization server MAY issue a new refresh token, in which case
>    the client MUST discard the old refresh token and replace it with the
>    new refresh token.  The authorization server MAY revoke the old
>    refresh token after issuing a new refresh token to the client.  **If a
>    new refresh token is issued, the refresh token scope MUST be
>    identical to that of the refresh token included by the client in the
>    request.**

Here is my interpretation of the RFC:

1. **Client scopes**
  a. Mutable, may be changed at any time to broaden or narrow the requested scopes **for new tokens**.
  b. Only applies to **net new** refresh tokens. Existing refresh tokens are unmodified by and completely agnostic to client scope changes.

2. **Refresh token scopes**
  a. Immutable. Once a refresh token is granted, its set of scopes is locked in stone. 
  b. The immutable nature of refresh tokens is why the RFC above mentions `If a new refresh token is issued, the refresh token scope MUST be identical to that of the refresh token included by the client in the request.`

3. **Access token scopes**
  a. Immutable, but may be different from the refresh token scopes if the `scope` form parameter is used.
  b. If the client wants to narrow the set of scopes **for a single access token**, they must use the `scope` form parameter.
  c. It is impossible to increase scopes beyond the refresh token scopes that the resource owner granted.

aeneasr: > MUST be identical to that of the refresh token included by the client in the request.

They mean that the "scope" parameter in the refresh token request must be identical or the request should fail. This does not refer to the original issued refresh token. You can ask for a scope in the refresh request: https://www.rfc-editor.org/rfc/rfc6749#section-6

In my view, the current implementation in fosite is following this guidance, and the suggested code changes would regress on the security implications.

Regarding immutability of the scope parameter, I disagree. The scope can change because you can remove scopes from your application. It would be irresponsible to keep highly privileged scopes around because they were granted a year ago, but were flagged during (e.g.) a security review and removed.

james-d-elliott: Ah right, because this section of code is related to refresh tokens, not the access tokens. I completely missed that.

silverspace: I appreciate you keeping an open mind on our disagreement with regard to the RFC interpretation. I will strive to do the same. Please see my inline comments:

> > MUST be identical to that of the refresh token included by the client in the request.
> 
> They mean that the "scope" parameter _in the refresh token request_ must be identical or the request should fail. 

I'm afraid I must respectfully disagree. If this was the RFC author's intent, then they would not say _"**IF** a new refresh token is issued..."_. Their wording also does not mention the form parameter; they simply state that the *new* refresh token scope should be identical to the scope of the refresh token in the request. 

I'd also like to point out one area where we agree 100%: I am in complete agreement that the request should fail if an unauthorized scope is requested in the form parameter. I **DO** agree that the returned access token must have a scope exactly identical to the requested `scope` form parameter, or the request should fail.  

> This does not refer to the original issued refresh token. You can ask for a scope in the refresh request: https://www.rfc-editor.org/rfc/rfc6749#section-6

If you look closer at the wording of the RFC, they are not talking about changing the scope of the **refresh token**. They are talking about returning a more narrowly scoped **access token**: 
```
OPTIONAL.  The scope of the *access request* as described by
         [Section 3.3](https://www.rfc-editor.org/rfc/rfc6749#section-3.3).
```
Note that they're referring to the scope of the "**access request**", and even the _title_ of **Section 3.3** is [Access Token Scope](https://www.rfc-editor.org/rfc/rfc6749#section-3.3).

> 
> In my view, the current implementation in fosite is following this guidance, and the suggested code changes would regress on the security implications.
> 
> Regarding immutability of the scope parameter, I disagree. The scope can change because you can remove scopes from your application. It would be irresponsible to keep highly privileged scopes around because they were granted a year ago, but were flagged during (e.g.) a security review and removed.

I would argue that this is precisely why the RFC authors added the optional `scope` form parameter -- to handle the case where a Client wishes to narrow the scope of their access tokens. 

Maybe we can look at this using a different lens: suppose that the Fosite implementation chose NOT to return a new refresh token in the response, and instead used a long lived immutable refresh token. In this hypothetical example, the RFC makes no mention of updating the refresh token scopes; it only talks about narrowing the scope of the returned **access token**. Can we at least agree that in this hypothetical implementation where we did not rotate refresh tokens, the refresh token scope would be immutable?

silverspace:
Perhaps we can put aside the "client scope" discussion for a moment, and simply agree that the Fosite library DOES have a bug whereby the `scope` form parameter is completely ignored. The refresh token handler does not read the `scope` form parameter in the current implementation, so it is currently impossible to request a more narrowly scoped access token.

handler/oauth2/flow_refresh.go
james-d-elliott: I think maybe it'd be more efficient to focus on one of these areas and resolve it either via discussion or by a commit. I'm  inclined to lean towards the adjustment of the access token scopes during a refresh, as this seems rather clear.

The other is somewhat ambiguous in my reading, specifically because it expressly mentions that the refresh token is *bound* to the client. Thus theoretically if the client has been changed then the new properties of that client; including its existence, type, secret, and permitted scopes; should take effect even though it's not expressly mentioned. 

In addition the spec explicitly states that the OP MAY issue a new refresh token, and that if it does it explicitly states that it MUST have the same scopes. It is therefore reasonable that if the client permitted scopes has been reduced that the OP will refuse to issue a new refresh token. As to how this occurs is the only thing we have to reasonably figure out. Does it error? Or does it issue a new access token WITHOUT a refresh token?

james-d-elliott:
> Perhaps we can put aside the "client scope" discussion for a moment, and simply agree that the Fosite library DOES have a bug whereby the `scope` form parameter is completely ignored. The refresh token handler does not read the `scope` form parameter in the current implementation, so it is currently impossible to request a more narrowly scoped access token.

Yeah I agree with focusing on the one aspect, it's more efficient. Especially with focusing on something I think is less ambiguous is a good idea.

silverspace:
@aeneasr please let me know if you have time to discuss https://github.com/ory/fosite/pull/698#discussion_r961835824

handler/oauth2/flow_refresh.go
aeneasr: Thank you for the detailed write down, in summary, I draw the following conclusion:

1. Authorization Grant is done with scope `a, b, c` - client has allowed scopes `a, b, c`
2. Refresh grant is executed without `?scope` parameter, thus scope `a, b, c` is granted
3. Refresh grant is executed again with `?scope=a,b`, thus scope `a,b` is granted
4. Client is updated to only allow scope `a, b`
5. Refresh grant is executed without `?scope` parameter, thus scope `a, b` is granted
6. Refresh grant is executed with `?scope=a,b,c` parameter, refresh fails
6. Refresh grant is executed with `?scope=a,b` parameter, refresh passes

If the current behavior does not match this pattern, we should fix it. If it does match it, then it is correct.

silverspace: Hi @aeneasr , the above does **NOT** follow [RFC6749](https://www.rfc-editor.org/rfc/rfc6749#section-6). This is precisely the bug that I am encountering in Fosite, and the reason I filed this PR to fix it. The implementation according to the RFC should be:

---
1. Authorization Grant is done with scope a, b, c - client is granted scopes a, b, c
2. Refresh grant is executed without ?scope parameter, thus scope a, b, c is granted
3. Refresh grant is executed again with ?scope=a,b, and scope a,b is granted
4. Client is updated to only allow scope a, b 
5. Refresh grant is executed without ?scope parameter, scope a, b`, c` is granted
6. Refresh grant is executed with ?scope=a,b,c parameter, refresh ~~fails~~ `succeeds`
7. Refresh grant is executed with ?scope=a,b parameter, and scope a,b is granted
---

Refreshing an access token **should be completely agnostic to the currently configured client scopes**. If the RFC wanted us to always use the latest client scopes, then there would be no point in having scopes associated with the refresh token.

silverspace: The other issue that this PR attempts to resolve is that steps (3) and (7) are broken in Fosite, since Fosite does not read the `?scope=a,b` parameter.

aeneasr: I respectfully disagree. https://github.com/ory/fosite/pull/698#discussion_r966244394 is not what I trust the correct sequence to be, I think it reduces security, and it's also not how I interpret the RFC. What could work to convince me otherwise is to get the conformance tests (which test OIDC conformity ...) to pass since they are failing on this PR (potentially due to the changes made), and to get someone from the OAuth2 IETF authors or maintainers to agree with your interpretation.

silverspace: Hi @aeneasr I can understand your reservations, and desire to get sign-off from the OAuth2 IETF authors. How about we focus on the more clear-cut issue of Fosite not honoring the `?scope=a,b` parameter for now? That would at least unblock my particular use case.

aeneasr: Yes, Fosite should definitely respect that parameter! :)

aeneasr:
Given the discussion and reasoning above I will close this PR. Any ideas for future work outlined in this PR we‘ll happily accept though :)

PR #718, opened 2022-11-02
fix(oauth2): requested scope ignored in refresh flow

### LINKED ISSUE #696: Unable to request narrower scopes in OAuth2 Refresh Token Exchange

james-d-elliott:
This addresses an issue where the clients requested scope is ignored during the refresh flow preventing a refresh token from granting an access token with a more narrow scope. It is critical to note that it is completely ignored, and previous behaviour only issued the same scope as originally granted. An access request which requests a broader scope is equally ignored and it has been thoroughly tested to ensure this cannot occur in HEAD. See https://www.rfc-editor.org/rfc/rfc6749#section-6 for more information.

Fixes #696 

## Checklist


handler/oauth2/flow_refresh.go
```diff
@@ -96,13 +95,39 @@ func (c *RefreshTokenGrantHandler) HandleTokenEndpointRequest(ctx context.Contex
...
+	default:
+		request.SetRequestedScopes(fosite.RemoveEmpty(strings.Split(scope, " ")))
+	}
+
 	request.SetRequestedAudience(originalRequest.GetRequestedAudience())
 
-	for _, scope := range originalRequest.GetGrantedScopes() {
+	for _, scope := range request.GetRequestedScopes() {
+		// Addresses point 2 of the text in RFC6749 Section 6.
+		if !originalRequest.GetGrantedScopes().Has(scope) {
```
james-d-elliott, line 123: This needs to be carefully looked at. We may need to check this against the ScopeStrategy instead but I'm not 100% sure so really need a careful evaluation of this particular point.

handler/oauth2/flow_refresh.go
```diff
@@ -96,13 +95,39 @@ func (c *RefreshTokenGrantHandler) HandleTokenEndpointRequest(ctx context.Contex
...
+		// Addresses point 1 of the text in RFC6749 Section 6.
+		request.SetRequestedScopes(originalRequest.GetGrantedScopes())
+	default:
+		request.SetRequestedScopes(fosite.RemoveEmpty(strings.Split(scope, " ")))
+	}
+
 	request.SetRequestedAudience(originalRequest.GetRequestedAudience())
 
-	for _, scope := range originalRequest.GetGrantedScopes() {
+	for _, scope := range request.GetRequestedScopes() {
```
james-d-elliott, line 121: This change here is what caused the change in the seemingly unrelated tests. This needs to be carefully checked as well.

james-d-elliott (commented):
I've done a self review to highlight sections that I am unsure on and to give some context to the changes.

handler/oauth2/flow_refresh.go
```diff
@@ -96,13 +95,39 @@ func (c *RefreshTokenGrantHandler) HandleTokenEndpointRequest(ctx context.Contex
...
+		   			the scope originally granted by the resource owner.
+
+			See https://www.rfc-editor.org/rfc/rfc6749#section-6
+	*/
+	switch scope := request.GetRequestForm().Get("scope"); scope {
+	case "":
+		// Addresses point 1 of the text in RFC6749 Section 6.
+		request.SetRequestedScopes(originalRequest.GetGrantedScopes())
+	default:
+		request.SetRequestedScopes(fosite.RemoveEmpty(strings.Split(scope, " ")))
```
aeneasr, line 116: Would this also work if the scope query parameter is `?scope=foo+bar`?

aeneasr (commented):
This looks very good! Just one minor question

handler/oauth2/flow_refresh.go
james-d-elliott: Let me double check. I believe I copied this specific line from elsewhere.

james-d-elliott: This is the source which should theoretically mean the issue exists here too: https://github.com/ory/fosite/blob/ef3e950835e8dd0ab3405f36745b962ede11e3a8/access_request_handler.go#L80

It appears the go http package parses this correctly as far as I can see locally, Both url.ParseQuery  and url.ParseRequestURI seem to treat the `+` as a space, similar to `%20`.

james-d-elliott:
Regaring the following sequence;

1. Request token with scope a,b,c
2. Do refresh token with scope a,b
3. Do refresh token with scope a,b,c

I can confirm with my high level testing that it fails at step 3.

james-d-elliott:
So looking here: https://github.com/ory/fosite/blob/cda9c24ce1a981f60d6c2e99b7fceec1654c807d/handler/oauth2/flow_refresh.go#L64

The original request we're looking up is the RT Session. Thus I think IF we want to preserve the original scope we just have to ensure the new RT session has the same scope as the original, and the AT session has the narrowed scope. We just have to decide if that's desirable. The issue is I'm not entirely sure how we'd accomplish that, looking at the code it looks like the RT session and AT sessions share their scope.

aeneasr:
> The original request we're looking up is the RT Session. Thus I think IF we want to preserve the original scope we just have to ensure the new RT session has the same scope as the original (granted), and the AT session has the narrowed scope. 

To me that sounds very, very reasonable!

james-d-elliott:
@aeneasr I'm fairly sure of the most effective way to achieve this, however you probably know this code base better than anyone so maybe I've missed something or you have an alternative way that is preferable. 

This is the critical path in making this functional I believe:

https://github.com/ory/fosite/blob/cda9c24ce1a981f60d6c2e99b7fceec1654c807d/handler/oauth2/flow_refresh.go#L190-L199

We need to ensure the `storeReq` for the AT contains the narrowed scope and that the `storeReq` for the RT contains the original scope I believe.

I think the best way to achieve this is to add an interface which extends `AccessRequester` with the following signature:

```go
type RefreshTokenAccessRequester interface {
	// GetRefreshTokenRequestedScopes returns the request's scopes specifically for the refresh token.
	GetRefreshTokenRequestedScopes() (scopes Arguments)
	// SetRefreshTokenRequestedScopes sets the request's scopes specifically for the refresh token.
	SetRefreshTokenRequestedScopes(scopes Arguments)
	// GetRefreshTokenGrantedScopes returns all granted scopes specifically for the refresh token.
	GetRefreshTokenGrantedScopes() (grantedScopes Arguments)
	// SetRefreshTokenRequestedScopes sets the request's scopes specifically for the refresh token.
	SetRefreshTokenRequestedScopes(scopes Arguments)
}
```

The key thing being that if that interface is not implemented that people would not have this functionality.

aeneasr:
I think that could work! We probably need a few more tests to ensure that this works without getting confused in the process in longer refresh token chains :)

access_request.go
```diff
@@ -7,6 +7,9 @@ type AccessRequest struct {
 	GrantTypes       Arguments `json:"grantTypes" gorethink:"grantTypes"`
 	HandledGrantType Arguments `json:"handledGrantType" gorethink:"handledGrantType"`
 
+	RefreshTokenRequestedScope Arguments
```
aeneasr, line 10: This needs `json` tags IMO

handler/oauth2/flow_refresh.go
```diff
@@ -151,8 +185,24 @@ func (c *RefreshTokenGrantHandler) PopulateTokenEndpointResponse(ctx context.Con
...
 	}
 
-	if err = c.TokenRevocationStorage.CreateRefreshTokenSession(ctx, refreshSignature, storeReq); err != nil {
-		return err
+	if rtRequest, ok := requester.(fosite.RefreshTokenAccessRequester); ok {
+		r := requester.Sanitize([]string{}).(*fosite.Request)
+		r.SetID(ts.GetID())
+
+		rtStoreReq := &fosite.AccessRequest{
+			Request: *r,
```
aeneasr, line 193: Isn't this missing a couple of fields like the handled request type?

```
	GrantTypes       Arguments `json:"grantTypes" gorethink:"grantTypes"`
	HandledGrantType Arguments `json:"handledGrantType" gorethink:"handledGrantType"`

	RefreshTokenRequestedScope Arguments
	RefreshTokenGrantedScope   Arguments
```

access_request.go
```diff
@@ -7,6 +7,9 @@ type AccessRequest struct {
...
+	RefreshTokenGrantedScope   Arguments
```
aeneasr, line 11: I assume these two new fields would need to be stored in the database?

aeneasr (commented):
Awesome, getting much closer! :)

access_request.go
james-d-elliott: They shouldn't specifically need to be stored I don't think. The only situation where they're essential is the refresh flow to ensure the value passed to `CreateRefreshTokenStorage` has exactly the same scopes as originally requested/granted.

This will mean the refresh token session will have static scopes in hydra, but the other session types may have narrower scopes than it at any point in the lifetime.

I think this may address some of your other points, and/or may give you a better idea about how we can achieve the desired result. The intention is between `HandleTokenEndpointRequest` and `PopulateTokenEndpointResponse` in the `RefreshTokenGrantHandler`, that we can restore the original scopes of specifically the refresh token session.

The original code is just passing the `Request` struct to the `CreateRefreshTokenStorage` method of `RefreshTokenStorage`, as per the following code section:

```go
	storeReq := requester.Sanitize([]string{})
	storeReq.SetID(ts.GetID())

	if err = c.TokenRevocationStorage.CreateAccessTokenSession(ctx, accessSignature, storeReq); err != nil {
		return err
	}

	if err = c.TokenRevocationStorage.CreateRefreshTokenSession(ctx, refreshSignature, storeReq); err != nil {
		return err
	}
```

That being said, I have just realized over-engineered this a bit. We could just do the following:

```go
	storeReq := requester.Sanitize([]string{})
	storeReq.SetID(ts.GetID())

	if err = c.TokenRevocationStorage.CreateAccessTokenSession(ctx, accessSignature, storeReq); err != nil {
		return err
	}

	if rtRequest, ok := requester.(fosite.RefreshTokenAccessRequester); ok {
		rtStoreReq := requester.Sanitize([]string{}).(*fosite.Request)
		rtStoreReq.SetID(ts.GetID())

		rtStoreReq.RequestedScope = rtRequest.GetRefreshTokenRequestedScopes()
		rtStoreReq.GrantedScope = rtRequest.GetRefreshTokenGrantedScopes()

		if err = c.TokenRevocationStorage.CreateRefreshTokenSession(ctx, refreshSignature, rtStoreReq); err != nil {
			return err
		}
	} else {
		if err = c.TokenRevocationStorage.CreateRefreshTokenSession(ctx, refreshSignature, storeReq); err != nil {
			return err
		}
	}
```

aeneasr: Awesome, thank you for the explainer! I think the next step is to create a PR in hydra, update it's go.mod for ory/fosite to cda9c24ce1a981f60d6c2e99b7fceec1654c807d and add an e2e test for this new feature. If that and OIDC conformity pass, I think we're good to go! :)

james-d-elliott: Hey @aeneasr sorry for the delay. This is now done in https://github.com/ory/hydra/pull/3403. Le me know if any adjustments are needed and if you want this test merged. I've fixed the lints locally I believe.

handler/oauth2/flow_refresh.go
```diff
@@ -151,8 +185,20 @@ func (c *RefreshTokenGrantHandler) PopulateTokenEndpointResponse(ctx context.Con
 		return err
 	}
 
-	if err = c.TokenRevocationStorage.CreateRefreshTokenSession(ctx, refreshSignature, storeReq); err != nil {
-		return err
+	if rtRequest, ok := requester.(fosite.RefreshTokenAccessRequester); ok {
+		rtStoreReq := requester.Sanitize([]string{}).(*fosite.Request)
```
vivshankar, line 189: @james-d-elliott Wondering if this negates the idea of having a Requester interface at all. I understand why you do it though.

A better approach might be to introduce an interface that allows you to invoke SetGrantedScopes and SetRequestedScopes?

handler/oauth2/flow_refresh.go
```diff
@@ -151,8 +185,20 @@ func (c *RefreshTokenGrantHandler) PopulateTokenEndpointResponse(ctx context.Con
...
+		rtStoreReq.SetID(ts.GetID())
+
+		rtStoreReq.RequestedScope = rtRequest.GetRefreshTokenRequestedScopes()
```
vivshankar, line 192: Is this not already available with `ts` in this function? Do we need to perform the added tasks of populating the requester with the original requested scopes and granted scopes as rt scopes in `HandleTokenEndpointRequest`?

REF: `ts, err := c.TokenRevocationStorage.GetRefreshTokenSession(ctx, signature, nil)`

vivshankar (changes requested):
@james-d-elliott Thanks for this PR! I had a couple of observations that you might consider.

handler/oauth2/flow_refresh.go
james-d-elliott: Yeah, the only intent behind this was to avoid breaking any other implementation. I'm completely fine with which ever approach most fits with how ory is using this however.

handler/oauth2/flow_refresh.go
james-d-elliott: Potentially, let me run some tests to confirm. I suspect I did this intentionally but I can't recall why and since it's not clear and I can't remember and it's important to clear that up one way or another.

handler/oauth2/flow_refresh.go
james-d-elliott: Yeah, the only intent behind this was to avoid breaking any other implementation. I'm completely fine with which ever approach most fits with how ory is using this however.

vivshankar: In the spirit of preserving the interfaces, I suggest a new interface be added that is able to mutate scopes in the requester.

handler/oauth2/flow_refresh.go
james-d-elliott: So the reason I believe did this and it actually warrants a discussion is that the difference between these two methods is that one will utilize the original Refresh Token requested at (as well as other values) value and one utilizes a new one (based on the time of the request). Realistically we're issuing a new token so at least a vast majority of the values should be based on the new request itself.

One curious question that maybe we should be asking is shouldn't the JTI be different since it's effectively a brand new token?

james-d-elliott: Small note which is not a strong argument, but the way this is formed in the new flow is identical in effective functionality of current behavior. See here: https://github.com/ory/fosite/blob/master/handler/oauth2/flow_refresh.go#L136-L156

handler/oauth2/flow_refresh.go
james-d-elliott: This has been addressed in a use specific way I believe.

access_request.go
james-d-elliott: This has been removed as a discussion and code examples from Vivek highlighted it was not necessary.

access_request.go
james-d-elliott: Same as above.

handler/oauth2/flow_refresh.go
james-d-elliott: Resolved, understand what you mean now.

vivshankar (approved):
LGTM. Approving FWIW :-)

handler/oauth2/flow_refresh.go
```diff
@@ -30,6 +29,11 @@ type RefreshTokenGrantHandler struct {
 		fosite.AudienceStrategyProvider
 		fosite.RefreshTokenScopesProvider
 	}
+
+	// IgnoreRequestedScopeNotInOriginalGrant determines the action to take when the requested scopes in the refresh
+	// flow were not originally granted. If false which is the default the handler will automatically return an error.
+	// If true the handler will filter out / ignore the scopes which were not originally granted.
+	IgnoreRequestedScopeNotInOriginalGrant bool
```
james-d-elliott, line 36: Side note when looking at this. I have a per-client implementation of this utilizing an interface. Maybe this is more desirable?

handler/oauth2/flow_refresh.go
```diff
@@ -78,13 +81,46 @@ func (c *RefreshTokenGrantHandler) HandleTokenEndpointRequest(ctx context.Contex
...
+	for _, scope := range request.GetRequestedScopes() {
+		if !strategy(originalScopes, scope) {
+			if c.IgnoreRequestedScopeNotInOriginalGrant {
+				// Skips addressing point 2 of the text in RFC6749 Section 6 and instead just prevents the scope
+				// requested from being granted.
+				continue
+			} else {
+				// Addresses point 2 of the text in RFC6749 Section 6.
+				return errorsx.WithStack(fosite.ErrInvalidScope.WithHintf("The requested scope '%s' was not originally granted by the resource owner.", scope))
+			}
```
james-d-elliott, line 117: Side note when looking at this. I have a per-client implementation of this utilizing an interface. Maybe this is more desirable?

james-d-elliott (commented):
RFC

handler/oauth2/flow_refresh.go
mitar: What would be the use case of having this per-client?

handler/oauth2/flow_refresh.go
mitar: Why is this not made through a new config interface?

mitar: To me it looks like currently all config is done through config interface, even if it is not meant for hot reloading?

mitar (approved):

handler/oauth2/flow_refresh.go
james-d-elliott: The only way I'll personally use it is per-client where the default is to enforce the spec and you make an exception on a per-client basis.

handler/oauth2/flow_refresh.go
```diff
@@ -134,26 +174,36 @@ func (c *RefreshTokenGrantHandler) PopulateTokenEndpointResponse(ctx context.Con
...
+	storeReq.SetID(originalRequest.GetID())
 
 	if err = c.TokenRevocationStorage.CreateAccessTokenSession(ctx, accessSignature, storeReq); err != nil {
 		return err
 	}
 
-	if err = c.TokenRevocationStorage.CreateRefreshTokenSession(ctx, refreshSignature, storeReq); err != nil {
-		return err
+	if rtRequest, ok := requester.(fosite.RefreshTokenAccessRequester); ok {
+		rtStoreReq := rtRequest.SanitizeRestoreRefreshTokenOriginalRequester(originalRequest)
```
mitar, line 196: What is this change about? Why it is necessary?

james-d-elliott: Because the scopes of the current request are for the access token, not the refresh token. The scopes of the original granted refresh token must be restored to make it possible to meet the strict requirements of the spec:

> If a  new refresh token is issued, the refresh token scope MUST be identical to that of the refresh token included by the client in the request.

mitar: No, I mean, my question is why you have two cases here. It looks like you added `RefreshTokenAccessRequester` for backwards compatibility? But it feels to me that if you do not implement `RefreshTokenAccessRequester` then you get wrong behavior? Not sure if we should keep the old behavior at all?

Also, I am not sure that an extra method is already needed here? `Sanitize` returns a clone of requester to begin with? And you then call just:

```
ar.SetID(requester.GetID())
ar.SetRequestedScopes(requester.GetRequestedScopes())
ar.SetGrantedScopes(requester.GetGrantedScopes())
```

So why not just call `originalRequest.Sanitize(nil)` and then those three calls above, here, at this point? Without a method?

Probably I am missing something, I just checked very quickly this.

mitar: It has `GrantScope`? Isn't this enough? You loop over all granted scopes and call `GrantScope`?

mitar: I see the same being done currently in the code:

```
	request.SetID(originalRequest.GetID())
	request.SetSession(originalRequest.GetSession().Clone())
	request.SetRequestedScopes(originalRequest.GetRequestedScopes())

		if !c.Config.GetScopeStrategy(ctx)(request.GetClient().GetScopes(), scope) {
			return errorsx.WithStack(fosite.ErrInvalidScope.WithHintf("The OAuth 2.0 Client is not allowed to request scope '%s'.", scope))
		}
		request.GrantScope(scope)
	}
```

mitar: I see this was a previous comment here: https://github.com/ory/fosite/pull/718#discussion_r1255198323

mitar: So to seems there is already some code to clone a requester. Maybe that should be just moved out into utility function and reused here for refresh token as well?

(And again sorry if I am missing something.)

PR #724, opened 2022-11-21
fix: write authorize error returns 303 for get method

james-d-elliott:
This presumably fixes an issue where the WriteAuthorizeError function sets a 303 See Other when it should return a 302 Found instead when the method of the original HTTP request used the method verb 'GET'. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302 and https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303. The 303 See Other is typically only used when the client is being instructed to change the request method verb to 'GET' which is not relevant if it is already 'GET'.

## Checklist

### DIFF
--- a/authorize_error.go
+++ b/authorize_error.go
@@ -83,5 +83,11 @@ func (f *Fosite) WriteAuthorizeError(ctx context.Context, rw http.ResponseWriter
 	}
 
 	rw.Header().Set("Location", redirectURIString)
-	rw.WriteHeader(http.StatusSeeOther)
+
+	switch ar.GetRequestMethod() {
+	case http.MethodGet:
+		rw.WriteHeader(http.StatusFound)
+	default:
+		rw.WriteHeader(http.StatusSeeOther)
+	}
 }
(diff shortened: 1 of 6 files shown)


james-d-elliott:
For reference this is *NOT* ready to merge. It needs to be carefully examined against the specs; and I'm far less convinced it's "correct" after reading all of the linked specs that it is actually the correct move. For the sake of security I'm leaning towards the implementer needs to make this adjustment.

mitar:
So https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics-25#name-307-redirect is pretty clear that 303 (see other) should simply always be returned:

> If an HTTP redirection (and not, for example, JavaScript) is used for such a request, the authorization server SHOULD use HTTP status code 303 (See Other).

So I am not completely convinced that the option to return 302 (found) is needed, but it seems some people reported that their clients do not work with 303. So having 302 be returned for GET seems a reasonable thing for them, but we should get feedback from them if this really fixes things.

authorize_error.go
```diff
@@ -83,5 +83,11 @@ func (f *Fosite) WriteAuthorizeError(ctx context.Context, rw http.ResponseWriter
 	}
 
 	rw.Header().Set("Location", redirectURIString)
-	rw.WriteHeader(http.StatusSeeOther)
+
+	switch ar.GetRequestMethod() {
```
mitar, line 87: I think you should add a comment explaining and linking why is this necessary.

mitar (approved):

PR #733, opened 2023-01-04
feat: customizable token prefix

mattslocum:
matching Hydra PR: https://github.com/ory/hydra/pull/3407

https://github.com/ory/fosite/pull/675 added `ory_at|pt|ac` prefixes to HMAC tokens. This is a good feature for detecting tokens in places they shouldn't exist. However, for security minded self-hosted implementations, it is often preferred to avoid underlying tooling providers. In this case, it is preferrable to be able to customize the `ory` name and replace it with another key. 

This PR adds ability to customize the token prefix before `at|pt|ac`. 

Related to https://github.com/ory/fosite/pull/675 and https://github.com/ory/hydra/issues/2845.

## Checklist

## Further comments
For backwards compatibility for integrations that already upgraded, it continues to remove the ory specific prefix when trimming.

### DIFF
--- a/compose/compose_strategy.go
+++ b/compose/compose_strategy.go
@@ -22,6 +22,7 @@ type CommonStrategy struct {
 type HMACSHAStrategyConfigurator interface {
 	fosite.AccessTokenLifespanProvider
 	fosite.RefreshTokenLifespanProvider
+	fosite.TokenPrefixProvider
 	fosite.AuthorizeCodeLifespanProvider
 	fosite.TokenEntropyProvider
 	fosite.GlobalSecretProvider
--- a/config.go
+++ b/config.go
@@ -34,6 +34,12 @@ type AccessTokenLifespanProvider interface {
 	GetAccessTokenLifespan(ctx context.Context) time.Duration
 }
 
+// TokenPrefixProvider returns the provider for configuring the token prefix.
+type TokenPrefixProvider interface {
+	// GetAccessTokenLifespan returns the access token lifespan.
+	GetTokenPrefixProvider(ctx context.Context) string
+}
+
 // IDTokenLifespanProvider returns the provider for configuring the ID token lifespan.
 type IDTokenLifespanProvider interface {
 	// GetIDTokenLifespan returns the ID token lifespan.
--- a/config_default.go
+++ b/config_default.go
@@ -26,6 +26,7 @@ var (
 	_ AuthorizeCodeLifespanProvider                = (*Config)(nil)
 	_ RefreshTokenLifespanProvider                 = (*Config)(nil)
 	_ AccessTokenLifespanProvider                  = (*Config)(nil)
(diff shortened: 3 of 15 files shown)


james-d-elliott:
Wouldn't this be easier via a newup function which sets the prefix field then provide the result as the core strategy?

mattslocum:
> Wouldn't this be easier via a newup function which sets the prefix field then provide the result as the core strategy?

Yeah, I agree that it seems like it should be easier. I was trying to match what I perceived as the pattern I found in https://github.com/ory/fosite/blob/master/config.go and https://github.com/ory/fosite/blob/master/config_default.go, and then I chased a couple of those examples and tried to mimic them. Can you expound on the better way or is there a similar example?
If you are referring to the `strategy_hmacsha.go` change only, then I can see how that is odd. `@aeneas` made the original change that added the private field, and I was struggling to figure out the full purpose. https://github.com/ory/fosite/commit/b652335c965d5cc523faebad9c9792c4135cfb75#diff-180026ce292f76d8c418fa1d6c09ef637ffd461c5772ec9441b1131642965f6f But I'm happy to refactor that field to be a different style.

james-d-elliott:
Yeah it makes sense why you did it. I think the intention in the pattern was primarily for hot reloading, whereas this probably isn't something that would benefit much from it. We'll see what Mr Hackerman thinks and if he has a preference.

vivshankar:
@aeneasr I know you have previously stated that Fosite is only meant to be used with Hydra but that wasn't my understanding when we adopted it for our own implementation. I believe there are others in the same boat.

Would you accept a change where the prefix can be set globally - maybe just a change from `setPrefix` to `SetPrefix` to let it be set over a public function? It doesn't make sense for our tokens to have a "ory_" prefix.

The alternative is for us to effectively clone the HMACStrategy and write our own without the prefix and that was my plan until this PR was submitted.

K3das:
Though I don't think it's feasible to try to hide your usage of hydra, I do agree with @mattslocum, especially considering that services reporting secret leaks will only report to ory, which isn't at all helpful in the case of a self-hosted hydra, where you'd want those services reporting to you...

mitar:
I also opened #789 for this. Missed this PR. I am also +1 for having this be configurable. There are also +8 upvotes on this PR. @aeneasr please reconsider this.

mitar:
I think this has been done with https://github.com/ory/fosite/pull/816.

mslocumjc:
If this was closed in response to Fosite's https://github.com/ory/fosite/pull/816, then I don't think that is accurate since Hydra is still not customizable. I still think this would be a valuable feature for independently hosted instances to supply a unique prefix so that way ory owned tokens aren't confused with tokens minted through a self-hosted system.

mitar:
For Hydra you should open an issue in Hydra repository.

PR #737, opened 2023-02-22
feat: add the ability to set jwt header type

mgyongyosi:
First, thank you for all your work on this library! 🙇 

We are implementing [RFC9068](https://datatracker.ietf.org/doc/rfc9068/) and during that we found the following issue: the JWT header's `typ` parameter is implicitly set to "JWT" inside token.go by the the `unsignedToken` function. Also the `ToMap` function filters out the `typ` as well and unfortunately it cannot be changed from the Session for example. And based on https://datatracker.ietf.org/doc/html/rfc9068#section-2.1-3 the `typ` must be set to `at+jwt`. 

This PR adds the ability to set the `typ` parameter from the `jwt.Headers` and when the `typ` is not set explicitly then the `typ` will be set to "JWT".

## Checklist

### DIFF
--- a/token/jwt/header.go
+++ b/token/jwt/header.go
@@ -14,7 +14,7 @@ func NewHeaders() *Headers {
 
 // ToMap will transform the headers to a map structure
 func (h *Headers) ToMap() map[string]interface{} {
-	var filter = map[string]bool{"alg": true, "typ": true}
+	var filter = map[string]bool{"alg": true}
 	var extra = map[string]interface{}{}
 
 	// filter known values from extra.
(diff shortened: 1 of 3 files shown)


aeneasr (approved):

aeneasr:
Thank you for the PR! This looks quite good to me :) Unfortunately, OIDC Conformity tests are failing. I think they are also failing on master, but given this change we need them to be green. I'll spend a bit of time to see if I can fix this on master but if not, merging this PR is blocked by that requirement :(

mgyongyosi:
> Thank you for the PR! This looks quite good to me :) Unfortunately, OIDC Conformity tests are failing. I think they are also failing on master, but given this change we need them to be green. I'll spend a bit of time to see if I can fix this on master but if not, merging this PR is blocked by that requirement :(

@aeneasr, is there anything I could do/assist with to fix the failing oidc-conformity tests?

PR #749, opened 2023-05-11
feat: grant requested audience(s) in client_credentials flow

mgyongyosi:
This change ensures that the audience claim is populated from the requested audience(s) when the token request explicitly contains the set of audiences (`audience` parameter) in case of client_credentials request.

## Why?
The `audience` claim is mandatory in access tokens that conform to RFC9068 JWT profile. I found that the audience claim is not populated by the library during a `client_credentials` flow, but fosite checks if the requested audience(s) are allowed by the specified client, so decided to implement the audience granting based on the rfc7523 implementation: https://github.com/ory/fosite/blob/45a6785cc54fcbd9195b0de4b821bb5fed6a41be/handler/rfc7523/handler.go#L104. 

I researched the topic a bit and I haven't found any standard which describes how this should behave (the closest is the proposed RFC8707), however I found that both Hydra (https://github.com/ory/hydra/blob/3703e5a7b1a910a15e2118c9ad5cba93e54e4d3d/oauth2/handler.go#L913) and Auth0 (https://auth0.com/docs/get-started/authentication-and-authorization-flow/call-your-api-using-the-client-credentials-flow#example-post-to-token-url) supports this behavior.

## Checklist

## Further comments

### DIFF
--- a/handler/oauth2/flow_client_credentials.go
+++ b/handler/oauth2/flow_client_credentials.go
@@ -40,6 +40,10 @@ func (c *ClientCredentialsGrantHandler) HandleTokenEndpointRequest(ctx context.C
 		return err
 	}
 
+	for _, audience := range request.GetRequestedAudience() {
+		request.GrantAudience(audience)
+	}
+
 	// The client MUST authenticate with the authorization server as described in Section 3.2.1.
 	// This requirement is already fulfilled because fosite requires all token requests to be authenticated as described
 	// in https://tools.ietf.org/html/rfc6749#section-3.2.1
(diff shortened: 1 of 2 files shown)


james-d-elliott:
t might help if you explain why the implementer should not  do this, the exact points in the RFC you feel is relevant, and why this shouldn't instead be handled with an error response instead.

 I feel like this is a breaking change where if a particular implementer was doing additional checks if an audience should be granted that it could cause serious issues on their end.

mgyongyosi:
> It might help if you explain why the implementer should not do this, the exact points in the RFC you feel is relevant, and why this shouldn't instead be handled with an error response instead.

I haven't found anything in the standards I checked which describe how this should behave. My thought process was that if someone explicitly requests a particular audience then it expects that the granted audiences will contain the requested audience if the requested audience is registered with the OAuth2 client. I might be wrong. 

So, you mean this should be handled by the implementer, right?

PR #760, opened 2023-07-21
fix: basic scheme rejects special characters

### LINKED ISSUE #753: clientCredentialsFromRequest should not expect Basic Authorization terms being URL Escaped
### Preflight checklist

- [ ] This issue affects my [Ory Network](https://www.ory.sh/) project.

### Describe the bug

Around https://github.com/ory/fosite/blob/master/client_authentication.go#L298

I expect `clientCredentialsFromRequest` not doing `QueryUnescape`.

# Rationale

1. Per https://datatracker.ietf.org/doc/html/rfc7617, it does not metion about escaping user and password part. In addition, it specifically mention that username with colon is invalid, which wouldn't be the case if it's escaped.
2. As tested in Chrome, password with `%` is just passed as is, without encoded to `%25`

# Impact

If the client key or client secret contains plaintext "%", it will failed with unable to unescape or never match (unless it's specifically escaped manually in RP configuration)

### Reproducing the bug

1. Create a client credential with `%`
2. Build a token endpoint, and use basic authorization to pass the client secret.
3. It will fails with `The client secret in the HTTP authorization header could not be decoded from 'application/x-www-form-urlencoded'.` or fail to authenticate due to secret mismatch

### Relevant log output

_No response_

### Relevant configuration

_No response_

### Version

latest

None

None

### Additional Context

_No response_

james-d-elliott:
Fixes an issue where the basic auth scheme rejects special characters in the base64 decoded username/password.

Fixes #753

## Checklist

### DIFF
--- a/client_authentication.go
+++ b/client_authentication.go
@@ -7,10 +7,12 @@ import (
 	"context"
 	"crypto/ecdsa"
 	"crypto/rsa"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/ory/x/errorsx"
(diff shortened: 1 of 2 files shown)


aeneasr (changes requested):
I don't think this implementation is correct and in fact breaks BC due to the missing QueryUnescape.

> Clients in possession of a client password MAY use the HTTP Basic
   authentication scheme as defined in [[RFC2617](https://datatracker.ietf.org/doc/html/rfc2617)] to authenticate with
   the authorization server.  The client identifier is encoded using the
   **"application/x-www-form-urlencoded"** encoding algorithm per
   [Appendix B](https://datatracker.ietf.org/doc/html/rfc6749#appendix-B), and the encoded value is used as the username; the client
   password is encoded using the **same algorithm** and used as the
   password.  The authorization server MUST support the HTTP Basic
   authentication scheme for authenticating clients that were issued a
   client password.

james-d-elliott:
Yeah looks like it is correct as is now I read it again. Not entirely sure which element I was reading or if the confusion came from what seemed to be two differing concepts at the time. Thanks!

aeneasr:
No problem :)

PR #762, opened 2023-08-04
fix: openid ignores missing redirect uri

### LINKED ISSUE #685: Investigate missing `redirect_uri` check when performing OIDC flows
### Preflight checklist

### Describe the bug

[Open ID Connect Core specification](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) says:

```
redirect_uri
REQUIRED. Redirection URI to which the response will be sent. This URI MUST exactly match one of the Redirection URI values for the Client pre-registered at the OpenID Provider, with the matching performed as described in Section 6.2.1 of [[RFC3986]](https://openid.net/specs/openid-connect-core-1_0.html#RFC3986) (Simple String Comparison). When using this flow, the Redirection URI SHOULD use the https scheme; however, it MAY use the http scheme, provided that the Client Type is confidential, as defined in Section 2.1 of OAuth 2.0, and provided the OP allows the use of http Redirection URIs in this case. The Redirection URI MAY use an alternate scheme, such as one that is intended to identify a callback into a native application.
```

It does however not appear as if we are checking for this condition, as OAuth2 has the redirect URI marked as optional.

### Reproducing the bug

This should be first confirmed with an integration test case

### Relevant log output

_No response_

### Relevant configuration

_No response_

### Version

-

_No response_

_No response_

### Additional Context

_No response_

james-d-elliott:
Fixes an issue where Authorize Requests which were intended for an OpenID Connect 1.0 client would incorrectly be allowed when missing the redirect URI when it's required by the specification.

Fixes #685

## Checklist

## Further comments

The added tests fail prior to the fix and pass after.


authorize_request_handler.go
```diff
@@ -163,6 +163,17 @@ func (f *Fosite) validateAuthorizeRedirectURI(_ *http.Request, request *Authoriz
...
 	rawRedirURI := request.Form.Get("redirect_uri")
 
+	// This ensures that the 'redirect_uri' parameter is present for OpenID Connect 1.0 authorization requests as per:
+	//
+	// Authorization Code Flow - https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
+	// Implicit Flow - https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthRequest
+	// Hybrid Flow - https://openid.net/specs/openid-connect-core-1_0.html#HybridAuthRequest
+	//
+	// Note: as per the Hybrid Flow documentation the Hybrid Flow has the same requirements as the Authorization Code Flow.
+	if len(rawRedirURI) == 0 && request.GetRequestedScopes().Has("openid") {
```
aeneasr, line 173: Is this not already handled by the openid handler itself?

james-d-elliott: I have added an integration test which shows this is unfortunately not the case. It only returns an error if the client has multiple redirect URI's. The issue is that OAuth 2.0 I believe allows assuming the Redirect URI if absent provided there is only one, however OpenID Connect 1.0 explicitly requires the redirect URI parameter and has no such leeway.

It should also be noted that OAuth 2.1 does not alter this behavior. The only alteration by my reading is in OpenID Connect 1.0.

james-d-elliott: Also see the linked issue, you were actually the person to raise the issue. Maybe that's helpful?

mitar (approved):

authorize_request_handler.go
mitar: I checked this and I agree with @james-d-elliott: OAuth 2.0 says that if you included Redirect URI in authentication request, then it has also be included in the token request. And Fosite currently checks that.

But OIDC spec says that Redirect URI is required in the authentication request. And this PR adds this check.

authorize_request_handler.go
```diff
@@ -363,27 +378,31 @@ func (f *Fosite) newAuthorizeRequest(ctx context.Context, r *http.Request, isPAR
...
 		return request, errorsx.WithStack(ErrRegistrationNotSupported)
 	}
 
-	if err := f.validateResponseTypes(r, request); err != nil {
+	if err = f.validateResponseTypes(r, request); err != nil {
 		return request, err
 	}
 
-	if err := f.validateResponseMode(r, request); err != nil {
+	if err = f.validateResponseMode(r, request); err != nil {
```
mitar, line 405: Why these changes? I think they are unrelated and just change the style?

authorize_request_handler.go
```diff
@@ -363,27 +378,31 @@ func (f *Fosite) newAuthorizeRequest(ctx context.Context, r *http.Request, isPAR
 		return request, err
 	}
 
-	if err := f.validateAuthorizeRedirectURI(r, request); err != nil {
+	if err = f.parseAuthorizeScope(r, request); err != nil {
```
mitar, line 381: I think that instead of splitting `validateAuthorizeScope` into `parseAuthorizeScope` and `validateAuthorizeScope`, it would be easier if you swapped the order of validation, first call (original) `validateAuthorizeScope` and then (updated) `validateAuthorizeRedirectURI`. In a way, if `validateAuthorizeRedirectURI` depends on parsed scopes, it is better if `validateAuthorizeRedirectURI` depends on parsed and validated scopes.

mitar (changes requested):

authorize_request_handler.go
aeneasr: Thank you both - I checked and indeed the expectation is that the redirect-url is properly validated before we call `HandleAuthorizeEndpointRequest`. Given that this is openID specific though, this should be moved to the openid handler instead. I'll do that now

aeneasr: I'm creating a separate PR as this is a breaking change for existing clients and it requires dedicated care for merging that to production on our infrastructure!

aeneasr: As far as I can tell, this implementation does not respect hybrid or implicit flows of OIDC, so it definitely makes sense to move that to the appropriate handlers in my view.

aeneasr:
Closing in favor of https://github.com/ory/fosite/pull/814

authorize_request_handler.go
james-d-elliott: That sounds like a solid plan, glad this will get merged regardless.

PR #763, opened 2023-08-05
feat: move OTEL to a new 'InstrumentedFosite'

vivshankar:
The recent change in [#761 ] introduces OTEL as part of Fosite. While this is valuable, this should
be provided as a wrapper and not as a hard wired part of Fosite, which is meant to be a SDK. Also,
popular instrumentation options and APM vendors have some level of support for OTEL but shine
when their wrapper SDKs are used that do not always comply with OTEL interfaces. See New Relic
and Instana SDKs for reference.

This change proposes that OTEL should be an opt-in as part of a wrapper implementation of Fosite
that is called `InstrumentedFosite` under the `otel` package. It also offers a pattern that can be applied
to any code blocks and handlers that should be instrumented to derive maximum benefits from a telemetry
perspective. For example - there may be no value in instrumenting a code block that interfaces with no
external components beyond the overall telemetry offered at a HTTP middleware level, whereas there
may be a great deal of value instrumenting call outs to external APIs and databases.

#761 

## Checklist

### DIFF
--- a/access_request_handler.go
+++ b/access_request_handler.go
@@ -10,8 +10,6 @@ import (
 
 	"github.com/ory/fosite/i18n"
 	"github.com/ory/x/errorsx"
-	"github.com/ory/x/otelx"
-	"go.opentelemetry.io/otel/trace"
 
 	"github.com/pkg/errors"
 )
@@ -42,9 +40,6 @@ import (
 //     client MUST authenticate with the authorization server as described
 //     in Section 3.2.1.
 func (f *Fosite) NewAccessRequest(ctx context.Context, r *http.Request, session Session) (_ AccessRequester, err error) {
-	ctx, span := trace.SpanFromContext(ctx).TracerProvider().Tracer("github.com/ory/fosite").Start(ctx, "Fosite.NewAccessRequest")
-	defer otelx.End(span, &err)
-
 	accessRequest := NewAccessRequest(session)
 	accessRequest.Request.Lang = i18n.GetLangFromRequest(f.Config.GetMessageCatalog(ctx), r)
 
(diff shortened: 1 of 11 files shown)


james-d-elliott:
Tests seem to just be failing due ory/hydra using a PR in go.mod. I like this change, I did some experimentation with it in making github.com/ory/fosite/otel it's own module so it's entirely standalone however I could not get it working, maybe it's unnecessary due to pruning and github.com/ory/x contains the imports anyway.

aeneasr (changes requested):
Thank you for the PR! I think it's nice to abstract away tracing and at the same time am critical of this approach. We will most likely add more OTEL tracing to lower-level functions and then this approach will no longer work (as well) without significant code wiring.

I think otel is a solid standard and would recommend to just go with it for now. If you want wrapping - you still can do that in your own codebase consuming Fosite!

vivshankar:
The way we abstracted this for future proofing is to introduce a metrics interface that allows you to swap in providers as needed. OTEL is great but is not completely adopted by major vendors (referenced in my post). Also the current implementation takes away any flexibility for adopters to use different libraries.

I am happy to take another stab at this to put it in Fosite directly if that will be acceptable

vivshankar:
@aeneasr Just got to my machine. To substantiate further, we could minimally add a config parameter to control the instrumentation. I was also thinking for handlers to add a wrapper handler that basically does the same thing and it becomes a simple matter of adding that to your compose factory/handler list.

For custom code blocks, it is simple enough to add an interface for spans and transactions.

aeneasr:
I don't think that we want to maintain an abstraction for telemetry in Fosite, especially as we are quite happy with otel at the moment.

If you don't have an otel compatible system, you can set the otel tracer to noop/nil in which case nothing happens. And you can still wrap Fosite any way you want to add instrumentation.

Being honest, I don't the value of doing a lot of work here right now.

Issue #631, opened 2021-10-19
Encrypted JWTs for both id_tokens and client assertions


vivshankar:
**Is your feature request related to a problem? Please describe.**

ID tokens and client assertions (both for JWT bearer grant flow and for private_key_jwt authentication) are generally signed and optionally encrypted. In general, an encrypted token requirement comes up in industries like government or finance.

**Describe the solution you'd like**

The request is to add this capability by default in Fosite with an extension, effectively, to the JWTStrategy.

**Describe alternatives you've considered**

1. Add a handler in the provider after the id_token is generated that encrypts as configured on the client/provider config.
2. Implement a client_authentication strategy that also caters for encrypted JWTs

**Additional context**

N/A

vivshankar:
@aeneasr For the ID token etc., this is relatively easy to accomplish with a custom implementation of JWTStrategy and performing the encryption on the Generate function. Perhaps a couple of client properties should be made available such as the algorithms to be used. I would just match it to the client_metadata properties in the DCR spec.

However, for inbound (client_assertion, request_object etc.), a new `Decrypt` function would be necessary in the JWTStrategy or a new `EncryptedJWTStrategy`. Again, the client properties should ideally match the DCR spec.

This should be a relatively simple extension with a simple example. Please let me know if this makes sense.

aeneasr:
Sorry for the late reply, that sounds like a good approach!

vivshankar:
@aeneasr This is a regarding a change I see in the master branch.

```
type Signer interface {
	Generate(ctx context.Context, claims MapClaims, header Mapper) (string, string, error)
	Validate(ctx context.Context, token string) (string, error)
	Hash(ctx context.Context, in []byte) ([]byte, error)
	Decode(ctx context.Context, token string) (*Token, error)
	GetSignature(ctx context.Context, token string) (string, error)
	GetSigningMethodLength(ctx context.Context) int
}
```

It looks like it is just a name change from `JWTStrategy` to `Signer`. I am curious about the thought process behind this and changes that might have been proposed/are being made.

aeneasr:
Primarily because it is only used to sign tokens currently, but we can also name it something else

PR #764, opened 2023-08-05
feat: [#631] JWT Encryption support for client authentication

vivshankar:
See #631 

See #631 

## Checklist

### DIFF
--- a/client.go
+++ b/client.go
@@ -71,6 +71,17 @@ type OpenIDConnectClient interface {
 	GetTokenEndpointAuthSigningAlgorithm() string
 }
 
+// ClientWithAllowedVerificationKeys adds a security control to the client configuration to only allow
+// specific verification keys. This ensures that a key that is valid for client X can't be used for client Y
+// unless allowed. This becomes especially important for cases where the clients are controlled by third-parties
+// and are issued specific keys from a central organization, which may be the OP's org or a central regulatory authority,
+// and the security controls of the clients cannot be guaranteed.
+type ClientWithAllowedVerificationKeys interface {
+	// AllowedVerificationKeys provides a list of key IDs that can be used in the JWT
+	// header for private_key_jwt authentication and for JWT bearer grant flow
+	AllowedVerificationKeys() []string
+}
+
 // ResponseModeClient represents a client capable of handling response_mode
 type ResponseModeClient interface {
 	// GetResponseMode returns the response modes that client is allowed to send
--- a/client_authentication.go
+++ b/client_authentication.go
@@ -8,7 +8,6 @@ import (
 	"crypto/ecdsa"
 	"crypto/rsa"
 	"encoding/json"
-	"fmt"
 	"net/http"
 	"net/url"
 	"time"
@@ -16,7 +15,6 @@ import (
 	"github.com/ory/x/errorsx"
 
 	"github.com/go-jose/go-jose/v3"
-	"github.com/pkg/errors"
 
 	"github.com/ory/fosite/token/jwt"
 )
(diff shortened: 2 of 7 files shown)


james-d-elliott (commented):
Looking really good from my perspective. I am wondering if we want to add an interface for client ID Token encryption options similar to the OpenIDConnectClient interface (specifically the GetRequestObjectSigningAlgorithm implementation). But I'm not entirely sure this is the best way (just sharing ideas as I have no concrete opinion).

vivshankar:
@aeneasr Before I go too far down this rabbit hole, I wanted your opinion on replacing the current mechanism for validating JWTs that effectively relies on a KeyFunc returning a key, typically from jwks/jwks_uri config. In the implementation I am introducing here, starting with client authentication, I am expanding support to do the following -

- Preserve current behavior to support jwks/jwks_uri based validation
- Decryption for JWEs
- Custom validation strategy for JWTs that might, for example, leverage a trust store or a key management service that offers crypto APIs etc.
- Introducing some security controls around "allowed key IDs" and "allowed algs". The former is important IMO as seen in a recent hack. The latter is useful to enforce requirements for PS256 and ES256 type algorithms required by FAPI and other high assurance flows.

I have a few cases where the approach I am introducing here applies or will apply -

1. Request object
2. DCR with SSA validation and DCR client_metadata as application/jose (AU-CDR, for example, uses this)
3. Token exchange for custom JWT type as a subject or actor token
4. JWT bearer grant flow (would be an enhancement on top of what is available today)

In all cases, I am trying to preserve the current behavior (unit tests confirm it) while adding this extra option to validate the JWT but it may effectively negate the need for some functions that are in use today. It also introduces decryption as part of the same set of changes.

Note here that I am specifically referring to incoming JWTs. The PR also carries a mechanism to encrypt outgoing JWTs for id_tokens, JARM, userinfo as JWT etc. I know some of these aren't yet in place but this sets the foundation to add those capabilities.

I am, by no means, done with this PR though it is ready for review.

james-d-elliott:
I believe the following is a spec compliant way to derive symmetric key types:

```go
package example

import (
	"crypto/aes"
	"crypto/sha256"
	"crypto/sha512"
	"fmt"
	"hash"

	"github.com/go-jose/go-jose/v4"
)

func DeriveSymmetricKey(secret []byte, kid, alg, enc, use string) (jwk *jose.JSONWebKey, err error) {
	if len(secret) == 0 {
		return nil, fmt.Errorf("error occurred deriving symmetric jwk: client secret is not configured")
	}

	switch use {
	case "sig":
		return &jose.JSONWebKey{
			Key:       secret,
			KeyID:     kid,
			Algorithm: alg,
			Use:       use,
		}, nil
	case "enc":
		var (
			hasher hash.Hash
			bits   int
		)

		keyAlg := jose.KeyAlgorithm(alg)

		switch keyAlg {
		case jose.A128KW, jose.A128GCMKW, jose.A192KW, jose.A192GCMKW, jose.A256KW, jose.A256GCMKW, jose.PBES2_HS256_A128KW:
			hasher = sha256.New()
		case jose.PBES2_HS384_A192KW:
			hasher = sha512.New384()
		case jose.PBES2_HS512_A256KW, jose.DIRECT:
			hasher = sha512.New()
		default:
			return nil, fmt.Errorf("error occurred deriving symmetric jwk: the encryption key algorithm '%s' is not supported", enc)
		}

		switch keyAlg {
		case jose.A128KW, jose.A128GCMKW, jose.PBES2_HS256_A128KW:
			bits = aes.BlockSize
		case jose.A192KW, jose.A192GCMKW, jose.PBES2_HS384_A192KW:
			bits = aes.BlockSize * 1.5
		case jose.A256KW, jose.A256GCMKW, jose.PBES2_HS512_A256KW:
			bits = aes.BlockSize * 2
		case jose.DIRECT:
			switch jose.ContentEncryption(enc) {
			case "":
				enc = string(jose.A128CBC_HS256)
				fallthrough
			case jose.A128CBC_HS256:
				bits = aes.BlockSize * 2
			case jose.A192CBC_HS384:
				bits = aes.BlockSize * 3
			case jose.A256CBC_HS512:
				bits = aes.BlockSize * 4
			default:
				return nil, fmt.Errorf("error occurred deriving symmetric jwk: the encryption key algortihm '%s' does not support content encryption '%s'", alg, enc)
			}
		}

		if _, err = hasher.Write(secret); err != nil {
			return nil, fmt.Errorf("error occurred deriving symmetric jwk: error occurred hasing the secret: %w", err)
		}

		return &jose.JSONWebKey{
			Key:       hasher.Sum(nil)[:bits],
			KeyID:     kid,
			Algorithm: alg,
			Use:       use,
		}, nil
	default:
		return nil, fmt.Errorf("error occurred deriving symmetric jwk: the use '%s' is not supported", use)
	}
}
```

PR #775, opened 2023-11-10
fix: remove redundant map access

dtam-cybozu:
Remove a redundant check for whether an element is present in a map. Since we return the zero value of a `time.Time` anyway, it's simpler to use the single return value version to access the map.
## Checklist

I have not added any documentation or tests for this since it is such a small fix.

### DIFF
--- a/session.go
+++ b/session.go
@@ -52,9 +52,6 @@ func (s *DefaultSession) GetExpiresAt(key TokenType) time.Time {
 		s.ExpiresAt = make(map[TokenType]time.Time)
 	}
 
-	if _, ok := s.ExpiresAt[key]; !ok {
-		return time.Time{}
-	}
 	return s.ExpiresAt[key]
 }
 
--- a/session_test.go
+++ b/session_test.go
@@ -15,3 +15,11 @@ func TestSession(t *testing.T) {
 	assert.Empty(t, s.GetUsername())
 	assert.Nil(t, s.Clone())
 }
+
+func TestZeroSession(t *testing.T) {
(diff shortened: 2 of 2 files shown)


aeneasr (changes requested):
I believe that this will panic if the key does not exist

dtam-cybozu:
> I believe that this will panic if the key does not exist

I don't think this is right. See https://go.dev/ref/spec#Index_expressions

> For a of [map type](https://go.dev/ref/spec#Map_types) M:
>
> - x's type must be [assignable](https://go.dev/ref/spec#Assignability) to the key type of M
> - if the map contains an entry with key x, a[x] is the map element with key x and the type of a[x] is the element type of M
> - if the map is nil or does not contain such an entry, a[x] is the [zero value](https://go.dev/ref/spec#The_zero_value) for the element type of M

Here is a Go Playground example https://go.dev/play/p/udUlbIUWIfq

james-d-elliott:
Yeah this will not panic and will return exactly the same output. I would imagine there is a test in this scenario or one could be easily added to prove unchanged function.

dtam-cybozu:
BTW @aeneasr if the session itself is nil then a call to `GetExpiresAt` *will* panic but this is a separate discussion.

Anyway, I've added a test to show this does not panic.

aeneasr (approved):

aeneasr:
Thank you for the follow up - I think I was confused because it would panic if we had a pointer and would try to access it. Since we don't have either, this is fine! :)

PR #791, opened 2024-02-03
fix: pass standard params to storage

aeneasr:
## Checklist

### DIFF
--- a/request.go
+++ b/request.go
@@ -159,10 +159,12 @@ func (a *Request) Merge(request Requester) {
 	}
 }
 
+var defaultAllowedParameters = []string{"grant_type", "response_type", "scope", "client_id"}
+
 func (a *Request) Sanitize(allowedParameters []string) Requester {
 	b := new(Request)
 	allowed := map[string]bool{}
-	for _, v := range allowedParameters {
+	for _, v := range append(allowedParameters, defaultAllowedParameters...) {
 		allowed[v] = true
 	}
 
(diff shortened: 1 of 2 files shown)


request.go
```diff
@@ -159,10 +159,12 @@ func (a *Request) Merge(request Requester) {
 	}
 }
 
+var defaultAllowedParameters = []string{"grant_type", "response_type", "scope", "client_id"}
```
mitar, line 162: In my own code, I use the following list:

```
			"max_age", "prompt", "acr_values", "id_token_hint", "nonce",
			// PCRE.
			"code_challenge", "code_challenge_method",
			// Other fields.
			"display", "ui_locales", "login_hint",
```

mitar: BTW, this is not really a "default" list (you cannot set it to non-default) but an "always allowed list".

aeneasr: > BTW, this is not really a "default" list (you cannot set it to non-default) but an "always allowed list".

True :)

I would probably drop `id_token_hint` because it contains a valid ID token

mitar: Hm, thanks. You are right. I am storing that to support logout. But I should just store the subject ID, not the whole token.

mitar: In fact, `id_token_hint` is from this list in fosite: https://github.com/ory/fosite/blob/master/handler/openid/flow_explicit_auth.go#L29-L35

So if you are saying this should not be stored into the database? Because fosite currently does so:

```
	if err := c.OpenIDConnectRequestStorage.CreateOpenIDConnectSession(ctx, resp.GetCode(), ar.Sanitize(oidcParameters)); err != nil {
		return errorsx.WithStack(fosite.ErrServerError.WithWrap(err).WithDebug(err.Error()))
	}
```

mitar: Maybe this means that subject should be extracted out earlier from the `id_token_hint` instead of being passed to the store inside form?

PR #793, opened 2024-02-12
fix: call DeleteOpenIDConnectSession during successful authcode exchange

### LINKED ISSUE #790: openid session storage should be deleted when the authcode is exchanged
### Preflight checklist

### Ory Network Project

_No response_

### Describe the bug

I work on a project that use fosite as a golang package and implements the fosite storage interfaces in a custom way. The `DeleteOpenIDConnectSession` interface has never been called and more recently has been marked deprecated. However, this is a problem for the fosite storage interfaces.

The storage interfaces are very nicely designed and gives implementors lots of flexibility in how the underlying storage works. In our project, for several technical reasons (including that we are not using SQL), each storage interface gets stored in the underlying datastore separately, with no connection to each other. Because the openid sessions are never deleted, we have no choice but to allow them to linger for longer than is actually necessary, using up storage in the underlying datastore.

If I understand the code correctly, there is no reason to keep the openid session data after the authcode is successfully exchanged [in `PopulateTokenEndpointResponse` of flow_explicit_token.go](https://github.com/ory/fosite/blob/f4114878826c6d26b6751a1ea61f69ebbc25d4f2/handler/openid/flow_explicit_token.go#L20), so that function could call `DeleteOpenIDConnectSession` before it returns.

I understand that other implementations of the storage interface might choose to store the openid session data along with the authcode session data (e.g. together in a SQL table), and therefore incur less overhead in the underlying datastore. It's nice that the fosite storage interface gives that flexibility. However, the storage interfaces should not require or assume that style of implementation, and none of the other fosite storage interfaces require it except for the openid interface.

Might you consider accepting a PR which removes the deprecation notice on `DeleteOpenIDConnectSession` and adds a call to that function in the appropriate place? Storage interfaces which do not need it could simply implement it as an empty (no-op) function, which they are most likely already doing anyways.

### Reproducing the bug

In the source code, you can see that the `DeleteOpenIDConnectSession` is never called, and is marked deprecated.

### Relevant log output

_No response_

### Relevant configuration

_No response_

### Version

v0.44.0

None

None

### Additional Context

_No response_

cfryanr:
Fixes https://github.com/ory/fosite/issues/790. Please see description and comments in https://github.com/ory/fosite/issues/790 for more information about why this is a necessary and safe change.

```
Remove deprecation of `DeleteOpenIDConnectSession` storage interface function and call it during
authorization code exchange. This function was not previously called. Implementors of the openid
storage interface who which to preserve the old behavior should implement this function as a
no-op which returns `nil`.
```

Fixes #790.

## Checklist

- [X] I have referenced an issue containing the design document if my change introduces a new feature.
- [X] I have added the necessary documentation within the code base (if appropriate).

### DIFF
--- a/handler/openid/flow_explicit_token.go
+++ b/handler/openid/flow_explicit_token.go
@@ -22,7 +22,9 @@ func (c *OpenIDConnectExplicitHandler) PopulateTokenEndpointResponse(ctx context
 		return errorsx.WithStack(fosite.ErrUnknownRequest)
 	}
 
-	authorize, err := c.OpenIDConnectRequestStorage.GetOpenIDConnectSession(ctx, requester.GetRequestForm().Get("code"), requester)
+	authorizeCode := requester.GetRequestForm().Get("code")
+
+	authorize, err := c.OpenIDConnectRequestStorage.GetOpenIDConnectSession(ctx, authorizeCode, requester)
 	if errors.Is(err, ErrNoSessionFound) {
 		return errorsx.WithStack(fosite.ErrUnknownRequest.WithWrap(err).WithDebug(err.Error()))
 	} else if err != nil {
@@ -45,6 +47,11 @@ func (c *OpenIDConnectExplicitHandler) PopulateTokenEndpointResponse(ctx context
 	claims := sess.IDTokenClaims()
 	if claims.Subject == "" {
 		return errorsx.WithStack(fosite.ErrServerError.WithDebug("Failed to generate id token because subject is an empty string."))
+	}
+
+	err = c.OpenIDConnectRequestStorage.DeleteOpenIDConnectSession(ctx, authorizeCode)
+	if err != nil {
+		return errorsx.WithStack(fosite.ErrServerError.WithWrap(err).WithDebug(err.Error()))
 	}
 
 	claims.AccessTokenHash = c.GetAccessTokenHash(ctx, requester, responder)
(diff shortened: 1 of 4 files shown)


cfryanr:
The format check seems to have failed complaining about almost every copyright date in the repo. Please advise if I should update any copyright dates in this PR.

aeneasr (approved):

aeneasr:
Nice! I recently found the same issue on our prod system. I think we can also delete the PKCE entry if it's not already being deleted

james-d-elliott:
Small aside, this makes me think there is an issue with the tests since there is only one mock EXPECT added for `DeleteOpenIDConnectSession`.

cfryanr:
Thanks for accepting the PR.

> I think we can also delete the PKCE entry if it's not already being deleted

I believe those are already correctly deleted here https://github.com/ory/fosite/blob/v0.46.0/handler/pkce/handler.go#L133.

> Small aside, this makes me think there is an issue with the tests since there is only one mock EXPECT added for DeleteOpenIDConnectSession.

I added two EXPECTs in the PR. One for each pre-existing happy path unit test.

aeneasr:
You're right - those are indeed already covered!

james-d-elliott:
> I added two EXPECTs in the PR. One for each pre-existing happy path unit test.

Oh your right my bad. I swore when I looked at the "should pass" test it didn't have one. Was really confusing why it was passing without it. But it's there!

PR #795, opened 2024-02-13
fix: use stdlib to check loopback address

mitar:
See: https://github.com/ory/fosite/commit/44fd2cc2356954292d435e45712624d4145debc7#r138568658


aeneasr (approved):

mitar:
Thanks!

PR #806, opened 2024-03-29
feat: add ResourceAccessScopeStrategy

zoop-btc:
Related to [this hydra issue](https://github.com/ory/hydra/issues/3751).

This scope strategy splits a scope into resources delimited with `:`.
Each resource can have dynamic values if it has the suffix `-*`.
The last resource can specify a verb delimited with `.`.

The hierarchy of the matcher and scope resources needs to be identical.

Examples:
* `users.*` matches `users.read`
* `users.write` does not match `users.read`
* `users:settings` matches `users:settings`
* `users:settings` does not match `users:settings.read`
* `users:client-*.read` does match `users:client-bar.read`
* `users:client-*` does not match `users:client-bar.read`
* `users:client-*.*` does match `users:client-foo.write`

Open questions:
Should I make the resource/verb delimiter configurable? What about allowing dynamic resources?

### DIFF
--- a/scope_strategy.go
+++ b/scope_strategy.go
@@ -83,3 +83,69 @@ func WildcardScopeStrategy(matchers []string, needle string) bool {
 
 	return false
 }
+
+func ResourceAccessScopeStrategy(matchers []string, needle string) bool {
+	needleResources := strings.Split(needle, ":")
+
+matcherloop:
+	for _, matcher := range matchers {
+		matcherResources := strings.Split(matcher, ":")
+
+		if len(matcherResources) != len(needleResources) {
+			continue matcherloop
+		}
+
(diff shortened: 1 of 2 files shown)


medhost-chara:
Before I open a separate discussion, I wanted to check if there are plans for defining a scope matching strategy for use in ORY Hydra. As part of the Smart App Launch US HTI-1 Guidelines to support Smart App Launch 2.0.0 - the scope formats were updated to the following format: https://hl7.org/fhir/smart-app-launch/STU2/scopes-and-launch-context.html#scopes-for-requesting-clinical-data

zoop-btc:
closing this since there is no interest by devs it seems

PR #814, opened 2024-07-09
fix: require redirect_uri for OpenID Connect calls

aeneasr:
Fixes an issue where Authorize Requests which were intended for an OpenID Connect 1.0 client would incorrectly be allowed when missing the redirect URI when it's required by the specification.

Closes https://github.com/ory/fosite/issues/685
Closes https://github.com/ory/fosite/pull/762

BREAKING CHANGES: Going forward, calls to `/oauth2/auth` which trigger OpenID Connect require the `redirect_uri` query parameter to be set.


aeneasr:
Blocked until August 1st

alnr (approved):

PR #826, opened 2024-09-30
Implement RFC 8628

nsklikas:
```
BREAKING CHANGES: This patch breaks up `OAuth2AuthorizeExplicitFactory` into
`OAuth2AuthorizeExplicitAuthFactory` and `Oauth2AuthorizeExplicitTokenFactory`
```

## Related Design Document

Implements [RFC 8628](https://datatracker.ietf.org/doc/html/rfc8628).

## Checklist

## Further comments

This PR is based on the work done on https://github.com/ory/fosite/pull/701, by @supercairos and @BuzzBumbleBee. That PR was based on an older version of fosite and was missing some features/tests.

Comments:
- Due to some dependency issues we had to switch to `go.uber.org/mock/gomock` (`github.com/golang/mock/gomock` is deprecated)
- We had to do a small refactor of the `AuthorizeExplicitGrantHandler` to generalize it to reduce code duplication


nsklikas:
Looks like the conformity tests are failing because it tries to use this version of fosite on hydra `master`. But they pass on the hydra [PR](https://github.com/ory/hydra/pull/3851) with this version of fosite. Changes are needed on the hydra config `DefaultProvider` object, how should we fix this?

go.mod
```diff
@@ -69,31 +71,32 @@ require (
...
+	google.golang.org/grpc v1.59.0 // indirect
 	google.golang.org/protobuf v1.33.0 // indirect
 	gopkg.in/ini.v1 v1.67.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-go 1.20
+go 1.21
+
+toolchain go1.21.4
```
aeneasr, line 102: Is 1.21 needed?

handler/rfc8628/strategy_hmacsha.go
```diff
@@ -0,0 +1,242 @@
...
+
+	return h.Enigma.Validate(ctx, strings.TrimPrefix(code, "ory_dc_"))
+}
+
+// ShouldRateLimit is used to decide whether a request should be rate-limited
+func (h *DefaultDeviceStrategy) ShouldRateLimit(context context.Context, code string) (bool, error) {
+	key := code + "_limiter"
+
+	keyBytes := []byte(key)
+	object, err := h.RateLimiterCache.Get(keyBytes)
```
aeneasr, line 179: This won't work in real world scenarios. The cache is in memory which means:

1. It won't work if more than one instance (which is almost always the case) of the library/hydra running
2. It won't persist across restarts

In my view, rate limiting should be the concern of the firewall/ingress/reverse proxy and I think we could simplify this PR a bit by removing this part or making it optional and/or more generic

handler/oauth2/flow_authorize_code_token.go
```diff
@@ -5,197 +5,99 @@ package oauth2
```
aeneasr, line 5: This diff is quite challenging to read, would it be possible to reduce the refactoring and separating into different files? Since this is security relevant code, the smaller the diff the faster the review :)

aeneasr (commented):
Thank you very much for this massive PR! There's still thousands of lines to review, but here are some first remarks.

Primarily, some of the diffs are hard to read because they re-organize code into different files. Would it be possible to focus the changes on what's required to change and leave the refactoring out? We can always re-organize the code later, but given the sensitivity of the files touched (core token methods), it will be significantly easier to review those changes. Thank you!

aeneasr:
> Looks like the conformity tests are failing because it tries to use this version of fosite on hydra `master`. But they pass on the hydra [PR](https://github.com/ory/hydra/pull/3851) with this version of fosite. Changes are needed on the hydra config `DefaultProvider` object, how should we fix this?

You should be able to change the target commit in the CI file to target your PR in hydra, which should make the tests pass

handler/oauth2/flow_authorize_code_token.go
zepatrik, line 1: I am wondering whether the split up is actually required. Can't we just do some internal refactoring to reuse the code, instead of this breaking change?
Regardless, I agree with @aeneasr that this should happen in a separate PR.

zepatrik (changes requested):
This looks great already, however due to the security implications I agree with @aeneasr that we should pull out some of the unrelated refactoring into separate PRs to reduce the size and complexity of this one, especially the migration from github.com/golang/mock/gomock to go.uber.org/mock/gomock and maybe also some other code style changes.
I did not yet fully review all parts, but these are some preliminary suggestions already.

handler/rfc8628/strategy_hmacsha.go
nsklikas: Agreed on all your points. The reason we included this change is to be spec compliant and the reason we didn't implement a more sophisticated solution (which would probably require an external caching system) is that rate limiting should be handled by the ingress/firewall. But looking back at it, it is probably better to leave it unhandled, instead of having partly correct logic. 

It is probably better to remove this and leave it as future work.

handler/oauth2/flow_authorize_code_token.go
nsklikas: We will try to revert the refactored code and push the changes in the coming days

handler/oauth2/flow_generic_code_token.go
```diff
@@ -0,0 +1,267 @@
...
+
+	// ValidateRedirectURI validates the redirect uri in the access request.
+	ValidateRedirectURI(accessRequester fosite.AccessRequester, authorizeRequester fosite.Requester) error
+
+	// GetGrantType retrieves the grant type from the request.
+	GetGrantType(requester fosite.AccessRequester) fosite.GrantType
+}
+
+// CodeHandler handles authorization/device code related operations.
+type CodeHandler interface {
```
zepatrik, line 37: As the device auth code grant has two codes (user & device), the naming should be clear at any point which one is used. Here I'm struggling really to understand which one it is.

J-tt: I strongly agree with this, I've been trying to follow along with an implementation (outside of Hydra), and it's been extremely unclear what the correct steps to handle user_code's are.

go.mod
nsklikas: It was required for the update of `ory/x`, I will roll it back to the lateset version that supports  `1.20`.

nsklikas:
Thank you for the reviews.

I reverted the refactor on the oauth2 handlers and tried to address all the comments. Also pinned the ci to use hydra from the PR branch and now the tests pass. Please have another go.

handler/rfc7523/handler_test.go
```diff
@@ -773,9 +773,9 @@ func (s *AuthorizeJWTGrantRequestHandlerTestSuite) createStandardClaim() jwt.Cla
 		Issuer:    "trusted_issuer",
 		Subject:   "some_ro",
 		Audience:  jwt.Audience{"https://www.example.com/token", "leela", "fry"},
-		Expiry:    jwt.NewNumericDate(time.Now().AddDate(0, 0, 23)),
+		Expiry:    jwt.NewNumericDate(time.Now().UTC().AddDate(0, 0, 23)),
 		NotBefore: nil,
-		IssuedAt:  jwt.NewNumericDate(time.Now().AddDate(0, 0, -7)),
+		IssuedAt:  jwt.NewNumericDate(time.Now().UTC().AddDate(0, 0, -7)),
```
nsklikas, line 778: Had to change this because tests were failing locally for me

zepatrik: I was a bit concerned by this one, as I also could reproduce it. Took me quite some time to understand it, but I finally got it:
This test uses the maximum configured JWT duration (720h = 30d). The problem is now with daylight saving, as `AddDate` respects that in an unexpected way:
https://go.dev/play/p/tqJSlsabYmr?v=

nsklikas: Ooh interesting, should I remove this then and just switch `AddDate` with `Add`?

handler/rfc8628/storage.go
```diff
@@ -0,0 +1,55 @@
...
+	// InvalidateDeviceCodeSession is called when a device code is being used. The state of the user
+	// code should be set to invalid and consecutive requests to GetDeviceCodeSession should return the
+	// ErrInvalidatedDeviceCode error.
+	InvalidateDeviceCodeSession(ctx context.Context, signature string) (err error)
+}
+
+// UserCodeStorage handles the user_code storage
+type UserCodeStorage interface {
+	// CreateUserCodeSession stores the device request for a given user code.
+	CreateUserCodeSession(ctx context.Context, signature string, request fosite.Requester) (err error)
```
zepatrik, line 42: I do see a problem with this signature:

When the signing secret is rotated, the same user code will have a different signature and we could run into problems with duplicates that the database will not be able to detect. Even worse, the users will both see the same user code on their device, and therefore everything looks correct to them. In the end they might still be able to log into a different device than they expected.

Is there any way we can solve this? If not, can you provide a rationale why this is not a problem?

nsklikas: That is correct, but isn't this true for all the tokens? AFAICT the same logic is used in https://github.com/ory/fosite/blob/master/handler/oauth2/storage.go#L47 and https://github.com/ory/fosite/blob/master/handler/oauth2/flow_authorize_code_token.go#L32 (even though the `GetAuthorizeCodeSession` calls the 2nd argument `code`, a `signature` is used). 
This approach is probably not needed for user codes, but the reason for implementing this way is to keep it consistent with the rest of the fosite tokens and to allow the implementer (e.g. hydra) to have a single function/logic for handling and storing them.

handler/rfc8628/auth_handler.go
```diff
@@ -0,0 +1,88 @@
...
+
+	return code, nil
+}
+
+func (d *DeviceAuthHandler) handleUserCode(ctx context.Context, dar fosite.DeviceRequester) (string, error) {
+	var err error
+	var userCode, signature string
+	// Retry when persisting user code fails
+	// Possible causes include database connection issue, constraints violations, etc.
+	for i := 0; i < MaxAttempts; i++ {
```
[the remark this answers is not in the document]
nsklikas: Exactly, I will update the comment to make this more clear.

handler/rfc7523/handler_test.go
zepatrik: I'd say either is fine, as this is just a test.

handler/rfc8628/storage.go
zepatrik: Ok, we also had some more internal discussion about user-code entropy and collisions. These were our thoughts:

A high user-code entropy is needed so that it gets hard (enough) to guess a code valid at any given point in time. Because user-codes are valid only for a short period, something like 1M guesses before a hit are probably sufficient. With rate-limiting you can obviously make it harder, but it will always be possible for someone with enough resources to circumvent the rate-limiting. For database conflicts, the same statistics apply, although a hit probability will be higher because user-codes will likely not be deleted immediately. The point here is that a production system will in a way "brute-force itself". This means that even if the chances of collisions are low, they will still happen given enough time. Here it helps to detect these rare cases early instead of presenting the user with an error later just so they have to retry.
Now considering secret rotation, it is probably OK if conflicts are not detected because it should not happen too often. However, as HMAC is relatively cheap, we would suggest to always check the codes with all or at least multiple past secrets before inserting. This could be a SQL query similar to `INSERT INTO ... SELECT 'sig1', ... WHERE NOT EXISTS (SELECT 1 FROM ... WHERE signature NOT IN ('sig1', 'sig2', 'sig3')`.
The consequence of duplicate codes is that users might be logged into different devices, but everything looks legitimate. This situation should really be avoided at any cost, and checking for duplicates during insert is a way better UX than an error when the user inputs the code and has to restart.

zepatrik: We could do that also for the other code types, but I think the important part here is that the entropy of user-codes is significantly less than all other codes, and therefore we really need to consider it here.

nsklikas: Thank you for the analysis, I agree with what you said. I think that the default character set we are using suffices, even in the rfc it is suggested that a smaller set can be used.

I am not sure any of this applies to the changes in fosite, as it does not implement the databases queries. Perhaps talking about this in the hydra PR would make more sense. Should I write this somewhere as guidelines for how to use the library?

zepatrik: Sure, this was more about how it should be used. I think it makes sense to add some of these considerations as docs to fosite.
It does impact fosite in the way that if one would want to use the multi-signature optimization, fosite would need to support that.

nsklikas: Could you point me to which doc files I should add this to?

I think that the current API allows for an implementer to check for multiple signatures (the implementation can be anything they want), unless what you want is for `UserCodeSignature` and `GenerateUserCode` to return an array of signatures and `CreateUserCodeSession`, `GetUserCodeSession` and `InvalidateUserCodeSession` to get an array of signatures as input. Can you be more specific on what you would like to change?

handler/openid/flow_device_auth.go
```diff
@@ -0,0 +1,50 @@
...
+	if resp.GetDeviceCode() == "" {
+		return errorsx.WithStack(fosite.ErrMisconfiguration.WithDebug("The device code has not been issued yet, indicating a broken code configuration."))
+	}
+
+	signature, err := c.DeviceCodeStrategy.DeviceCodeSignature(ctx, resp.GetDeviceCode())
+	if err != nil {
+		return err
+	}
+
+	if err := c.OpenIDConnectRequestStorage.CreateOpenIDConnectSession(ctx, signature, dar.Sanitize(oidcParameters)); err != nil {
```
nsklikas, line 45: One question is whether we want to create a session at this point (when the device authorization request happens) or if we want to persist the opendID session in the database after the user has logged in.

Currently we create the session when the request happens and we update it after the user authenticates. We could avoid one the first write.

cc @zepatrik what do you think?

zepatrik: From a high-level it sounds like we should rather avoid the initial write. Not sure if it has any implications later in the flow and on code complexity?

nsklikas: I think it should be fine, I will have a go on it

handler/openid/flow_device_token.go
```diff
@@ -0,0 +1,64 @@
...
+	if claims.Subject == "" {
+		return errorsx.WithStack(fosite.ErrServerError.WithDebug("Failed to generate id token because subject is an empty string."))
+	}
+
+	claims.AccessTokenHash = c.GetAccessTokenHash(ctx, requester, responder)
+
+	idTokenLifespan := fosite.GetEffectiveLifespan(requester.GetClient(), fosite.GrantTypeDeviceCode, fosite.IDToken, c.Config.GetIDTokenLifespan(ctx))
+	return c.IssueExplicitIDToken(ctx, idTokenLifespan, ar, responder)
+}
+
```
aeneasr, line 57: In flow_explicit_token, we are deleting the openid connect session on use:


	err = c.OpenIDConnectRequestStorage.DeleteOpenIDConnectSession(ctx, authorizeCode)
	if err != nil {
		return errorsx.WithStack(fosite.ErrServerError.WithWrap(err).WithDebug(err.Error()))
	}

I'm not sure if we need to do that here as well, or if this is covered by flow_explicit_token. However, as I understand it, flow_explicit_token is not called when performing an device auth code grant? So is this call missing?

handler/openid/flow_device_auth.go
```diff
@@ -0,0 +1,44 @@
...
+	DeviceCodeStrategy          rfc8628.DeviceCodeStrategy
+
+	Config interface {
+		fosite.IDTokenLifespanProvider
+	}
+
+	*IDTokenHandleHelper
+}
+
+func (c *OpenIDConnectDeviceHandler) HandleDeviceEndpointRequest(ctx context.Context, dar fosite.DeviceRequester, resp fosite.DeviceResponder) error {
```
aeneasr, line 27: Is this method even needed or can it just return nil? As far as I can tell, the only error state is when the code configuration is incorrect, however I'm not sure why we need to fail in that scenario. Can you please explain why we need to verify that the device code is set here?

handler/rfc8628/auth_handler.go
```diff
@@ -0,0 +1,76 @@
...
+		if err != nil {
+			return "", "", err
+		}
+
+		if err = d.Storage.CreateDeviceAuthSession(ctx, deviceCodeSignature, userCodeSignature, dar.Sanitize(nil).(fosite.DeviceRequester)); err == nil {
+			return deviceCode, userCode, nil
+		}
+	}
+
+	errMsg := fmt.Sprintf("Exceeded user-code generation max attempts %v: %s", MaxAttempts, err.Error())
```
aeneasr, line 74: err is potentially nil, which can panic

handler/rfc8628/auth_handler.go
```diff
@@ -0,0 +1,76 @@
...
+	// used codes will probably still be around for some time before they get cleaned,
+	// the chances of hitting a duplicate here can be higher.
+	// Three retries should be plenty, as otherwise the entropy is definitely off.
+	for i := 0; i < MaxAttempts; i++ {
+		userCode, userCodeSignature, err = d.Strategy.GenerateUserCode(ctx)
+		if err != nil {
+			return "", "", err
+		}
+
+		if err = d.Storage.CreateDeviceAuthSession(ctx, deviceCodeSignature, userCodeSignature, dar.Sanitize(nil).(fosite.DeviceRequester)); err == nil {
```
aeneasr, line 69: Please only retry on conflict errors (the ones SQL / storage returns when a duplicate is found). Otherwise, for example, you will add 3 attempts to a broken or slow database which increases load which increases risk of a longer outage.

handler/rfc8628/auth_handler.go
```diff
@@ -0,0 +1,76 @@
...
+		return "", "", errorsx.WithStack(fosite.ErrServerError.WithWrap(err).WithDebug(err.Error()))
+	}
+
+	dar.GetSession().SetExpiresAt(fosite.UserCode, time.Now().UTC().Add(d.Config.GetDeviceAndUserCodeLifespan(ctx)).Round(time.Second))
+	// Note: the retries are added here because we need to ensure uniqueness of user codes.
+	// The chances of duplicates should however be diminishing, because they are the same
+	// chance an attacker will be able to hit a valid code with few guesses. However, as
+	// used codes will probably still be around for some time before they get cleaned,
+	// the chances of hitting a duplicate here can be higher.
+	// Three retries should be plenty, as otherwise the entropy is definitely off.
```
aeneasr, line 62: Understand where this is coming from, but not super happy with the fact that we need it.

aeneasr (commented):
Solid work! A few more coments left. I still have to go through a few more tests and verify the implementation on Hydra side. But this looks very good and I think we're close to merging this

handler/rfc8628/auth_handler.go
nsklikas: I believe that the error cannot be nil.

If the error was nil, the function would return on https://github.com/ory/fosite/pull/826/files/73097533d43a8fdd13fe8c461ce50716209baf51#diff-0846345bc5315b41b4f3ac45f5ae5dd7257ef1e053dc844ba87d7d86919f7bbeR69 (unless someone set `MaxAttempts=0`):
```
 		if err = d.Storage.CreateDeviceAuthSession(ctx, deviceCodeSignature, userCodeSignature, dar.Sanitize(nil).(fosite.DeviceRequester)); err == nil {
```

aeneasr: It was flagged by CI / tooling that it might be nil, so it would be great to properly handle it! It might also become nil in the future, if code changes are made, because the value checking isn't explicit enough

integration/authorize_device_grant_request_test.go
```diff
@@ -0,0 +1,227 @@
...
+			},
+			check: func(t *testing.T, token *goauth.Token, err error) {
+				assert.ErrorContains(t, err, "invalid_client")
+			},
+			cleanUp: func() {
+				oauthClient.ClientID = "device-client"
+			},
+		},
+		{
+			description: "should pass",
```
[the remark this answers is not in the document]
nsklikas: Unless you have an issue with that, I will update the existing test case to check for id/refresh tokens.

handler/openid/flow_device_auth.go
nsklikas: You are right! The only use for these checks is to catch an invalid configuration when composing the fosite structs. We put it there because similar logic exists in https://github.com/ory/fosite/blob/master/handler/openid/flow_explicit_auth.go#L46, BUT I now realize that this check exists there because there is openid specific logic for the auth req.

For now no extra logic is required for the device authn request in oidc, I will remove it.

eseliger:
Just wanted to leave a big thank you for the massive work you put into this @nsklikas!

handler/openid/flow_device_token.go
aeneasr: @nsklikas could you please check this comment?

handler/openid/flow_device_auth.go
aeneasr: Great!

aeneasr (commented):
Amazing work, there is one comment left around DeleteOpenIDSession, otherwise this is good to merge IMO

Next step is to get the hydra changes reviewed and then we can merge everything

handler/rfc8628/strategy_hmacsha.go
```diff
@@ -0,0 +1,94 @@
...
+	return userCode, signUserCode, nil
+}
+
+// UserCodeSignature generates a user_code signature
+func (h *DefaultDeviceStrategy) UserCodeSignature(ctx context.Context, token string) (string, error) {
+	return h.Enigma.GenerateHMACForString(ctx, token)
+}
+
+// ValidateUserCode validates a user_code
+func (h *DefaultDeviceStrategy) ValidateUserCode(ctx context.Context, r fosite.DeviceRequester, code string) error {
```
aeneasr, line 51: code is not being validated here?

handler/rfc8628/strategy_hmacsha.go
```diff
@@ -0,0 +1,94 @@
...
+	exp := r.GetSession().GetExpiresAt(fosite.UserCode)
+	if exp.IsZero() && r.GetRequestedAt().Add(h.Config.GetDeviceAndUserCodeLifespan(ctx)).Before(time.Now().UTC()) {
+		return errorsx.WithStack(fosite.ErrDeviceExpiredToken.WithHintf("User code expired at '%s'.", r.GetRequestedAt().Add(h.Config.GetDeviceAndUserCodeLifespan(ctx))))
+	}
+	if !exp.IsZero() && exp.Before(time.Now().UTC()) {
+		return errorsx.WithStack(fosite.ErrDeviceExpiredToken.WithHintf("User code expired at '%s'.", exp))
+	}
+	return nil
```
aeneasr, line 59: Actually, I think you forgot to validate the code here

nsklikas: Good catch, fixed

handler/openid/flow_device_token.go
nsklikas: You were right that it was missing, we either missed it or something had changed after we started working on this feature. Your suggestion has been added (see https://github.com/ory/fosite/pull/826/files#diff-408286bd02ab957e8c252a7ae6beb1c848873cb66ffba56ca0cf29dbe0658bd6R52)

handler/rfc8628/strategy_hmacsha.go
aeneasr: So you store the code plain text in the database? Normally, we only store the HMAC-SHA signature of the token/code in the DB.

nsklikas: No, only the signature is stored. 
When we get a user_code, we calculate its signature, fetch the device request from the database and then call `ValidateUserCode`. This is similar to how the `authorization_code` is validated (https://github.com/ory/fosite/blob/master/handler/oauth2/flow_authorize_code_token.go#L30).

The only difference is that the auth_code consists of a plain token + its signature, so you can validate the signature. The `user_code` is just a plain token because it is of limited length, so we can't validate the token on it's own.

Perhaps the name `ValidateUserCode` is a little misleading, as it validates only that the code is still active and not the actual string.

aeneasr: Ok, I understand, that makes sense. Maybe we adjust the API naming slightly to make it less confusing, or we add a code comment to this specific part (potentially also just pointing to this comment) to make sure this is preserved as a reason for the future / pen testing.

nsklikas: I tried to add a comment about it, hope it makes sense. I decided not to update the method name as I felt that it is better to keep the naming consistent with other strategies.

aeneasr (approved):

PR #827, opened 2024-10-14
chore: minor improvements to readability and updated code style

zepatrik:
## Checklist

- [ ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md) and signed the CLA.
- [ ] I have referenced an issue containing the design document if my change introduces a new feature.
- [ ] I have read the [security policy](../security/policy).
- [ ] I confirm that this pull request does not address a security vulnerability. 
      If this pull request addresses a security vulnerability, 
      I confirm that I got approval (please contact [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push the changes.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have added the necessary documentation within the code base (if appropriate).

PR #835, opened 2024-11-04
revert: signature extraction in the HMAC strategy

zepatrik:
This reverts a small change from #827 and adds tests that would have caught this. We noticed this in hydra, because there are some tests that ran into this.
@nsklikas


aeneasr:
Looks like there are some build issues now in Hydra?

PR #838, opened 2024-11-29
refactor: refresh token rotation

aeneasr:
## Checklist

### DIFF
--- a/handler/oauth2/flow_refresh.go
+++ b/handler/oauth2/flow_refresh.go
@@ -69,7 +69,6 @@ func (c *RefreshTokenGrantHandler) HandleTokenEndpointRequest(ctx context.Contex
 		scopeNames := strings.Join(c.Config.GetRefreshTokenScopes(ctx), " or ")
 		hint := fmt.Sprintf("The OAuth 2.0 Client was not granted scope %s and may thus not perform the 'refresh_token' authorization grant.", scopeNames)
 		return errorsx.WithStack(fosite.ErrScopeNotGranted.WithHint(hint))
-
 	}
 
 	// The authorization server MUST ... and ensure that the refresh token was issued to the authenticated client
@@ -134,25 +133,18 @@ func (c *RefreshTokenGrantHandler) PopulateTokenEndpointResponse(ctx context.Con
 		err = c.handleRefreshTokenEndpointStorageError(ctx, err)
 	}()
 
-	ts, err := c.TokenRevocationStorage.GetRefreshTokenSession(ctx, signature, nil)
-	if err != nil {
-		return err
-	} else if err := c.TokenRevocationStorage.RevokeAccessToken(ctx, ts.GetID()); err != nil {
-		return err
-	}
-
-	if err := c.TokenRevocationStorage.RevokeRefreshTokenMaybeGracePeriod(ctx, ts.GetID(), signature); err != nil {
-		return err
-	}
-
 	storeReq := requester.Sanitize([]string{})
-	storeReq.SetID(ts.GetID())
+	storeReq.SetID(requester.GetID())
+
+	if err = c.TokenRevocationStorage.RotateRefreshToken(ctx, requester.GetID(), signature); err != nil {
+		return err
+	}
 
 	if err = c.TokenRevocationStorage.CreateAccessTokenSession(ctx, accessSignature, storeReq); err != nil {
 		return err
 	}
 
-	if err = c.TokenRevocationStorage.CreateRefreshTokenSession(ctx, refreshSignature, storeReq); err != nil {
+	if err = c.TokenRevocationStorage.CreateRefreshTokenSession(ctx, refreshSignature, accessSignature, storeReq); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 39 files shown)


alnr (approved):

handler/oauth2/flow_refresh.go
```diff
@@ -51,25 +51,31 @@ func (c *RefreshTokenGrantHandler) HandleTokenEndpointRequest(ctx context.Contex
 			return errorsx.WithStack(rErr)
 		}
 
-		return errorsx.WithStack(fosite.ErrInactiveToken.WithWrap(err).WithDebug(err.Error()))
+		return fosite.ErrInvalidGrant.WithWrap(err).
```
zepatrik, line 54: Why remove the stack for all these errors?
Also, why change the explicit error here? Could this be breaking? From the hint it seems like this should still be `fosite.ErrInactiveToken`.

handler/oauth2/flow_refresh.go
```diff
@@ -51,25 +51,31 @@ func (c *RefreshTokenGrantHandler) HandleTokenEndpointRequest(ctx context.Contex
...
+			WithDebugf("Refresh token re-use was detected. All related tokens have been revoked.")
 	} else if errors.Is(err, fosite.ErrNotFound) {
-		return errorsx.WithStack(fosite.ErrInvalidGrant.WithWrap(err).WithDebugf("The refresh token has not been found: %s", err.Error()))
+		return fosite.ErrInvalidGrant.WithWrap(err).
+			WithHint("The refresh token is malformed or not valid.").
+			WithDebug("The refresh token can not be found.")
 	} else if err != nil {
-		return errorsx.WithStack(fosite.ErrServerError.WithWrap(err).WithDebug(err.Error()))
-	} else if err := c.RefreshTokenStrategy.ValidateRefreshToken(ctx, originalRequest, refresh); err != nil {
+		return fosite.ErrInvalidGrant.WithWrap(err).WithDebug(err.Error())
```
zepatrik, line 62: Also here, the error should probably stay `fosite.ErrServerError`.

zepatrik (changes requested):

handler/oauth2/flow_refresh.go
zepatrik: OK, the spec is just weird here and `fosite.ErrInvalidGrant` is correct.

aeneasr: > Why remove the stack for all these errors?

`WithWrap` uses the passed error to do the stack trace - so the internal error already has a stack trace and then we just re-use this. It's a bit like `fmt.Errorf("%w",err)`

handler/oauth2/flow_refresh.go
```diff
@@ -130,30 +136,20 @@ func (c *RefreshTokenGrantHandler) PopulateTokenEndpointResponse(ctx context.Con
...
-		return err
-	} else if err := c.TokenRevocationStorage.RevokeAccessToken(ctx, ts.GetID()); err != nil {
-		return err
-	}
+	storeReq := requester.Sanitize([]string{})
+	storeReq.SetID(requester.GetID())
 
-	if err := c.TokenRevocationStorage.RevokeRefreshTokenMaybeGracePeriod(ctx, ts.GetID(), signature); err != nil {
-		return err
+	if err = c.TokenRevocationStorage.RotateRefreshToken(ctx, requester.GetID(), signature); err != nil {
```
mitar, line 143: @aeneasr I think this introduced a bug. After we upgraded our systems to the version with this change, no refresh tokens in our tests (where we use memory storage) got invalidated. The reason I think is that `requester` here has a completely new and random ID, which of course has nothing with the existing refresh token.

Previous code had it right, it retrieved refresh token requester with `GetRefreshTokenSession` and used `ts.GetID()` here. I think this should be fixed and correct requester ID should be passed here.

mitar: My current workaround is to fix this at store level:

```
func (s *MemoryStore) RotateRefreshToken(ctx context.Context, _ string, refreshTokenSignature string) error {
	ts, err := s.GetRefreshTokenSession(ctx, refreshTokenSignature, nil)
	if err != nil {
		return err
	}
	if err := s.RevokeRefreshToken(ctx, ts.GetID()); err != nil {
		return err
	}
	return s.RevokeAccessToken(ctx, ts.GetID())
}
```

But this is just a workaround, I think, it is completely useless to pass new requester ID into this function. Or is this something you use and need in Hydra?

aeneasr: The request ID is supposed to remain stable for the lifetime duration of a token chain, so this is correct! You can check the implementation in Hydra for reference

mitar: You are right. It works. I had in our code a part where I was unconditionally setting request ID to a new value. The reason is that I want a bit shorter ID representation than what default `uuid.New().String()` provides. I will have to find another way to do that. I could not find an easy way to patch `GetID()` to use a different ID constructor. :-(

Thanks.

aeneasr: Fosite’s not the easiest library to work with

mitar: Oh, it is great!

PR #847, opened 2025-04-08
feat: configure client assertion issuer

tilgovi:
Allow clients to have a custom client assertion issuer so that a client assertion can be issued by a separate identity provider rather than only self-issued.

BREAKING CHANGES: The `Client` interface defines a new method to return the issuer expected in client assertions. Applications that define their own type must implement `Client.GetJSONWebTokenClientAssertionIssuer`.

## Checklist

- [X] I have added the necessary documentation within the code base (if appropriate).

### DIFF
--- a/client.go
+++ b/client.go
@@ -58,6 +58,9 @@ type OpenIDConnectClient interface {
 	// public key used by the client to authenticate.
 	GetJSONWebKeysURI() string
 
+	// GetJSONWebTokenClientAssertionIssuer returns the issuer of client assertions used to authenticate.
+	GetJSONWebTokenClientAssertionIssuer() string
+
 	// JWS [JWS] alg algorithm [JWA] that MUST be used for signing Request Objects sent to the OP.
 	// All Request Objects from this Client MUST be rejected, if not signed with this algorithm.
 	GetRequestObjectSigningAlgorithm() string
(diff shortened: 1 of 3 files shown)


tilgovi:
I did not open a separate issue with a design document because the change being proposed here seems small enough that it might not make sense.

The idea here is that [RFC 7521 Section 6.1](https://datatracker.ietf.org/doc/html/rfc7521#section-6.1) suggests that assertions for client authentication do not need to be self-issued:

> The Subject of the assertion identifies the client.  If the assertion is self-issued by the client, the Issuer of the assertion also identifies the client.

The use of "if" in the second sentence quoted above suggests that the issuer does not need to identify the client if the assertion is not self-issued.

Hydra requires that client assertions have the `client_id` in the `iss` claim, but this feature removes that requirement, enabling Hydra (or other applications) to allow customization of the issue for each client. This feature helps with federation scenarios where another issuer exists already. This scenario is achievable with Hydra today by using JWT authorization grants, but requires managing the trust relationship separately from the client and requires clients to specify the Client ID in a request parameter. Using a client assertion more accurately captures the intent for the client to obtain authorization to act as itself.

tilgovi:
I'm very open to any feedback at all as this is my first fosite contribution.

tilgovi:
I see that there are tests that are failing because they try to build Hydra against this branch. I have a working Hydra branch, but it doesn't have any tests yet.

tilgovi:
Here is the [Hydra branch](https://github.com/ory/hydra/compare/master...tilgovi:hydra:feat-jwt-client-assertion-issuer) in progress.

mitar:
Please rebase from latest master branch to fix CI issues.

PR #854, opened 2025-06-20
fix: sign with jose.OpaqueSigner check incorrect key type

elct9620:

I am working on an internal OAuth2 server that needs to sign JWTs using AWS KMS. 

However, the `jwt.DefaultSigner` doesn't work correctly. It is caused by checking the public key against the private key type.

https://github.com/ory/fosite/blob/8052806a6efe52832867d24cd41b17463190903d/token/jwt/jwt.go#L58-L79

If I provide a `jose.OpaqueSigner` wrapper to the AWS KMS, it will get this error

```
Error creating access response: unsupported private / public key pairs: *example.KmsSigner, *rsa.PublicKey
```

To fix it, I change `PrivateKey` to `PublicKey` and add new test case to cover it.

## Checklist


mitar:
Please rebase from latest master branch to fix CI issues.

elct9620:
@mitar done 💪

aeneasr (changes requested):
To create a JWT you need a private key to sign it

To verify a JWT you need the public key to verify the signature

This change will break that

elct9620:
Hi, @aeneasr 

Can you explain more details about it?

The [jose.OpaqueSigner](https://pkg.go.dev/github.com/go-jose/go-jose/v3#OpaqueSigner) have following interface:

```go
type OpaqueSigner interface {
	// Public returns the public key of the current signing key.
	Public() *JSONWebKey
	// Algs returns a list of supported signing algorithms.
	Algs() []SignatureAlgorithm
	// SignPayload signs a payload with the current signing key using the given
	// algorithm.
	SignPayload(payload []byte, alg SignatureAlgorithm) ([]byte, error)
}
```

It means we only know the public key but cannot access the private key (stored on HSM or remote service, e.g., AWS KMS)

I changed the type from `PrivateKey` to `PublicKey` because we do not have any private key in our memory, and the `Generate` method uses `t.Public().Key.(type)` to get its type.

It enables the correct detection of the key type and continues to generate the key. After the key is signed by HSM, we are still able to use `t.Public()` to get the public key to verify it.

This change only applies to `Generate` to allow it can choose the correct alg type, I think it does not break the `Validate` method.

PR #859, opened 2025-07-02
chore: slightly reduce session JSON size

zepatrik:
If someone stores the session as JSON (as we do in Hydra), omitting empty keys can slightly reduce the storage size, while resulting in the exact same behavior.

### DIFF
--- a/handler/oauth2/strategy_jwt.go
+++ b/handler/oauth2/strategy_jwt.go
@@ -111,10 +111,10 @@ func toRFCErr(v *jwt.ValidationError) *fosite.RFC6749Error {
 func (h *DefaultJWTStrategy) generate(ctx context.Context, tokenType fosite.TokenType, requester fosite.Requester) (string, string, error) {
 	if jwtSession, ok := requester.GetSession().(JWTSessionContainer); !ok {
 		return "", "", errors.Errorf("Session must be of type JWTSessionContainer but got type: %T", requester.GetSession())
-	} else if jwtSession.GetJWTClaims() == nil {
+	} else if claims := jwtSession.GetJWTClaims(); claims == nil {
 		return "", "", errors.New("GetTokenClaims() must not be nil")
 	} else {
-		claims := jwtSession.GetJWTClaims().
+		claims.
 			With(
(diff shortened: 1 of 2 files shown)


handler/oauth2/strategy_jwt.go
```diff
@@ -111,10 +111,10 @@ func toRFCErr(v *jwt.ValidationError) *fosite.RFC6749Error {
 func (h *DefaultJWTStrategy) generate(ctx context.Context, tokenType fosite.TokenType, requester fosite.Requester) (string, string, error) {
 	if jwtSession, ok := requester.GetSession().(JWTSessionContainer); !ok {
 		return "", "", errors.Errorf("Session must be of type JWTSessionContainer but got type: %T", requester.GetSession())
-	} else if jwtSession.GetJWTClaims() == nil {
+	} else if claims := jwtSession.GetJWTClaims(); claims == nil {
 		return "", "", errors.New("GetTokenClaims() must not be nil")
 	} else {
-		claims := jwtSession.GetJWTClaims().
+		claims.
```
zepatrik, line 117: Avoiding double computation, in Hydra this is not cheap.

handler/openid/strategy_jwt.go
```diff
@@ -36,8 +36,8 @@ type DefaultSession struct {
 	Claims    *jwt.IDTokenClaims             `json:"id_token_claims"`
 	Headers   *jwt.Headers                   `json:"headers"`
 	ExpiresAt map[fosite.TokenType]time.Time `json:"expires_at"`
-	Username  string                         `json:"username"`
-	Subject   string                         `json:"subject"`
+	Username  string                         `json:"username,omitempty"`
+	Subject   string                         `json:"subject,omitempty"`
```
zepatrik, line 40: In case you're wondering, for strings empty == zero. Fosite is not yet on go1.24.

aeneasr: We can definitely move this library to 1.24 but since omitempty also works we don't have to.

aeneasr (approved):

PR #862, opened 2025-07-25
feat: pass new expiry to refresh rotation

aeneasr:


zepatrik (approved):

zepatrik:
Why exactly do the gomock change?

james-d-elliott:
~~The old gomock module was archived, uber picked up the project and has made improvements to it but had a module rename etc. This could also be handled in a separate chore / refactor PR technically.~~

Ignore me, it's the opposite of what I thought. Think this is because of an old local mockgen?

aeneasr:
> ~The old gomock module was archived, uber picked up the project and has made improvements to it but had a module rename etc. This could also be handled in a separate chore / refactor PR technically.~
> 
> Ignore me, it's the opposite of what I thought. Think this is because of an old local mockgen?

Canonical introduced uber's go mock, and broke the tooling in the process (generating new mocks), so I reverted the change and since this PR needs updated mocks, I did it in this PR

james-d-elliott:
Ah that makes sense (didn't see the second commit). Maybe this was missed? https://github.com/ory/fosite/blob/b252c87be3214e5b3c73fe5f2429f7495b9dc942/Makefile#L24

PR #864, opened 2025-09-08
fix: context passing and response size limit

alnr:
Stumbled across this by chance.

### DIFF
--- a/authorize_request_handler.go
+++ b/authorize_request_handler.go
@@ -11,16 +11,15 @@ import (
 	"strings"
 
 	"github.com/go-jose/go-jose/v3"
+	"github.com/hashicorp/go-retryablehttp"
+	"github.com/pkg/errors"
 	"go.opentelemetry.io/otel/trace"
 
 	"github.com/ory/fosite/i18n"
 	"github.com/ory/fosite/token/jwt"
+	"github.com/ory/go-convenience/stringslice"
 	"github.com/ory/x/errorsx"
 	"github.com/ory/x/otelx"
-
-	"github.com/pkg/errors"
-
-	"github.com/ory/go-convenience/stringslice"
 )
 
 func wrapSigningKeyFailure(outer *RFC6749Error, inner error) *RFC6749Error {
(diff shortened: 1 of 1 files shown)


gaultier (approved):
LGTM but there's something wrong with the CI setup

authorize_request_handler.go
```diff
@@ -66,11 +65,16 @@ func (f *Fosite) authorizeRequestParametersFromOpenIDConnectRequest(ctx context.
...
+		req, err := retryablehttp.NewRequestWithContext(ctx, "GET", location, nil)
+		if err != nil {
+			return errorsx.WithStack(ErrInvalidRequestURI.WithHintf("Unable to fetch OpenID Connect request parameters from 'request_uri' because: %s.", err.Error()).WithWrap(err).WithDebug(err.Error()))
+		}
+		response, err := hc.Do(req)
 		if err != nil {
 			return errorsx.WithStack(ErrInvalidRequestURI.WithHintf("Unable to fetch OpenID Connect request parameters from 'request_uri' because: %s.", err.Error()).WithWrap(err).WithDebug(err.Error()))
 		}
 		defer response.Body.Close()
+		response.Body = io.NopCloser(io.LimitReader(response.Body, 10*1024*1024)) // limit to 10MiB
```
vivshankar, line 77: Should the limit be configurable? You could make it a config property in Fosite struct itself.

alnr: I don't think that's necessary. This is a defence mechanism only. Do you think 10MiB is too small?

vivshankar: Hard to say, particularly with RAR. I don't see any reason why it shouldn't be parameterized.

alnr: > I don't see any reason why it shouldn't be parameterized.

The reason is that it's more complicated (more code).

vivshankar: IMO this should be controlled by the OP using Fosite. Perhaps just a property in the struct? It doesn't have to be a full blown Configurator...

alnr:
Closing in favor of https://github.com/ory-corp/cloud/pull/9912

PR #879, opened 2026-05-25
feat: add RFC 8707 Resource Indicators support

cohendvir:
Adds support for [RFC 8707 Resource Indicators](https://www.rfc-editor.org/rfc/rfc8707) in fosite.

> **Note:** This PR stacks on #880 (copyright year bump). Please merge #880 first — once it lands on `master`, the diff here will show only the RFC 8707 changes.


james-d-elliott:
Shouldn't this be in the authorize code / implicit auth handlers and refresh / client credentials / device code access handlers?

james-d-elliott:
Actually looking more closely it's similar to refresh flow, the values requested on the auth endpoint restrict what can be requested on the token endpoint, same for the original request in refresh flows.

>  The resource value(s) that is acceptable to an authorization server in fulfilling an access token request is at its sole discretion based on local policy or configuration. In the case of a refresh_token or authorization_code grant type request, such policy may limit the acceptable resources to those that were originally granted by the resource owner or a subset thereof. In the authorization_code case where the requested resources are a subset of the set of resources originally granted, the authorization server will issue an access token based on that subset of requested resources, whereas any refresh token that is returned is bound to the full original grant.

